PR #15, opened 2016-09-01
RF+ENH: make pass travis tests (so depends fixed up)

yarikoptic:


yarikoptic:
now we have "baseline" coverage which we should only improve ;)

PR #16, opened 2016-09-14
ENH: working on example as an example for tests

yarikoptic:


repronim/examples/example.py
```diff
@@ -4,20 +4,21 @@
...
-provenance = ProvenanceParser.factory('reprozip_output.yml', format='reprozip')
-# orchestrator = Orchestrator.factory('localhost', provenance)
-# orchestrator = Orchestrator.factory('docker', provenance)
-dpm = DebianPackageManager()
+def create_reprozip_factor():
+    # provenance = ProvenanceParser.factory('pipype_output.trig', format='trig')
+    provenance = ProvenanceParser.factory('reprozip_output.yml', format='reprozip')
+    # orchestrator = Orchestrator.factory('localhost', provenance)
+    # orchestrator = Orchestrator.factory('docker', provenance)
+    dpm = DebianPackageManager()
```
yarikoptic: why is it here???

yarikoptic:
d'oh -- now I need to make that .rst file I have contributed a proper .rst to pass all the tests (documentation is also built)

yarikoptic:
Here we go - coverage reported but apparently there were none recorded. So merging to establish the baseline

yarikoptic:
And I forgot that you can't merge while on the phone wherever any test fails, so feel welcome to merge our I will do latter when on the laptop

PR #22, opened 2016-09-30
DOC: Update example modline for vim and emacs

rbuccigrossi:
With some testing, setting the modline to 76 allows essentially the same behavior across vim, emacs and pyCharm (when the code width in pyCharm is set to 79 characters).  Why the difference of 3 characters? I believe that:

1. PyCharm adds spaces at the end of every line in the paragraph
1. PyCharm cuts off *before* 79 characters instead of *at* 79 characters
1. PyCharm starts counting at 1 instead of 0

Fun, eh?

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -1,5 +1,5 @@
-.. -*- mode: rst; fill-column: 78; indent-tabs-mode: nil -*-
-.. vi: set ft=rst sts=4 ts=4 sw=4 et tw=79:
+.. -*- mode: rst; fill-column: 76; indent-tabs-mode: nil -*-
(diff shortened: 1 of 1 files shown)


rbuccigrossi:
I switched the fill column to 79 for both vi and emacs (they both fill to the same column correctly and don't add spaces, so behaves much better than fill-paragraph in pyCharm.  However, I discovered View->Active Editor-> Use Soft Wraps, which may be a better way to go anyway (it wraps the text to the edge of the editor.  The same effect can be done in emacs with M-x visual-line-mode and in vi with 

:set wrap
:set linebreak
:set nolist

Anyway, I'm done playing around with word wrapping, and it's time to actually get to work...

yarikoptic:
Let it be! Visual mode is suboptimal for diffs since they are done per line

Issue #2, opened 2016-05-04
package python-prov


yarikoptic:
dev version of nipype relies on it (https://pypi.python.org/pypi/prov/1.4.0)

for now issues just a RFP (request for packaging) issue within debian bts (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=823436).  If noone does it, will have to take care about it

PR #25, opened 2016-10-06
Rewriting high-level architecture discussion centered on the concept of packages

rbuccigrossi:
I'm not done (so don't accept this yet). However, I wanted to use this pull request to spark discussion and allow you to see the progress as I go along...

### DIFF
--- a/docs/source/sandbox/packages.rst
+++ b/docs/source/sandbox/packages.rst
@@ -1,7 +1,5 @@
-High-level brain dumps
-**********************
-
-This is just a thinking aloud notes to motivate and structure design decisions.
+High-level Package Handling (and ReproZip Architecture Discussion)
+******************************************************************
 
 What ReproNim aims (not) to be
 ==============================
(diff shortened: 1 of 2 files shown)


docs/source/sandbox/packages.rst
```diff
@@ -11,40 +9,63 @@ providers etc), which we will call 'backends', and provide a very high level,
...
+operate). 
+
+Packages are installed, removed, and queried through the use of "package
+managers." There are different package managers for different components of an
+environment and have slightly different capabilities.  For example, "yum" and
+"apt-get" are used to install binary and source files on a Linux operating
+system.  "pip" provides download and compilation capability for the Python
+interpreted language, while "conda" is another Python package manager that can
+supports "virtual environments" (essentially subdirectories) that provide
+separate parallel Python environments.
```
yarikoptic: see `Myth #2` of https://jakevdp.github.io/blog/2016/08/25/conda-myths-and-misconceptions/   ;)  so conda aims to provide more generic than just Python environments support

rbuccigrossi:
Oh, that's interesting.  I'm still doing work and committing to this branch. Does this go into a new pull request since you merged?

Also, I see why you called these "Overlay" distributions and package managers.  I was trying to break them into different categories to tease out the classes of package managers that we can automatically analyze (trace) to identify contributed files and those that would be impossible (like pulling git sources).  "Overlay" doesn't seem obvious to me, so I was hoping to find a term like "platform Independent" or "non-OS" but nothing seems to fit well.

As for conda being language independent, that makes sense.  Package managers in general are much less platform or language dependent than the repositories and distributions upon which they pull.

PR #32, opened 2016-10-24
Code cleanup. Removing old Packer code that is not going to be used to create images.

mjtravers:


yarikoptic:
I wouldn't yet swear that we wouldn't use packer... We as well might to establish the base images, and then use our handler/installer to populate/configure them.
But for clarity we could indeed for now remove these pieces. We could always come back to them if decide to do so

PR #37, opened 2016-11-30
Added resource management and AWS EC2 container management.

mjtravers:
Yarik,

There is a new repronim.resource package for managing resources defined in a repronim.cfg file. You can see and example in repronim/tests/files/repronim.cfg.

I also created an container.ec2 package that can start up and push commands to EC2 instances.

The install command now has 3 functioning parameters:
--spec = path to spec file
--resource = identifier to a resource defined in repronim.cfg
--config = path to the repronim.cfg file

Matt


repronim/container/dockercontainer.py
```diff
@@ -1,156 +1,146 @@
...
+        -------
+        list
+            List of STDOUT lines from the container.
+        """
+
+        command_env = self.get_updated_env(env)
+
+        if command_env:
+            # TODO: might not work - not tested it
+            command = ['%s=%s;' % k for k in command_env.items()] + command
```
yarikoptic: coverage says that it is covered ;)   
but I think you need to export those all so they propagate into child processes within, i.e `'export %s=%s'`.  Then also look out for those which might have spaces in them, so you might want to use `export %s=%r` ?

repronim/container/localshell.py
```diff
@@ -6,25 +6,39 @@
...
 
-    def create(self, base_image_id=None):
+    def create(self, image_id=None):
         """
         Create a container instance.
 
         Parameters
         ----------
-        base_image_id : string
+        image_id : string
```
yarikoptic: I guess `image_id` is not applicable here, right? so may be assure that below, (assert), or at least issue a warning if non None?

repronim/distribution/conda.py
```diff
@@ -1,55 +1,55 @@
...
+
+        Parameters
+        ----------
+        container : object
+            Container sub-class instance.
+        """
+        for package in self.provenance['packages']:
+            container.add_command(['conda',
+                                   'install',
+                                   package['name']])
```
yarikoptic: keep in mind that we might need specific versions of the packages...

repronim/distribution/debian.py
```diff
@@ -1,59 +1,59 @@
...
+        Perform any initialization commands needed in the container environment.
+
+        Parameters
+        ----------
+        container : object
+            The container sub-class object the hold the environment.
+        """
+        self.lgr.debug("Adding Debian update to container command list.")
+        container.add_command(['apt-get', 'update'])
+        return
```
yarikoptic: no harm done but there is no need in blank returns ;)

mjtravers:
Yarik,

The broken test has been fixed. Branch ready for merge into master branch.

PR #40, opened 2016-12-22
BF: Improved error handling from commands sent to env.

mjtravers:
Small changes to the docker and AWS container packages.


repronim/container/dockercontainer.py
```diff
@@ -80,8 +75,12 @@ def execute_command(self, command, env=None):
             # TODO: might not work - not tested it
             command = ['export %s=%s;' % k for k in command_env.items()] + command
         execute = self._client.exec_create(container=self._container_id, cmd=command)
-        response = [line for line in self._client.exec_start(exec_id=execute['Id'], stream=True)]
-        return response
+
+        for i, line in enumerate(self._client.exec_start(exec_id=execute['Id'], stream=True)):
+            if 'error' in line.lower():
```
yarikoptic: what is output? if that is just the standard output of the running the command, then we can't rely on sensing 'error' in it (might program might output "there were no error" ;)... or is that some "standardized" output on status of docker running the command?

mjtravers:
After doing some testing on the docker process, "rpc error" seems to be consistently returned from the docker server when a command is executed and fails. Will have to look further into AWS to do a better check there as well.

The python package to communicate was changed from the "dockery-py" package to the "docker" package.

In this commit, I also updated the testing to pytest when running "repronim test".

repronim/environment/dockerenvironment.py
```diff
@@ -0,0 +1,165 @@
...
+    def __init__(self, config = {}):
+        """
+        Class constructor
+
+        Parameters
+        ----------
+        config : dictionary
+            Configuration parameters for the environment.
+        """
+        if not 'base_image_id' in config:
```
yarikoptic: in general `x not in` is preferred to `not x in` (pycharm should have hinted on that I believe)

yarikoptic:
since tests pass, let's merge and proceed from there ;)

PR #43, opened 2017-01-08
Added AWS SSH key creation

mjtravers:
This pull request includes:

1. Adding code to prompt the user for information if they don't have a key_name declared in their ec2-environment configuration.
2. Documenting the options in repronim.cfg at the top of the file in tests/files/repronim.cfg
3. After reading the first chapter of "Fluent Python", made chages to the Resource class to make it more Pythonic.
4. Enhanced error handling and user notices for command line options.

The unit testing of the new Ec2Environment.create_key_pair() method is beyond me at this point as it contains user prompts. Will work on that and bring it up at Monday's meeting to see if you have any thoughts on how to handle it.

My next step is to work on enhancing "repronim ls" and on allowing users to log directly into their environments through the repronim command line.

Enjoy the new code!

### DIFF
--- a/repronim/environment/ec2environment.py
+++ b/repronim/environment/ec2environment.py
@@ -169,8 +169,11 @@ class Ec2Environment(Environment):
         Walk the user through creating an SSH key pair that is saved to
         the AWS platform.
         """
+
+        # TODO: Should move this to something usable application-wide.
+        py3 = version_info[0] > 2
+
         def get_user_input(prompt):
-            py3 = version_info[0] > 2
             if py3:
                 response = input(prompt)
             else:
@@ -239,7 +242,10 @@ Please enter a unique name to create a new key-pair or press [enter] to exit.> "
                 KeyName=key_name
             )
             key_file.write(key_pair.key_material)
-        chmod(key_filename, 0400)
+        if py3:
+            chmod(key_filename, 0o400)
+        else:
+            chmod(key_filename, 0400)
         self._lgr.info("Created private key file %s.", key_filename)
 
         # Save the new info to the resource.
--- a/repronim/environment/ec2environment.py
+++ b/repronim/environment/ec2environment.py
@@ -9,8 +9,10 @@
 """Environment sub-class to provide management of an AWS EC2 instance."""
 
 import boto3
+from sys import version_info, exit
+from os import makedirs, chmod, getcwd
+from os.path import expanduser, dirname, exists, isfile
 
-from repronim.resource import Resource
 from repronim.environment.base import Environment
 from repronim.support.sshconnector2 import SSHConnector2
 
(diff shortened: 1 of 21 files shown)


repronim/environment/ec2environment.py
```diff
@@ -153,9 +157,93 @@ def execute_command_buffer(self):
...
\ No newline at end of file
+                self.execute_command(ssh, command['command'], command['env'])
+
+    def create_key_pair(self):
+        """
+        Walk the user through creating an SSH key pair that is saved to
+        the AWS platform.
+        """
+
+        def get_user_input(prompt):
```
yarikoptic: I think we might benefit from centralizing user interactions ... actually in the "template" copy from datalad I have already copied `repronim.ui` submodule, so you could 

```python
from repronim.ui import ui
response = ui.question(prompt)
```
benefits -- centralization of ui handling so we get uniform appearance and also handling e.g. in the case of non-interactive sessions

repronim/environment/ec2environment.py
```diff
@@ -153,9 +157,93 @@ def execute_command_buffer(self):
...
+                exit(1)
+            else:
+                prompt = 'That key name exists already, try again. > '
+                key_name = get_user_input(prompt)
+
+        # Prompt for a path to store the SSH private key.
+        default_path = '{}/.ssh/{}.pem'.format(expanduser("~"), key_name)
+        for i in range(3):
+            prompt = 'Please enter the path to save the private key. [{}]> '.format(default_path)
+            key_filename = get_user_input(prompt)
```
yarikoptic: I think we shouldn't bother user with this information and just keep those under e.g.

```python
In [9]: os.path.join(AppDirs('repronim', 'repronim.org').user_data_dir, 'ec2_keys')
Out[9]: '/home/yoh/.local/share/repronim/ec2_keys'
```

repronim/environment/ec2environment.py
```diff
@@ -153,9 +157,93 @@ def execute_command_buffer(self):
...
+
+            if not key_filename:
+                key_filename = default_path
+            # Put .pem at end of private key file if not there already.
+            elif not key_filename.endswith('.pem'):
+                key_filename += '.pem'
+
+            if isfile(key_filename):
+                if i == 2:
+                    print('File exists, exiting.')
```
yarikoptic: when using that `ui` helper, use `ui.message` method

repronim/environment/ec2environment.py
```diff
@@ -153,9 +157,93 @@ def execute_command_buffer(self):
...
+            else:
+                break
+
+        # Make sure the directory for the private key file exists.
+        basedir = dirname(key_filename)
+        if not basedir or not basedir.startswith('/'):
+            basedir = getcwd()
+            key_filename = basedir + '/' + key_filename
+        if not exists(basedir):
+            makedirs(basedir)
```
yarikoptic: there is an `repronim.utils.assure_dir` helper

repronim/environment/ec2environment.py
yarikoptic: that would also simplify testing -- you could use `@with_testsui(responses=XXX)` decorator where XXX could be a single string or a list of responses given by the "user"

mjtravers:
I made the changes you requested to the key-pair code.

PR #49, opened 2017-01-21
Refactored resource classes.

mjtravers:
Yarik,

Here is the refactoring of the resources to simplify the architecture. A couple of interface classes were declared (Backend and Environment) in place of a 3-level hierarchy of class inheritance and the resource configuration was split out into its own class, ResourceConfig.

Matt


yarikoptic:
number of renames forbids adequate review so just merging ;)

PR #62, opened 2017-02-21
RF: Downgraded docker package

mjtravers:


yarikoptic:
I guess in setup.py you need to say now "docker-py<2" so correct one gets installed for testing

yarikoptic:
Tests pass on my laptop, so "we can work with this", thus merging!

PR #61, opened 2017-02-21
NF: trace files belonging to VCS

yarikoptic:
Sits on top of @rbuccigrossi #58 

TODOs
- [x] make it work at all
- [x] basic testing
- [x] sources (remotes) for git repos
- [x] record current branch as well where applicable
- [ ] DataLad datasets support (those are just 'git' repos, but in direct mode might not look as such ;) )
- [ ] more tests!  currently barely covered

It also
- [x] refactors retrace a bit so you can now just specify bunch of files on your filesystem to be 'traced' in addition to the spec -- really handy for interactive "investigation"

Future TODOs is to support
- [ ] hg 
- [ ] CVS
- [ ] bzr
and refactor code accordingly

Also possibly
- [ ] extend to annotate changed files (might we want even to store diffs???)

## Questions

- For VCS types -- should we may be list only relative path names in files? then full paths should be easily reconstructable with a simple rule `if not isabs(path): fullpath = opj(pkg['path'], path)`?
- For full paths under '$HOME' -- I wonder if we should do analysis right away and convert them to use that (e.g. to make them '$HOME/blah' instead of '/home/yoh/blah' in my case)... and then record 'run' session and associate with tracking one which would contain value for HOME?

Sample output:
```shell
(git)hopa:~/proj/repronim/niceman[nf-git-trace]
$> niceman --dbg retrace ~/proj/repronim/niceman/README.md /home/yoh/deb/debs/imagezoom-ru-svn/trunk.upgraded/debian/changelog /usr/bin/fail2ban-client                                                                         2017-02-27 20:23:16,003 [INFO   ] DpkgManager: 1 packages with 1 files, and 2 other files 
2017-02-27 20:23:16,238 [INFO   ] VCSManager: 2 packages with 2 files, and 0 other files 
# NICEMAN Environment Configuration File
# This file was created by NICEMAN 0.0.1 on 2017-02-27 20:23:16.239261

# Package Origins 

origins:
- {name: apt_Debian_testing_main_0, type: apt, origin: Debian, label: Debian, component: main,
  archive: testing, site: http.debian.net}
- {name: apt_Debian_unstable_main_0, type: apt, origin: Debian, label: Debian, component: main,
  archive: unstable, site: http.debian.net}
- {name: apt_NeuroDebian_sid_main_0, type: apt, origin: NeuroDebian, archive: sid,
  date: '2017-02-22 16:10:55+00:00', component: main, site: neuro.debian.net, label: NeuroDebian}
[...]

PR #63, opened 2017-02-22
BF: Use ArchiveURI to identify local repositories

### LINKED ISSUE #59: apt sources analysis assumes netloc
didn't check yet but we need to verify that we could handle/record `file://` url sources correctly, where `.netloc` could be empty.

on a quick check where I have generated a local repo with
```shell
mkdir /tmp/repo
cp some.deb /tmp/repo
cd /tmp/repo
dpkg-scanpackages . >| Packages
apt-ftparchive release . >| Release
echo 'deb file:///tmp/repo ./' > /etc/apt/sources.list.d/local.list
apt-get update
```
I have got:
```
lrwxrwxrwx 1 root root       20 Feb 20 13:09 _tmp_repo_._Packages -> /tmp/repo/./Packages
-rw-r--r-- 1 root root      816 Feb 20 13:08 _tmp_repo_._Release
```
so might be worth a quick test of some kind to check if new code in https://github.com/ReproNim/niceman/pull/58/files#diff-e6a1b8061859cc341e0dadc10ae5574dR288 handles it correctly

rbuccigrossi:
To address #59 , I use the lower level apt_pkg python library to pull the archive_uri associated with an origin. This allows me to more accurately create the release file name, and pull the release file for local repositories.

To test, I created a repository with the alienblaster package:

```
mkdir /my
mkdir /my/repo
cd /my/repo
wget http://mirrors.kernel.org/ubuntu/pool/universe/a/alienblaster/alienblaster_1.1.0-9_amd64.deb
dpkg-scanpackages . >| Packages
apt-ftparchive release . >| Release
echo 'deb file:///my/repo ./' > /etc/apt/sources.list.d/local.list
apt-get update
apt-get install alienblaster
```

The resulting origin entry is:

```
- {name: apt____0, type: apt, component: '', label: '', site: '', archive_uri: 'file:///my/repo/',
  archive: '', origin: '', date: '2017-02-21 02:42:30+00:00'}
```

Closes #59

### DIFF
--- a/niceman/retrace/packagemanagers.py
+++ b/niceman/retrace/packagemanagers.py
@@ -265,10 +265,11 @@ class DpkgManager(PackageManager):
                 if indexfile:
                     archive_uri = indexfile.archive_uri("")
                     origin["archive_uri"] = archive_uri
-                    # Get the release date
-                    rdate = self._find_release_date(archive_uri, pf.archive)
-                    if rdate:
-                        origin["date"] = rdate
+                # Get the release date
+                rdate = self._find_release_date(
+                    self._find_release_file(pf.filename))
+                if rdate:
+                    origin["date"] = rdate
                 # Now add our crafted origin to the list
                 origins.append(origin)
             v_info["origins"] = origins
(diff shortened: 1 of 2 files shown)


rbuccigrossi:
Hmmm... There is a "." in the release file name that I currently hard coded when the archive is empty. I realize that this probably came from the second parameter "deb file:///my/repo ./" in the local.list.  So to better test out local repositories, I need to create additional cases, like when there is a subdirectory that is used in the deb line of local.list.

So I marked this as WIP since it needs more work.

niceman/retrace/tests/test_packagemanagers.py
```diff
@@ -13,6 +13,7 @@
 
 def test_identify_packages():
     files = ["/usr/share/doc/xterm/copyright",
+             "/usr/games/alienblaster",
```
yarikoptic: so could the test now become an actual basic test (verify that if file is present -- package should be identified correctly etc), may be without even mocking (i.e. debian systems specific), not just a smoke run with prints? ;)

niceman/retrace/packagemanagers.py
```diff
@@ -273,29 +280,24 @@ def _create_package(self, pkgname):
...
-                                (site, archive))
-                        rfile = opj(dirname, name)
-
-        # Now extract and format the date from the release file
+        while "_" in rfprefix:
+            rfprefix = rfprefix.rsplit("_", 1)[0]
+            for ending in ['_InRelease', '_Release']:
+                if os.path.exists(rfprefix + ending):
+                    rfile = rfprefix + ending
+                    break
```
yarikoptic: I think here a bit of an odd and rarely used python feature could help to avoid duplication

```python
for ending in ['_InRelease', '_Release']:
    rfile = rfprefix + ending
    if os.path.exists(rfile):
        break
else:
     rfile = None
```

or if you want a really Pythonic one (never used such but stackoverflow came to help):
```python
return next((p for p in [rprefix+e for e in ['_InRelease', '_Release']] 
                    if os.path.exists(p)),
                   None)
```
which with a little bit of help becomes actually easy to read one liner (ignore import, needs to be adjusted to use six):
```python
from itertools import ifilter
return next(ifilter(os.path.exists, map(rprefix.__add__, ['_InRelease', '_Release'])), None)
```
;-)

niceman/retrace/packagemanagers.py
```diff
@@ -253,14 +253,21 @@ def _create_package(self, pkgname):
...
+                indexfile = v.package._pcache._list.find_index(pf)
+                if indexfile:
+                    archive_uri = indexfile.archive_uri("")
+                    origin["archive_uri"] = archive_uri
                 # Get the release date
-                rdate = self._find_release_date(o.site, o.archive)
+                rdate = self._find_release_date(
+                    self._find_release_file(pf.filename))
                 if rdate:
                     origin["date"] = rdate
```
yarikoptic: so could be just
```python
origin["date"] = self._find_release_date(self._find_release_file(pf.filename)) or None
```

rbuccigrossi: Here's another case where populating an object differs from my mental picture of this as a configuration file I'm generating.  I was avoiding populating "date" if it didn't exist so it wouldn't appear in the configuration file.  self._find_release_date returns "None" if not found, so the whole thing could simply become:

`origin["date"] = self._find_release_date(self._find_release_file(pf.filename))`

(Which I should just break down and do.  However, my hope is to create an Origin object next...)

yarikoptic: You are correct, I have misread... In my PR I had the same mental picture as you but didn't want to deal with every setting this way so came up with a helper to prune all empty ones at the end.. have a look, we should merge the efforts ;-)

rbuccigrossi:
I just added the origin architecture, which should resolve the double origin you experienced in #61 .  However, I still need to improve the test case.

PR #64, opened 2017-03-05
ENH: Added clone interface command and cleaned up resource code using…

mjtravers:
A "--clone" parameter was added to the "create" interface command so that an existing resource could be copied with a new name.

Also, the "attr" package was used to cleanup and streamline the niceman.resource packages.


yarikoptic:
@mjtravers  FYI: going through this PR code ... hopefully everything was pushed ;)

yarikoptic:
ok, tests pass so let's merge for now and improve later

PR #65, opened 2017-03-06
DOC: fix epilog from datalad ramification

yarikoptic:
stroke me when I was looking at output of `create --clone` ;)

better ideas for a "slogan"?


yarikoptic:
ok since no further comments -- will merge for now ;)

PR #74, opened 2017-03-16
NF/WIP: DpkgOrigin class as a drop-in replacement for origin dict

rbuccigrossi:
Right now I'm experimenting with a DpkgOrigin class (using attr) as a drop-in replacement for the origin dict.  The next things are to tackle its conversion to a HashableDict (an attr object is already hashable) and then an OrderedDict (which is really used for YAML conversion).


yarikoptic:
some conflict creeped in after I merged Matt's branch, please rebase or merge

PR #78, opened 2017-03-20
DOC: adjusted CONTRIBUTING to correspond to current state of requirements

yarikoptic:
@chaselgrove please verify if this works for you ;)  (seemed to work on clean debian jessie)


yarikoptic:
should get only better and I think it worked for me, thus merging

Issue #77, opened 2017-03-20
Store Codename with Origin for dpkg


rbuccigrossi:
Archive may point to links that change (such as testing), so store codename as well.

PR #79, opened 2017-03-22
NF: Address issue #77 by adding codename to origin

rbuccigrossi:
Address issue #77 by adding codename to origin

Note: This also updates the travis configuration to install python-apt 0.9.3.5, which works with Python 2 and 3.

TODOs
- [x] Add a test with a known file to ensure codename is being populated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,9 +52,7 @@ before_install:
 install:
   # all mess below to get an elderly python-apt installed on elderly precise within python env -- we must build it
   - sudo apt-get build-dep python-apt
-  # only for Python2 ATM
-  # - if [ ${TRAVIS_PYTHON_VERSION:0:1} = 2 ]; then sudo apt-get install xz-utils; curl -O http://archive.ubuntu.com/ubuntu/pool/main/p/python-apt/python-apt_0.9.3.5.tar.xz; xz -d python-apt_0.9.3.5.tar.xz; pip install http://launchpad.net/python-distutils-extra/trunk/2.28/+download/python-distutils-extra-2.28.tar.gz python-apt_0.9.3.5.tar; rm python-apt_0.9.3.5.tar; fi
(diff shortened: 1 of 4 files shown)


yarikoptic:
I guess at some point we should get some test for this functionality, right? ;)

rbuccigrossi:
What would be better, 1) mocking the apt library transaction, or 2) getting it to work on Travis and using a file that everyone should have?

And it appears that I have a dependency on a more recent apt module :/

rbuccigrossi:
Bah....  While I was able to get apt_1.1.0~beta5.tar up and running on a virtualenv in my box, on Travis I get the error unrecognized command line option '-std=c++11' while compiling.  So I'll have to explore that tomorrow.  

The older version of python-apt doesn't have codename in origin.  I could code around that, but I'd rather get newer versions up and running...

rbuccigrossi:
Alas, the upgrade to python-apt is not working, so please do not merge.

yarikoptic:
idea #1:  that sucks -- I am afraid making apt build might be like fighting with a windmill .  I would say:  make it conditional in the code -- if codename is not provided, just do not record it for now.
any reasonable system (travis's 12.04 is not reasonable any longer ;) ) will have more recent apt module available.  Known from which apt module version they started to report codename could help I guess.

idea #2: you are already accessing anyways Release file!  so you can just get the codename from there without relying on apt

rbuccigrossi:
Windmills is right.  I was able to get Travis to work with G++ version 4.8, but now I'm getting missing header files in the older version of libapt-pkg-dev.

One more thing to try is to find if a version between 0.8.3 and 1.1.0 supported codename...

rbuccigrossi:
Yay!  I got it working with apt 0.9.3.

So now I need to clean up, and actually build a test on a common file to make sure that these values are truly populated.

PR #84, opened 2017-05-04
Initial dockerpty setup

mjtravers:
This is not an actual pull request yet. Just sharing code.

The tty terminal replays all the commands run from a previous session when the container is restarted. Still figuring that one out.

To login to a docker container, first create it and then login (like so):

niceman create -r my-container -t docker-container
niceman login -r my-container

### DIFF
--- a/niceman/interface/login.py
+++ b/niceman/interface/login.py
@@ -20,16 +20,16 @@ from ..support.constraints import EnsureStr
 from ..resource import Resource
 
 from logging import getLogger
-lgr = getLogger('niceman.api.start')
+lgr = getLogger('niceman.api.delete')
 
 
-class Stop(Interface):
-    """Stop a computation environment
+class Login(Interface):
+    """Log into a computation environment
 
     Examples
     --------
 
-      $ niceman stop --resource=my-resource --config=niceman.cfg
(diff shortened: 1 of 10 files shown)


satra:
@mjtravers - that would be a very nice functionality

mjtravers:
Yarik,

I'm done doing all I think I am doing for this pull request and it is passing the tests.

Feel free to critique and/or reject and/or merge.

yarikoptic:
BTW, read the readme for dockerpty -- `C-p C-q` pretty much detaches from the session so if you hit enter, it would crash and exit but then would leave docker running.  not sure yet on how/either we should (ab)use that "feature" ;)

niceman/interface/login.py
```diff
@@ -0,0 +1,84 @@
...
+import re
+
+from .base import Interface, get_resource_info
+import niceman.interface.base # Needed for test patching
+from ..support.param import Parameter
+from ..support.constraints import EnsureStr
+from ..resource import Resource
+
+from logging import getLogger
+lgr = getLogger('niceman.api.delete')
```
yarikoptic: ok -- I guess we will fix all those logging targets later anyways

Issue #83, opened 2017-05-04
tracing files with utf-8 name


yarikoptic:
as @rbuccigrossi asked why `except UnicodeDecodeError as exc:` was added while handling `dpkg-query -S` call:  The cause was 
```shell
$> dpkg-query -S . | grep -5 Class_Gold
libdate-manip-perl: /usr/share/perl5/Date/Manip/Offset/off061.pm
tango-icon-theme: /usr/share/icons/Tango/24x24/apps/kuser.png
vlc-plugin-base:amd64: /usr/lib/x86_64-linux-gnu/vlc/lua/playlist/jamendo.luac
gimp-data: /usr/share/locale/nn/LC_MESSAGES/gimp20-tips.mo
calibre: /usr/share/calibre/content_server/stacktrace.js
ca-certificates: /usr/share/ca-certificates/mozilla/NetLock_Arany_=Class_Gold=_Főtanúsítvány.crt
texlive-fonts-extra: /usr/share/texlive/texmf-dist/fonts/tfm/public/fira/FiraSans-Light-osf-ts1.tfm
psychtoolbox-3-common: /usr/share/psychtoolbox-3/PsychOpenGL/MOGL/wrap/glFogfv.m
libghc-lens-dev: /usr/lib/haskell-packages/ghc/lib/x86_64-linux-ghc-8.0.1/lens-4.14-9dVzUyOAXSjIX3UgJ3NnoU/Control/Lens/Internal/Setter.hi
texlive-fonts-extra: /usr/share/texlive/texmf-dist/fonts/enc/dvips/libertine/lbtn_s5bq4i.enc
linux-headers-4.8.0-1-amd64: /usr/src/linux-headers-4.8.0-1-amd64/include/config/dma
```
well, obviously seeking dpkg-query output for '.' was dubious -- need to fix to use full path. 
But we indeed should fix for all the unicode handling here, since otherwise can't trace files with unicode:
```shell
$> niceman  retrace /usr/share/ca-certificates/mozilla/TÜBİTAK_UEKAE_Kök_Sertifika_Hizmet_Sağlayıcısı_-_Sürüm_3.crt   
2017-05-04 11:37:49,843 [WARNING] Got unicode decode problem. Happened within cates/mozilla/T��B��TAK_UEKAE_K. Decoding allowing for errors. 
2017-05-04 11:37:49,843 [INFO   ] DebTracer: 0 packages with 0 files, and 1 other files 
[...]

PR #85, opened 2017-05-17
BF: Fix for #83 to support Unicode filenames and package data

rbuccigrossi:
I still want to add unit tests for a couple of utility functions I added (safe_decode, safe_encode, and safe_write).

### DIFF
--- a/niceman/retrace/tests/test_packagemanagers.py
+++ b/niceman/retrace/tests/test_packagemanagers.py
@@ -1,5 +1,6 @@
 # emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
-# ex: set sts=4 ts=4 sw=4 noet:
+# -*- coding: utf-8 -*-
+#  ex: set sts=4 ts=4 sw=4 noet:
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 #
 #   See COPYING file distributed along with the niceman package for the
@@ -34,6 +35,17 @@ def test_identify_packages():
              "/usr/share/bug/vim/script",
              "/home/butch"]
     # Simple sanity check that the pipeline works
+    packages, origins, files = identify_packages(files)
+    pprint(files)
+    pprint(origins)
+    pprint(packages)
+    assert True
+
+
+@skip_if(not apt)
+def test_utf8_file():
+    files = [u"/usr/share/ca-certificates/mozilla/TÜBİTAK_UEKAE_Kök_Sertifika_Hizmet_Sağlayıcısı_-_Sürüm_3.crt"]
+    # Simple sanity check that the pipeline works with utf-8
     packages, origins, files = identify_packages(files)
     pprint(files)
     pprint(origins)
--- a/niceman/retrace/tests/test_packagemanagers.py
+++ b/niceman/retrace/tests/test_packagemanagers.py
@@ -1,6 +1,5 @@
-# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
-# -*- coding: utf-8 -*-
-#  ex: set sts=4 ts=4 sw=4 noet:
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil; coding: utf-8 -*-
+# ex: set sts=4 ts=4 sw=4 noet:
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 #
 #   See COPYING file distributed along with the niceman package for the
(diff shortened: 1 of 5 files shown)


niceman/utils.py
```diff
@@ -888,17 +890,51 @@ def _path_(p):
         return p
 
 
-def unicode(s):
+def is_unicode(s):
+    """Return true if an object is unicode"""
+    return  (six.PY3 and isinstance(s, str)) or \
+            (six.PY2 and isinstance(s, unicode))
```
yarikoptic: `isinstance(s, six.text_type)` ?

niceman/utils.py
```diff
@@ -888,17 +890,51 @@ def _path_(p):
...
-            return s
+    # If unicode, just return it
+    if is_unicode(s):
+        return s
+    try:
+        if six.PY3:
+            return str(s, encoding)
         else:
-            raise TypeError("Incorrect type for unicode()")
+            return __builtin__.unicode(s, encoding)
```
yarikoptic: how above block is different from `s.decode(encoding)` ?

niceman/utils.py
```diff
@@ -888,17 +890,51 @@ def _path_(p):
...
+
+
+def safe_write(os, s, encoding="utf-8"):
+    """safe_write safely writes different string types to an output stream"""
+    if PY3:
+        os.write(to_unicode(s))
     else:
-        return __builtin__.unicode(s)
+        try: # For PY2 try str first, then unicode
+            os.write(safe_encode(to_unicode(s, encoding), encoding))
```
yarikoptic: is it really worth doing `str->unicode->str` dance whenever a non-unicode regular py2 str is provided?

niceman/utils.py
```diff
@@ -888,17 +890,51 @@ def _path_(p):
...
+        except TypeError:
+            os.write(to_unicode(s, encoding))
```
yarikoptic: when/why above dance would fail but writing unicode would succeed?

niceman/retrace/tests/test_packagemanagers.py
```diff
@@ -42,6 +43,17 @@ def test_identify_packages():
...
 
 @skip_if(not apt)
+def test_utf8_file():
+    files = [u"/usr/share/ca-certificates/mozilla/TÜBİTAK_UEKAE_Kök_Sertifika_Hizmet_Sağlayıcısı_-_Sürüm_3.crt"]
+    # Simple sanity check that the pipeline works with utf-8
+    packages, origins, files = identify_packages(files)
+    pprint(files)
+    pprint(origins)
+    pprint(packages)
+    assert True
```
yarikoptic: could we actually check any output?  btw -- test probably should be skipped if ca-certificates package is not installed (or assert that we do not get some spurious result)

niceman/retrace/tests/test_packagemanagers.py
```diff
@@ -1,5 +1,6 @@
 # emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
-# ex: set sts=4 ts=4 sw=4 noet:
+# -*- coding: utf-8 -*-
```
yarikoptic: oh my wonderful mode line for emacs is gone! ;)  I do still use emacs from time to time so why not keep it and just extend with `; coding: utf-8` -- we already have that in few files

niceman/utils.py
rbuccigrossi: You're right!

niceman/utils.py
rbuccigrossi: Good question.  Something to look into.

niceman/utils.py
rbuccigrossi: No.  This is code version of the joke where a mathemetician and engineer are asked to boil water: https://www-users.cs.york.ac.uk/susan/joke/3.htm 

So far you are 3 for 3.

niceman/utils.py
rbuccigrossi: For once I have a reason!  I have a unit test (test_read_reprozip_yaml) that uses io.StringIO() as the output string.  In Python 2 the resulting object's write() expects unicode. So that will cause the first write to fail throwing a TypeError, which is caught and corrected with the second write.

niceman/retrace/tests/test_packagemanagers.py
rbuccigrossi: That can definitely be combined.

niceman/retrace/tests/test_packagemanagers.py
rbuccigrossi: If the file exists, then we should assume that it came from *some* package. So I can definitely add a test for a reasonable result.

niceman/utils.py
rbuccigrossi: (I should note that it isn't *complete* madness.  to_unicode() is designed to leave existing unicode alone, so this allowed a single line to handle the case when the source was unicode or ascii.  The way to avoid the full dance and keep it in one line is to instead write a to_ascii() instead. The result would provide a nice parallel to the to_unicode() and make the purpose of the TypeError exception more obvious.)

yarikoptic: I didn't call it madness ;-) just some waste of cycles ;-)

niceman/utils.py
yarikoptic: Oh! That is just wonderful. Thanks!

yarikoptic:
tried to merge it while on the  phone yesterday but smart github refused because coverage went down significantly... checked today -- apparently python-apt fails to build and thus we aren't really testing majority of code here now. So ideally we should have #86 fixed first to see if any side-effects within this one... didn't look in detail what travis has changed leading to failure to build python-apt.  You should be able to push to #86 (my branch on my clone) directly @rbuccigrossi

yarikoptic:
actually -- tests pass for me locally and conflicts already start to pile up so I will just merge for now

PR #86, opened 2017-05-18
BF(BK): moved installing of python-apt from source into a script to not mask failures to build it

yarikoptic:
By default in bash, exit status of the sequence of commands separated with ; is as of the last
command executed.  So if any of then fail before then, shell would still report "all is ok", unless

1. you set -e   setting for the session (then would fail as soon as any command fails)
2. join them with  & instead of ;

Since our line got too long already, I just moved that installation into a script which should fail
if any step fails (see set -eux at the top)

Detected when saw decreased coverage within https://github.com/ReproNim/niceman/pull/85 which should havenot changed coverage that drastically

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # vim ft=yaml
-# travis-ci.org definition for DataLad build
+# travis-ci.org definition for NICEMAN build
 language: python
 
 cache:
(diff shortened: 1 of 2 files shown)


yarikoptic:
yeap -- now failed reliably with
```
    python/acquire.cc:358:5: warning: deprecated conversion from string constant to ‘char*’ [-Wwrite-strings]

    error: command 'gcc' failed with exit status 1
```

yarikoptic:
ok -- I guess they also changed how they clone the repo so vcs.py gets also less coverage.. but that is ok for now. merging

PR #87, opened 2017-05-18
BF: Python-apt working under Travis & support for attrs 17.1.0

rbuccigrossi:
Clearly the purpose of this branch changed. After figuring out what was different between our environments and travis' docker images, I finally got python-apt working (and a quick fix to support attrs 17.1.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,9 @@ before_install:
 
 install:
   # all mess below to get an elderly python-apt installed on elderly precise within python env -- we must build it
+  - pip list
   - sudo tools/ci/build_install_apt
+  - pip list
   # See if apt is installed
   - python -c "import apt"
   - cd ..; pip install -q codecov; cd -
@@ -74,7 +76,6 @@ script:
   - if [ ! -z "$NICEMAN_TESTS_NONETWORK" ]; then sudo route add -net 0.0.0.0 netmask 0.0.0.0 dev lo; fi
   # - $NOSE_WRAPPER `which nosetests` -s -v --with-doctest --doctest-tests --with-cov --cover-package niceman --logging-level=INFO
   # See if apt is installed again
-  - python -c "import apt"
   - coverage run `which py.test` niceman
   - if [ ! -z "$NICEMAN_TESTS_NONETWORK" ]; then sudo route del -net 0.0.0.0 netmask 0.0.0.0 dev lo; fi
   # Generate documentation and run doctests
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ before_install:
 
 install:
   # all mess below to get an elderly python-apt installed on elderly precise within python env -- we must build it
-  - bash tools/ci/build_install_apt
+  - tools/ci/build_install_apt
   - cd ..; pip install -q codecov; cd -
   - pip install --upgrade pip
   # needed by html5lib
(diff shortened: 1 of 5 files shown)


yarikoptic:
Might be worth keeping ultimate discovery .travis.yml handy somewhere under tools/ci so we could just copy it when needed to gather info

rbuccigrossi:
@yarikoptic Please take a look. I think I finally have python-apt working on travis' docker images!

.travis.yml
```diff
@@ -51,7 +51,7 @@ before_install:
 
 install:
   # all mess below to get an elderly python-apt installed on elderly precise within python env -- we must build it
-  - sudo tools/ci/build_install_apt
+  - bash tools/ci/build_install_apt
```
yarikoptic: FWIW -- there should be no need for explicit `bash` here -- file has shebang and is executable

niceman/retrace/packagemanagers.py
```diff
@@ -153,7 +153,7 @@ def _create_package(self, pkgname):
         raise NotImplementedError
 
 
-@attr.s
+@attr.s(cmp=True, hash=True, frozen=True)
```
yarikoptic: from which version of apt it is ?  adjust setup.py accordingly please

niceman/retrace/packagemanagers.py
```diff
@@ -215,8 +215,16 @@ def _create_named_origin(o, used_names):
...
+        new_o = DpkgOrigin(name=name,
+                           component=o.component,
+                           codename=o.codename,
+                           archive=o.archive,
+                           architecture=o.architecture,
+                           origin=o.origin,
+                           label=o.label,
+                           site=o.site,
+                           archive_uri=o.archive_uri,
+                           date=o.date)
```
yarikoptic: there should be a better way ;)  e.g. if you go through all _attrs, compile a dict (e.g. params) with adjusted name and pass into `DpkgOrigin(**params)`

niceman/retrace/tests/test_packagemanagers.py
```diff
@@ -80,7 +80,7 @@ def test_dpkg_manager_identify_packages():
             assert o.label, "A non-local origin needs a label"
             assert o.site, "A non-local origin needs a site"
             assert o.archive_uri, "An archive_uri should have been found"
-            assert o.date, "An package should have been found"
```
yarikoptic: irrelevant to this PR but "duplication is evil".  use `.__attrs_attrs__` and `getattr` to get them in a loop ;)

niceman/retrace/packagemanagers.py
rbuccigrossi: Frozen was introduced in 16.1.0 of attrs, so I'll add that dependency,

.travis.yml
rbuccigrossi: Hey, look at that: git kindly maintains executable status! Nice to know.

niceman/retrace/tests/test_packagemanagers.py
rbuccigrossi: Well, I still want to explicitly list the mandatory attributes, so let's see if you like the result.

yarikoptic: sure... then
```python
for mandatory_attribute in ['date', ...]:
    assert getattr(o, mandatory_attribute), "Value for %s is missing" % mandatory_attribute
```

PR #93, opened 2017-07-05
ENH: Moved resource specific create parameters to resource Class.

mjtravers:
The --help output is not formatted very well.

niceman create --help

Otherwise, working well. The change was to add a --backend ArgParse command which takes multiple resource-specific config parmaters

 e.g.

niceman create -r test1 -t docker-container --backend base_image_id=ubuntu:latest

### DIFF
--- a/niceman/interface/create.py
+++ b/niceman/interface/create.py
@@ -11,6 +11,7 @@
 
 __docformat__ = 'restructuredtext'
 
+from importlib import import_module
 from .base import Interface
 import niceman.interface.base # Needed for test patching
 # from ..provenance import Provenance
@@ -20,6 +21,29 @@ from ..resource import ResourceManager
 
 from logging import getLogger
 lgr = getLogger('niceman.api.create')
+
+def backend_help(resource_type=None):
+    types = ResourceManager._discover_types()
+
+    help_message = "One or more backend parameters in the form KEY=VALUE. Options are: "
+    help_args = []
+
+    for module_name in types:
+        class_name = ''.join([token.capitalize() for token in module_name.split('_')])
+        try:
+            module = import_module('niceman.resource.{}'.format(module_name))
+        except ImportError as exc:
+            raise ResourceError(
+                "Failed to import resource: {}.  Known ones are: {}".format(
+                    exc_str(exc),
+                    ', '.join(ResourceManager._discover_types()))
+            )
+        cls = getattr(module, class_name)
+        args = cls.get_backend_properties()
+        for k, v in args.items():
+            help_args.append('"{}" ({})'.format(k, v))
+
+    return help_message + ", ".join(help_args)
 
 
 class Create(Interface):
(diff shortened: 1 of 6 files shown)


niceman/resource/aws_ec2.py
```diff
@@ -58,6 +58,26 @@ class AwsEc2(Resource):
...
+    def get_backend_properties():
+        """
+        Return the config properties specific to this resource type.
+
+        Returns
+        -------
+        dict : key = resource property, value = property description
+        """
+        return {
+            'access_key_id': 'AWS access key for remote access to your Amazon subscription.',
```
yarikoptic: IMHO those better live along within `attr.ib` definition, i.e. within its metadata field.  I think we might even benefit from
providing a tiny adapter
```python
def attrib(*args, **kwargs):
   doc = kwargs.get('doc')
   metadata = kwargs.get('metadata', {})
   if doc:
        metadata['doc'] = doc
   if metadata:
        kwargs['metadata'] = metadata
   return attr.ib(*args, **kwargs)
```

then we could do 
```python
access_key_id = attrib(
   default=None,
   doc="AWS secret access key for remote access to your Amazon subscription")
```

mjtravers: Oh, I like that. Okay, I'll refactor.

mjtravers:
Yarik,

Suggested updates have been made to the code. Ready for second code review.

Matt

niceman/resource/base.py
```diff
@@ -27,12 +28,18 @@
...
-]
+def attrib(*args, **kwargs):
+    """
+    Extend the attrs decorator to include a doc metadata element.
+    """
+    doc = kwargs.get('doc')
+    metadata = kwargs.get('metadata', {})
+    if doc:
+        metadata['doc'] = doc
+        del kwargs['doc']
```
yarikoptic: Nit picking: You could just .pop it (instead of get) I believe giving it a default None this avoiding del

niceman/resource/shell.py
```diff
@@ -28,6 +28,17 @@ class Shell(Resource):
...
+    def get_backend_properties():
+        """
+        Return the config properties specific to this resource type.
+
+        Returns
+        -------
+        dict : key = resource property, value = property description
+        """
+        return {}
+
```
yarikoptic: Is this one still needed?

mjtravers:
I did not know about .pop(). Very nice! Ready for code review.

niceman/interface/create.py
```diff
@@ -73,69 +99,26 @@ class Create(Interface):
...
-            doc="Path to SSH private key file matched with AWS key name parameter.",
-            constraints=EnsureStr(),
-        ),
         existing=Parameter(
             args=("-e", "--existing"),
             choices=("fail", "redefine"),
             doc="Action to take if name is already known"
         ),
+        backend=Parameter(
+            args=("-b", "--backend"),
```
yarikoptic: I think we don't really need it as an additional option and "create" could just take then as arguments (just avoid typing `--backend` pretty much).  but ok for now -- we will RF later.

Cheers

PR #97, opened 2017-07-24
WiP: Making sessions to be delegated by the resources and used by retrace etc

yarikoptic:


yarikoptic:
BTW @mjtravers -- pushed state is already quite sane.  I think all failing tests are due to may be now failing assumptions for mocking of execution of those 'add_command' items ... if you would be working on niceman some time in upcoming days -- I would appreciate if you have a look ;-)

rbuccigrossi:
The role and impact of session on retrace and distributions makes sense to me. Once this is merged, I should pull it into my branch and update the conda distribution code to use the session as well.

yarikoptic:
@mjtravers yeap... "delegation" -- this way, different backends might reuse the same session constructs more flexibly (the same backend providing ssh non-interactive, ssh pty, or some other session depending on parametrization) , which wouldn't be possible to achieve with subclassing etc

PR #98, opened 2017-07-24
RF(BK): Manager RF to be an instance with get/create methods

yarikoptic:


tools/ci/build_install_apt
```diff
@@ -7,7 +7,7 @@ set -eux
 ver=0.9.3.5
 
 sudo apt-get build-dep python-apt
-sudo apt-get install -y --allow-downgrades  libapt-pkg-dev=0.8.16~exp12ubuntu10.27
+sudo apt-get install -y --allow-downgrades  libapt-pkg-dev=0.8.16~exp12ubuntu10.27 libapt-{inst1.5,pkg4.12}==1.0.1ubuntu2
```
rbuccigrossi: Looking at the job: https://travis-ci.org/ReproNim/niceman/jobs/274655354 it looks like you're doing your libapt attempts on Precise still.  Would this line work under Trusty?

yarikoptic: hahaha -- got lost in the branches... was intended to add this to #112 

THANKS!

yarikoptic:
will be superseded by @kyleam 's PR

PR #100, opened 2017-08-04
ENH: Added StarCluster files.

mjtravers:
Added un-modified StarCluster files to niceman/support/starcluster


yarikoptic:
send it against branch 3rd -- now you are suggesting to merge them straight into master.  and in general can just push that updated 3rd branch directly -- no need for PR since we talked about it and there is nothing really to review etc

mjtravers:
I pushed the StarCluster code to the 3rd branch directly.

I though I had requested a pull request into the 3rd branch and not master. Apparently not.

Closing this request.

PR #101, opened 2017-08-08
ENH: Added StarCluster sshutils so users can log into AWS EC2 instances.

mjtravers:
Pull request for SSHing into AWS EC2 instance using StarCluster utils.

### DIFF
--- a/niceman/interface/create.py
+++ b/niceman/interface/create.py
@@ -11,9 +11,7 @@
 
 __docformat__ = 'restructuredtext'
 
-import attr
-from importlib import import_module
-from .base import Interface
+from .base import Interface, backend_help, backend_set_config
 import niceman.interface.base # Needed for test patching
 # from ..formats import Provenance
 from ..support.param import Parameter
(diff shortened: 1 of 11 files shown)


niceman/interface/create.py
```diff
@@ -172,13 +144,8 @@ def __call__(resource, resource_type, config, resource_id, clone, only_env,
...
         # Set resource properties to any backend specific command line arguments.
-        for backend_arg in backend:
-            key, value = backend_arg.split("=")
-            if hasattr(env_resource, key):
-                config[key] = value
-                setattr(env_resource, key, value)
-            else:
-                raise NotImplementedError("Bad --backend paramenter '{}'".format(key))
+        if backend:
+            config = backend_set_config(backend, env_resource, config)
```
yarikoptic: imho better to have a function without 'side effect' of modifying config in place and then also returning it...  what if 

`backend_set_config` -> `get_backend_config(backend, env_resource)` which returns a backend_config, and then you just do  `config.update(backend_config)`, i.e. update with those new settings.  IMHO makes it clearer

niceman/interface/login.py
```diff
@@ -72,10 +77,15 @@ def __call__(resource, resource_id=None, config=None):
...
-        resource_info, inventory = ResourceManager.get_resource_info(config, resource, resource_id)
+        config, inventory = ResourceManager.get_resource_info(config, resource, resource_id)
 
         # Connect to resource environment
-        env_resource = ResourceManager.factory(resource_info)
+        env_resource = ResourceManager.factory(config)
+
+        # Set resource properties to any backend specific command line arguments.
+        if backend:
+            config = backend_set_config(backend, env_resource, config)
```
yarikoptic: same applies as above... I feel that it all might become also shorter after we look into making ResourceManager  a proper class etc, but that later

niceman/resource/aws_ec2.py
```diff
@@ -297,4 +300,13 @@ def create_key_pair(self, key_name=None):
...
+        self.key_filename = key_filename
+
+    def login(self):
+        """
+        Log into a container and get the command line
+        """
+        lgr.debug("Opening TTY connection to AWS EC2 instance.")
+        host = self._ec2_instance.public_ip_address
+        ssh = SSHClient(host, private_key=self.key_filename)
+        ssh.interactive_shell(self.user)
```
yarikoptic: could we also get a simple "SSHHost" resource or smth like that so we could register/login to remote hosts (which do not need startup etc) general?

niceman/support/starcluster/sshutils.py
```diff
@@ -517,10 +517,10 @@ def _get_output(self, channel, silent=True, only_printable=False):
                     line = ''.join(c for c in line if c in string.printable)
                 if line != '':
                     output.append(line)
-                    print line,
+                    print(line),
```
yarikoptic: oh -- starcluster is not python3 compatible?  we are asking for trouble I am afraid... we might want to just futurize its source code or smth like that at least...

yarikoptic:
Ok @mjtravers -- I will merge it so I could merge these changes with my branch as well since tests pass already, and then we could address those comments I have left.

PR #108, opened 2017-09-02
BF: Get libapt build working on Travis

rbuccigrossi:
Starting August 1st, Travis defaults to using Ubuntu 14.04 instead of 12.04 for docker images. This broke our libapt build script.  Until we remove our dependency on python-apt, as a quick fix, I updated .travis.yml to use 12.04 again.


yarikoptic:
I guess we could do that until we get rid of apt module dependency. Although I am not sure how tests did pass for my PR even just recently. Thanks

PR #111, opened 2017-09-07
RF: use default (trusty) distro, and install python-apt from the distro

yarikoptic:


yarikoptic:
```
Get:54 https://packagecloud.io/computology/apt-backport/ubuntu trusty/main amd64 libapt-pkg-dev amd64 1.2.10 [222 kB]
```
grrr.... that is what I don't like travis for -- their environment is a mess of various additional apt repos they pick up from what seems to be random locations, so their environment is quite far from the stock ubuntu release...

PR #113, opened 2017-09-19
WiP ENH: try to come up with re-based parsing of apt-cache policy pkg output

yarikoptic:
@rbuccigrossi  apparently it is kinda 'difficult' ;-)  I guess I still missing on some aspects of re flags etc, since what I think was obviously correct regexp didn't work for me.  This one gets close'ish but yet not complete


rbuccigrossi:
I merged this into my WIP branch.  Thanks!

PR #114, opened 2017-09-19
DOC: updated to reflect new testing (not nose)

chaselgrove:
Updated CONTRIBUTING.md.  The Coverage section also lost some content with the move from nose.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,7 +132,7 @@ and additionally, for development we suggest to use tox and new
 versions of dependencies from pypi:
 
 ```sh
-eatmydata apt-get install -y -q python-{nose,pip,vcr,virtualenv,tox}
+eatmydata apt-get install -y -q python-{pip,vcr,virtualenv,tox}
 ```
 
 some of which you could also install from PyPi using pip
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -132,7 +132,7 @@ and additionally, for development we suggest to use tox and new
 versions of dependencies from pypi:
 
 ```sh
-eatmydata apt-get install -y -q python-{nose,pip,vcr,virtualenv,tox}
+eatmydata apt-get install -y -q python-{pip,vcr,virtualenv,tox}
```
yarikoptic: although we use pytest to run the tests, some tests still (just do `git grep nose | grep import`) use nose's functions, so nose needs to be available

CONTRIBUTING.md
```diff
@@ -239,13 +239,7 @@ python setup.py develop
...
 
 ```sh
-python -m nose -s -v niceman
-```
-
-or similarly,
-
-```sh
-nosetests -s -v niceman
+niceman test
```
yarikoptic: probably better leave original version, just replace `-m nose` with `-m pytest` and `nosetests` with `pytest`.  `niceman test` doesn't expose all handy options (-s -v --pdb etc) so better to use pytest runners directly

chaselgrove:
Updated.

We still need a general reworking of the intro docs (perhaps by someone who's about to be wrestling with them. :) )

PR #116, opened 2017-10-05
NF: distribution operations

chaselgrove:
Primitives for checking if Debian packages and distributions satisfy requirements in other packages and distributions.


niceman/distributions/debian.py
```diff
@@ -179,6 +194,30 @@ def normalize(self):
...
+        """return True if this distribution (self) satisfies the requirements 
+        of the other distribution (other)"""
+        if not isinstance(other, Distribution):
+            raise TypeError('satisfies() requires a package argument')
+        if not isinstance(other, DebianDistribution):
+            return False
+        for p in other.packages:
+            if not self.satisfies_package(p):
+                return False
+        return True
```
yarikoptic: above four lines could be more pythonicly expressed
```python
return all(map(self.satisfies_package, other.packages))
```
in python3 map is already a generator (so you would get early termination), in python2 -- not yet so it would go through all possibly for no reason.  So just 
```python
from six.moves import map
```

niceman/distributions/debian.py
```diff
@@ -179,6 +194,30 @@ def normalize(self):
...
+        """return True if this distribution (self) satisfies the requirements 
+        of the passed package"""
+        if not isinstance(package, Package):
+            raise TypeError('satisfies_package() requires a package argument')
+        if not isinstance(package, DEBPackage):
+            return False
+        for p in self.packages:
+            if p.satisfies(package):
+                return True
+        return False
```
yarikoptic: see comment below ... eventually we might want to make it more efficient/avoid full list lookup every time.

niceman/distributions/tests/test_debian.py
```diff
@@ -136,4 +140,47 @@ def _run_dpkg_query(subfiles):
...
+    """set up the package comparison tests"""
+    a = DEBPackage(name='p1')
+    b = DEBPackage(name='p1', version='1.0')
+    c = DEBPackage(name='p1', version='1.1')
+    d = DEBPackage(name='p2')
+    return (a, b, c, d)
+
+def test_package_satisfies(setup_packages):
+    (p1, p1v10, p1v11, p2) = setup_packages
+    assert p1.satisfies(p1) is True
```
yarikoptic: sufficient to
```python
assert p1.satisfies(p1)
assert not p1.satisfies(p1v10)
```

chaselgrove: I'd set it up this way so I could return None and make all tests fail to start with.  You're right, at this point this idiom isn't needed any more.

yarikoptic:
I added a WiP label. @chaselgrove please remove it when you think it is worth merging

chaselgrove:
If we can merge #121, I'd like to refactor this PR to take advantage of that code.  Then let's merge this -- this could potentially be a WIP until _all_ the classes are fleshed out (read: indefinitely :) ) but I'd rather do this iteratively.

PR #121, opened 2017-10-13
NF: immutable packages

chaselgrove:
This PR makes DEBPackage immutable so they are hashable (and usable in sets -- my ultimate motivation :) ).


niceman/distributions/debian.py
```diff
@@ -282,8 +282,7 @@ def _create_package(self, name, architecture=None):
         except KeyError:  # Package not found
             lgr.warning("Was asked to create a spec for package %s but it was not found in apt", name)
             return None
-
-        # prep our pkg object:
+# prep our pkg object:
```
yarikoptic: why dedent?

niceman/distributions/debian.py
```diff
@@ -89,26 +89,26 @@ class APTSource(SpecObject):
...
-    version = attr.ib(default=None)
-    architecture = attr.ib(default=None)
-    size = attr.ib(default=None)
-    md5 = attr.ib(default=None)
-    sha1 = attr.ib(default=None)
-    sha256 = attr.ib(default=None)
-    source_version = attr.ib(default=None)
-    versions = attr.ib(default=None)
-    install_date = attr.ib(default=None)
+    version = attr.ib(default=None, hash=True)
```
yarikoptic: it is hard(er) to differentiate between those with True or False.  The default for `hash` seems to be True, so I would prefer if `, hash=True` ones were removed, then it would be easier to see which fields we consider not important

niceman/distributions/debian.py
```diff
@@ -89,26 +89,26 @@ class APTSource(SpecObject):
...
+    architecture = attr.ib(default=None, hash=False)
```
yarikoptic: there could be multiple instances of the package with the same name and version but different architecture
```shell
$> dpkg -l zlib1g:*
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                 Version         Architecture    Description
+++-====================-===============-===============-==============================================
ii  zlib1g:amd64         1:1.2.8.dfsg-5  amd64           compression library - runtime
ii  zlib1g:i386          1:1.2.8.dfsg-5  i386            compression library - runtime
```
so should participate in the hashing

yarikoptic:
Also note that you would conflict in changes with #115, so might be better to wait for that one to get merged first

rbuccigrossi:
I merged #115 so this can proceed (though I will make a new pull request to resolve some review suggestions).  GitHub claims that there are still no conflicts in this branch...

rbuccigrossi:
(I spoke too soon...  as soon as I submitted the comment github noted some conflicts in this branch...  sorry)

chaselgrove:
Updated in response to comments.

yarikoptic:
conflicts should be resolved before itcould be considered for merging

yarikoptic:
@chaselgrove ping

yarikoptic:
@chaselgrove ping ping

chaselgrove:
Fixed conflicts.

PR #99, opened 2017-07-27
ENH:  Conda Package Detection

rbuccigrossi:
Status:

- Can identify conda environments and root paths
- Associates files with packages (conda and pip)
- Conducts a test against a conda environment installed in an AppDir

PR #125, opened 2017-10-18
ENH: Resolving code review suggestions for conda tracing

rbuccigrossi:
This pull request addresses the code review comments and suggestions from pull request #99 .

### DIFF
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -193,11 +193,7 @@ class CondaTracer(DistributionTracer):
                 pip_deps = dep.get("pip")
 
         for pip_dep in pip_deps:
-            # Pip packages are recorded in conda exports as "name (loc)",
-            # "name==version" or "name (loc)==version".  So split on "=", then
-            # on " "
-            name = pip_dep.split("=")[0]
-            name = name.split(" ")[0]
+            name, origin_location = self.parse_pip_package_entry(pip_dep)
             try:
                 out, err = self._session.execute_command(
                     '%s/bin/pip show -f %s'
@@ -207,12 +203,9 @@ class CondaTracer(DistributionTracer):
                 # Record the details we care about
                 details = {"name": pip_info.get("Name"),
                            "version": pip_info.get("Version"),
-                           "installer": "pip"}
+                           "installer": "pip",
+                           "origin_location": origin_location}
                 packages[pip_dep] = details
-                # Record the origin location (if installed from a local source)
-                if "(" in pip_dep:   # We have an origin location
-                    details["origin_location"] = \
-                        re.search('\(([^)]+)', pip_dep).group(1)
                 # Map the package files to the package
                 for f in pip_info.get("Files"):
                     full_path = os.path.normpath(
@@ -225,6 +218,20 @@ class CondaTracer(DistributionTracer):
                 continue
 
         return packages, file_to_package_map
+
+    @staticmethod
+    def parse_pip_package_entry(pip_dep):
+        # Pip packages are recorded in conda exports as "name (loc)",
+        # "name==version" or "name (loc)==version".  So split on "=", then
+        # on " "
+        name = pip_dep.split("=")[0]
+        name = name.split(" ")[0]
+        # Record the origin location (if installed from a local source)
+        if "(" in pip_dep:  # We have an origin location
+            origin_location = re.search('\(([^)]+)', pip_dep).group(1)
+        else:
+            origin_location = None
+        return name, origin_location
 
     def _get_conda_env_export(self, root_prefix, conda_path):
         export = {}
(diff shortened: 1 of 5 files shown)


rbuccigrossi:
I believe I addressed all of the code review comments and suggestions from #99, so this is ready for review and merging.

niceman/distributions/conda.py
```diff
@@ -207,6 +209,10 @@ def _get_conda_pip_package_details(self, env_export, conda_path):
                            "version": pip_info.get("Version"),
                            "installer": "pip"}
                 packages[pip_dep] = details
+                # Record the origin location (if installed from a local source)
+                if "(" in pip_dep:   # We have an origin location
+                    details["origin_location"] = \
```
yarikoptic: could we get some basic test for this one?  For an integration test I wonder what would be the most lightweight way... may be we need some minimalistic `niceman/tests/files/samplemod` with some very basic setup.py, and then have it `pip install -e `'ed by conda?  
But I guess we don't need that and you could just test this specific method by feeding it different inputs, and then just mocking those few (if more than one) call to execute_command for pip show

niceman/distributions/tests/test_conda.py
```diff
@@ -8,45 +8,56 @@
...
-def test_conda_manager_identify_distributions():
-    dirs = AppDirs('niceman')
-    test_dir = os.path.join(dirs.user_cache_dir, 'conda_test')
-    create_test_conda(test_dir)
-    print (test_dir + "\n")
+def test_conda_manager_identify_distributions(get_conda_test_dir):
+    # Skip if network is not available (skip_if_no_network fails with fixtures)
+    test_dir = get_conda_test_dir
+    if not test_dir:
+        raise SkipTest("Skipping since no network settings")
```
yarikoptic: ha, apparently `skip_if_no_network` is smart enough so you could just use it as a function inside (just call it within the test and it would raise SkipTest if no network setting)

niceman/tests/utils.py
```diff
@@ -544,13 +544,29 @@ def newfunc(*args, **kwargs):
...
+            found = False
+            for b_val in b:
+                try:
+                    assert_is_subset_dict_recur(a_val, b_val)
+                    found = True
+                    break
+                except AssertionError:
+                    pass
+            if not found:
+                raise AssertionError("Array value %s is missing" % a_val)
```
yarikoptic: FYI, Python has 'else' for loop which is triggered if there were no 'break':
```
$> python /tmp/l.py
loop without break
1
2
else: finished, no break

loop with break
1

loop without looping
else: we do get here
2 14642.....................................:Thu 26 Oct 2017 09:44:34 AM EDT:.
hopa:~
$> cat /tmp/l.py

print("loop without break")
for i in [1,2]:
    print i
else:
    print "else: finished, no break"

print("\nloop with break")
for i in [1,2]:
    print i
    break
else:
    raise RuntimeError("must not get here since we break")

print("\nloop without looping")
for i in []:
    print i
else:
    print "else: we do get here"
```

so above could be
```python
   for b_val in b:
          try:
                 assert_is_subset_dict_recur(a_val, b_val)
                 break
          except AssertionError:
                 pass
     else:
          raise AssertionError("Array value %s is missing" % a_val)
```
right?

yarikoptic: now this assert functioning wouldn't also match its name.  I wondered if we should keep an old one as is and then add a new extended one? or parametrize it to do this `list` subset matching optionally?  in some functions (those which used it insofar I guess) we might like to have its more strict operation, just really matching subsets of dict key:value pairs while otherwise needing lists to be precisely the same

yarikoptic: BTW, lovely otherwise ;)

yarikoptic: suggested API
```python
assert_is_subset_recur(a, b, dicts=True, lists=False)
```

yarikoptic: Just a Python mindblow:  If we had it as a function
```
    from functools import partial
    if not any(map(partial(is_subset_dict_recur, a_val), b)):
         raise AssertionError("Array value %s is missing" % a_val)
```

niceman/distributions/conda.py
rbuccigrossi: Done: 2e1320f

niceman/distributions/tests/test_conda.py
rbuccigrossi: Done 2e1320f

niceman/tests/utils.py
rbuccigrossi: Addressed in 9a93c0c

niceman/distributions/conda.py
rbuccigrossi: I isolated the parsing into a static function so that I can better test it ( 9a93c0c )

Issue #123, opened 2017-10-16
Mac OSX client side support


yarikoptic:
TODOs:
- [ ] travis supports OSX (https://docs.travis-ci.com/user/reference/osx/)
- [ ] tests should pass or be properly skipped on OSX

chaselgrove:
No osx+python in travis!  See https://github.com/travis-ci/travis-ci/issues/2312.

PR #126 to address this issue.

yarikoptic:
re absent python -- use miniconda or virtualenv with custom python installation... see/adopt how e.g skimage did it: https://github.com/scikit-image/scikit-image/blob/master/.travis.yml#L55

yarikoptic:
@chaselgrove could you please look into github workflows -- https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions says that `macos-latest` corresponds to `macOS Catalina 10.15`, so it should generally be possible.
I have added it to the matrix of runs for dandi-client and it seems to work, and python is available there! (dandi tests fail there - but that is orthogonal issue probably) so have a look at https://github.com/dandi/dandi-cli/blob/ca8f6511e705361f255074bb76ae73a81cb467d4/.github/workflows/test.yml as a simple example on how to establish github workflow.  I already see "Actions" tab in our repo, so most likely it would just start working after a PR gets submitted

PR #126, opened 2017-10-19
ENH: OSX support

chaselgrove:
Addresses #123.

Installation works for me following CONTRIBUTING.

Removed debian tests.

Travis does not support python on osx; see https://github.com/travis-ci/travis-ci/issues/2312

### DIFF
--- a/niceman/distributions/tests/test_debian.py
+++ b/niceman/distributions/tests/test_debian.py
@@ -15,7 +15,10 @@ from niceman.distributions.debian import DebTracer
 
 import mock
 
+from ...tests.utils import skip_if_on_osx
 
+
+@skip_if_on_osx
 def test_dpkg_manager_identify_packages():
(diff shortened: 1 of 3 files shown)


niceman/distributions/tests/test_debian.py
```diff
@@ -15,7 +15,10 @@
 
 import mock
 
+from ...tests.utils import skip_if_on_osx
 
+
+@skip_if_on_osx
```
yarikoptic: should better be 

`@skip_if(not on_debian)`

this way we could run tests on centos etc

niceman/tests/utils.py
```diff
@@ -349,6 +349,17 @@ def newfunc(*args, **kwargs):
     return newfunc
 
 
+def skip_if_on_osx(func):
```
yarikoptic: probably no need in a dedicated decorator -- see above

PR #127, opened 2017-10-26
ENH: get/put for Docker containers

mjtravers:
On this branch:

1. Added get/put to docker container
2. Added handling of command exit status for docker containers
3. Added testing code for resource classes


niceman/resource/session.py
```diff
@@ -195,18 +197,16 @@ def niceman_exec(self, command, args):
     @abc.abstractmethod
     def exists(self, path):
         """Return if file exists"""
-        pass
+        raise NotImplementedError
 
     @abc.abstractmethod
-    def put(self, src_path, dest_path, preserve_perms=False,
-                owner=None, group=None, recursive=False):
+    def put(self, src_path, dest_path, owner=None, group=None):
```
yarikoptic: So will all of them operate rrcursively?

niceman/resource/session.py
```diff
@@ -385,9 +397,38 @@ def isdir(self, path):
...
+        """
+        try:
+            command = ['chmod']
+            if recursive: command += ["-R"]
+            command += [mode] + [remote_path]
+            self.execute_command(command)
+            return True
+        except CommandError as cmd:
+            lgr.error(cmd.msg)
+            return False
```
yarikoptic: Why catch errors? It isn't  a `issomething` function, so I'm hoping should throw exception if something goes wrong

niceman/resource/session.py
```diff
@@ -385,9 +397,38 @@ def isdir(self, path):
...
+            if recursive: command += ["-R"]
+            if int(uid) > 0 and int(gid) > 0: command += ["{}.{}".format(uid, gid)]
+            elif int(uid) > 0: command += [uid]
+            elif int(uid) > 0: command = ['chgrp'] + [gid]
+            else: raise CommandError
+            command += [remote_path]
+            self.execute_command(command)
+            return True
+        except CommandError:
+            lgr.error(cmd.msg)
```
yarikoptic: The same here

yarikoptic (changes requested):
Just minor classifications

niceman/resource/session.py
mjtravers: Yes, the gets and puts in the Starcluster package are magically recursive.

niceman/resource/session.py
mjtravers: I did that thinking that the command would return a success or fail status. We can throw an error instead. That is more in keeping with the control/flow of niceman.

yarikoptic: Well, we do throw CommandError if e.g. execution fails, so I think we should be consistent with that -- should make coding using this API easier (instead of constant checks for every command).  May be derive a new exception, e.g. `FileManipulationError`  and `TransferError` both derived from `CommandError` to be used in chmod/chown and put/get accordingly.  This way we could run sequence of commands catching the error once if needed

yarikoptic (changes requested):
I could be wrong but for consistency and coding efficiency we better use exceptions, or I'm not right?

mjtravers:
I committed the changes requested.

I updated the code to favor raising exceptions over error handling.

This PR is ready for another review.

mjtravers:
Yarik, this is ready for another code review, too

niceman/resource/shell.py
```diff
@@ -86,6 +86,13 @@ def mkdir(self, path, parents=False):
             else:
                 os.mkdir(path)
 
+    def get(self, src_path, dest_path, owner=None, group=None):
+        pass
```
yarikoptic: I think, ideally, in case if owner and/or group specified and do not match (thus do a check) we should issue a warning that nothing was changed but they differ (here and in put)

niceman/resource/ssh.py
```diff
@@ -26,7 +28,7 @@ class SSH(Resource):
     name = attr.ib()
 
     # Configurable options for each "instance"
-    host = attrib(doc="DNS or IP address of server")
+    host = attrib(default='127.0.0.1', doc="DNS or IP address of server")
```
yarikoptic: why there should be a default address??

yarikoptic: DNS is generally associated with "server". Here better just to use "Name or IP address"

niceman/resource/ssh.py
```diff
@@ -133,34 +135,36 @@ def _execute_command(self, command, env=None, cwd=None):
...
-            lgr.debug("exec#%i: %s", i, line.rstrip())
+        out = []
+        try:
+            for i, line in enumerate(self.ssh.execute(escaped_command)):
+                out.append(line.rstrip())
+                lgr.debug("exec#%i: %s", i, line.rstrip())
+        except RemoteCommandFailed as e:
+            raise CommandError(cmd=escaped_command, msg=e.msg)
+
+        return (out, None)
```
yarikoptic: same as with docker:  what is happening to stderr?
generally it is passed back:
```shell
$> ssh smaug.datalad.org "bash -c 'echo 123 >&2'" > /dev/null
123
```

niceman/resource/ssh.py
```diff
@@ -133,34 +135,36 @@ def _execute_command(self, command, env=None, cwd=None):
...
-        """Set the mode of a remote path
-        """
-        self.ssh.chmod(mode, remote_path)
-
-    def chown(self, uid, gid, remote_path):
-        """Set the user and group of a path
-        """
-        self.ssh.chown(uid, gid, remote_path)
+        if owner or group:
+            self.chown(owner, group, dest_path, recursive=True)
```
yarikoptic: again -- this sets permissions on remote end, and not locally

yarikoptic: when fixed, please add the test to get/put calls which would verify correct operation in this regard

niceman/resource/tests/test_docker_container.py
```diff
@@ -140,7 +148,27 @@ def test_dockercontainer_class():
...
+        'container_name': 'testing-container'
+    }
+    resource = ResourceManager.factory(config)
+    resource._container == None
+    session = resource.get_session()
+    assert resource._container['Command'] == '/usr/sbin/sshd -D'
+
+    assert session.exists(script) == False
+    session.put(script, "/", owner=None, group=None)
+    assert session.exists(script) == True
```
yarikoptic: would be useful to verify correct operation of providing owner and group settings here, as noted above there is a bit of inconsistencies ATM

niceman/resource/tests/test_docker_container.py
```diff
@@ -140,7 +148,27 @@ def test_dockercontainer_class():
         ]
         client.assert_has_calls(calls, any_order=True)
 
-
-
-
-
+@with_tempfile(content="abc 123")
```
yarikoptic: you could make the "script" actually a script (with shebang) you could `exec` (change permissions locally to make it executable, should get preserved upon `put` call) and which would output stuff to stdout and stderr, which you could then verify correctly picking up right here.  This way would be able to check an actual use-case

yarikoptic: Moreover -- make a generic test helper, e.g. `check_session_operations`, which you would then reuse across different backends -- eg, here with a docker session, then below in testing local session. That is the point that API should allow to perform consistently across different types of sessions.

niceman/resource/tests/test_shell.py
```diff
@@ -100,4 +102,53 @@ def test_source_file_param(script=None):
...
+
+    path = '/tmp/{}'.format(str(uuid.uuid4()))
+    assert os.path.exists(path) == False
+    session.mkdir(path)
+    assert os.path.exists(path) == True
+
+    path = '/tmp/{}/{}'.format(str(uuid.uuid4()), str(uuid.uuid4()))
+    assert os.path.exists(path) == False
+    session.mkdir(path, parents=True)
+    assert os.path.exists(path) == True
```
yarikoptic: all of the above is worthwhile testing in any session (just without using os.path.exists, but `session.exists`). See above comment in docker session testing about a unified test.

niceman/resource/tests/test_ssh.py
```diff
@@ -119,7 +121,42 @@ def test_ssh_class(setup_ssh):
...
+        session.mkdir(path)
+        assert session.isdir(path) == True
+
+        path = '/tmp/{}/{}'.format(str(uuid.uuid4()), str(uuid.uuid4()))
+        assert session.isdir(path) == False
+        session.mkdir(path, parents=True)
+        assert session.isdir(path) == True
+
         assert session.isdir('not-a-dir') == False
         assert session.isdir('/etc/hosts') == False
```
yarikoptic: the same -- generic session testing?

niceman/support/starcluster/sshutils.py
```diff
@@ -236,7 +236,7 @@ def load_remote_rsa_key(self, remote_filename):
         rfile.close()
         return key
 
-    def makedirs(self, path, mode="0755"):
+    def makedirs(self, path, mode=0o755):
```
yarikoptic: is it "identically" treated? or there is a bug in starcluster which needs to be fixed "uptream"?

yarikoptic (changes requested):
Uff... sorry for apparently not reviewing in time. Good stuff! But needs fix ups.  Overall, I think following should be addressed:
- uid/gid vs owner/group unification throughout the API
- more descriptive docstrings reflecting above etc
- fixing chmod'ing in get calls
- unified testing of all possible sessions (could be done via a helper method or just making that test parametrized with all those possible sessions which currently supported)
- consistent names for the `path` argument (no `remote_path` since might even be local in the local session)
- stderr destiny in ssh/docker sessions is unclear -- (theoretically) we must be able to obtain stderr as well!
-  few minor other things I commented on ;)


I am ok to get it merged, and above to go after in a separate PR, after following minor quick fixes:
- not starting manually docker image -- all taken care by fixture now ;-)
- anything else from above you want to address quickly first

niceman/resource/docker_container.py
```diff
@@ -177,31 +185,47 @@ def _execute_command(self, command, env=None, cwd=None):
...
-            if line.startswith(b'rpc error'):
-                raise CommandError(cmd=command, msg="Docker error - %s" % line)
+            out.append(line.rstrip())
             lgr.debug("exec#%i: %s", i, line.rstrip())
 
+        exit_code = self.client.exec_inspect(execute['Id'])['ExitCode']
+        if exit_code not in [0, None]:
+            raise CommandError(cmd=' '.join(command), msg='Exit code: {}'.format(exit_code))
+
+        return (out, None)
```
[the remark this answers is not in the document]
mjtravers: The problem is STDOUT and STDERR are returned co-mingled together by the output generator returned by client.exec_start()

https://docker-py.readthedocs.io/en/1.2.3/api/#exec_create
https://github.com/docker/docker-py/issues/704

The newer Docker SDK API also has this issue:

http://docker-py.readthedocs.io/en/stable/containers.html#container-objects

I have not been able to find a workaround out there somewhere in The Wild.

niceman/resource/docker_container.py
```diff
@@ -177,31 +185,47 @@ def _execute_command(self, command, env=None, cwd=None):
...
 
-    def put(self, src_path, dest_path, preserve_perms=False,
-                owner=None, group=None, recursive=False):
+    def put(self, src_path, dest_path, owner=None, group=None):
         """Take file on the local file system and copy over into the session
         """
-        # self.ssh.put([src_path], remotepath=dest_path)
-        pass
+        tar_stream = io.BytesIO()
+        tar_file = tarfile.TarFile(fileobj=tar_stream, mode='w')
```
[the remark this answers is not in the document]
mjtravers: It is recommended to go through the tar to upload one or more files. There is a "Docker cp" equivalent but it has been depreciated since version 1.2

http://docker-py.readthedocs.io/en/1.5.0/api/#copy

I have added an explanatory comment in the code.

niceman/resource/docker_container.py
```diff
@@ -177,31 +185,47 @@ def _execute_command(self, command, env=None, cwd=None):
...
         """Retrieve a file from the remote system
         """
-        # self.ssh.get(src_path, localpath=dest_path)
-        pass
+        stream, stat = self.client.get_archive(self.container, src_path)
+        tarball = tarfile.open(fileobj=io.BytesIO(stream.read()))
+        tarball.extractall(path=dest_path)
+
+        if owner or group:
+            self.chown(owner, group, dest_path, recursive=True)
```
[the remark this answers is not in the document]
mjtravers: Doh! right

niceman/resource/docker_container.py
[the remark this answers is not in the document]
yarikoptic: hm, sad. We need to poke more at this issue I guess later.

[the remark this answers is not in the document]
yarikoptic: I even start to wonder if we could/should implement something like 
1. set stderr=False 
2. add `2>/tmp/stderr.RANDOM` to the command call  (do we have a "shell" within that remote call?)
3. use `.copy_from` to get that file locally and provide as the stderr

niceman/resource/session.py
```diff
@@ -375,19 +384,46 @@ def read(self, path):
...
+
+    def chmod(self, mode, remote_path, recursive=False):
+        """Set the mode of a remote path
+        """
+        command = ['chmod']
+        if recursive: command += ["-R"]
+        command += [mode] + [remote_path]
+        self.execute_command(command)
+
+    def chown(self, uid, gid, remote_path, recursive=False):
```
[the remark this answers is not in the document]
mjtravers: What is the best way to install @borrowdoc

I am poking around http://www.pymvpa.org/index.html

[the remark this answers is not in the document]
yarikoptic: `from niceman.dochelpers import borrowdoc`  or by "install" did you mean "use"? ;)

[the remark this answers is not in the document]
yarikoptic: as for use, its docstring is even helpful ;)
```ipython
In [1]: from niceman.dochelpers import borrowdoc

In [2]: borrowdoc?
Signature: borrowdoc(cls, methodname=None)
Docstring:
Return a decorator to borrow docstring from another `cls`.`methodname`

It should not be used for __init__ methods of classes derived from
ClassWithCollections since __doc__'s of those are handled by the
AttributeCollector anyways.

Common use is to borrow a docstring from the class's method for an
adapter function (e.g. sphere_searchlight borrows from Searchlight)

Examples
--------
To borrow `__repr__` docstring from parent class `Mapper`, do::

   @borrowdoc(Mapper)
   def __repr__(self):
       ...
```
so in many cases you just need to decorate the method stating which class to borrow from.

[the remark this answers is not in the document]
mjtravers: I meant "install", not "use". Looks like we already have it from the datalad clone. :)
Am working on beefing up documentation and am loathe to copy/paste.

mjtravers:
Yarik,

This PR is ready for merging. 

I will pickup implementing the comments you have for this PR in subsequent PRs.

.travis.yml
```diff
@@ -55,6 +55,8 @@ before_install:
   - travis_retry sudo apt-get update -qq
   - git config --global user.email "test@travis.land"
   - git config --global user.name "Travis Almighty"
+  # Setup SSH docker service
+  - docker run --name testing-container -d -p 49000:22 rastasheep/ubuntu-sshd:14.04
```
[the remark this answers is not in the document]
yarikoptic: so why do we still need this one? please remove or leave a comment here

niceman/resource/base.py
```diff
@@ -291,8 +292,7 @@ def __repr__(self):
         return 'Resource({})'.format(self.name)
 
     def add_command(self, command, env=None):
-        """
-        Add a command to the command buffer so that all commands can be
+        """Add a command to the command buffer so that all commands can be
```
yarikoptic: oh man, I will have so much fun merging ;-)
I know that I somewhat sent you down the rabbit hole with my comments on generic API which required changes only loosely related  to the PR, but these are just stylistic changes better to be avoided when multiple folks known to work on the same pieces (just a note for future)

niceman/resource/docker_container.py
```diff
@@ -26,9 +30,12 @@ class DockerContainer(Resource):
     Environment manager which manages a Docker container.
     """
 
-    # Container properties
+    # Generic properties of any Resource
     name = attr.ib()
+
+    # Container properties
     id = attr.ib(default=None)
+    container_name = attr.ib(default=None)
```
yarikoptic: why do we need both `.name` and a `.container_name`?

niceman/resource/docker_container.py
```diff
@@ -194,22 +207,40 @@ def _execute_command(self, command, env=None, cwd=None):
             lgr.debug("exec#%i: %s", i, line.rstrip())
         return (out, self.client.exec_inspect(execute['Id'])['ExitCode'])
 
+        exit_code = self.client.exec_inspect(execute['Id'])['ExitCode']
+        if exit_code not in [0, None]:
```
yarikoptic: so it is just `if exit_code`?

niceman/resource/docker_container.py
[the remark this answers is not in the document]
yarikoptic: ok - let it get its own issue: https://github.com/ReproNim/niceman/issues/156

niceman/resource/session.py
```diff
@@ -194,49 +256,137 @@ def niceman_exec(self, command, args):
...
+    def put(self, src_path, dest_path, uid=-1, gid=-1):
+        """Take file on the local file system and copy over into the resource
+        
+        Parameters
+        ----------
+        src_path : string
+            Path to file to push to resource environment
+        dest_path : string
+            Path of resource directory to put local file in
+        uid : number, optional
```
yarikoptic: where `number, ` came from?  https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt  does not mention it. Should be an `int` (can't be float, right?)

niceman/resource/session.py
```diff
@@ -375,19 +521,57 @@ def read(self, path):
...
+        self.execute_command(command)
+
+    def chown(self, path, uid=-1, gid=-1, recursive=False):
+        """Set the user and gid of a path
+        """
+        command = ['chown']
+        if recursive: command += ["-R"]
+        if int(uid) > 0 and int(gid) > 0: command += ["{}.{}".format(uid, gid)]
+        elif int(uid) > -1: command += [uid]
+        elif int(uid) > -1: command = ['chgrp'] + [gid]
```
yarikoptic: typo: `uid` --> `gid` in `if`
suggests that tests were not thorough ;-)

niceman/resource/session.py
```diff
@@ -194,49 +256,137 @@ def niceman_exec(self, command, args):
...
+        bool
+            True if path exists
+        """
+        raise NotImplementedError
 
     @abc.abstractmethod
-    def put(self, src_path, dest_path, preserve_perms=False,
-                owner=None, group=None, recursive=False):
-        """Take file on the local file system and copy over into the session
+    def put(self, src_path, dest_path, uid=-1, gid=-1):
```
yarikoptic: eh heh -- so we will stick to C / `os.chown` agreement on `-1` to state "do nothing"... ok for now

niceman/resource/session.py
```diff
@@ -375,19 +521,57 @@ def read(self, path):
...
+        if recursive: command += ["-R"]
+        command += [mode] + [path]
+        self.execute_command(command)
+
+    def chown(self, path, uid=-1, gid=-1, recursive=False):
+        """Set the user and gid of a path
+        """
+        command = ['chown']
+        if recursive: command += ["-R"]
+        if int(uid) > 0 and int(gid) > 0: command += ["{}.{}".format(uid, gid)]
```
yarikoptic: root has id 0 and we should allow to change to it as well. So, `if uid > -1` as well I guess

niceman/resource/session.py
```diff
@@ -375,19 +521,57 @@ def read(self, path):
...
+        elif int(uid) > -1: command += [uid]
+        elif int(uid) > -1: command = ['chgrp'] + [gid]
+        else: raise CommandError(cmd='chown', msg="Invalid command \
+            parameters.")
+        command += [path]
+        self.execute_command(command)
+
+    def local_chown(self, src_path, dest_path, uid=-1, gid=-1):
+        target = dest_path + os.sep + os.path.basename(src_path)
+        os.chown(target, int(uid), int(gid))
```
yarikoptic: what for all those `int` calls here and everywhere?

please add a docstring to `local_chown` that it is a helper (so may be should go to `_protected` method?) to set permissions for the local copy of a file (directory?) ... seems like dest_path is assumed to be a directory and them src_path a file? or why its basename is taken?

niceman/resource/session.py
```diff
@@ -375,19 +521,57 @@ def read(self, path):
...
+        if os.path.isdir(target):
+            for root, dirs, files in os.walk(target):
+                for dir in dirs:
+                    os.chown(os.path.join(root, dir), int(uid), int(gid))
+                for file in files:
+                    os.chown(os.path.join(root, file), int(uid), int(gid))
```
yarikoptic: `file` and `dir` are internal keywords of python.. ideally do not overload. disambiguate with `_`, e.g. `file_` and `dir_`.  I know that it is minor -- but believe me, I did have to fix bugs where I had to stare at the screen for a while to figure out that `file` is not actually a Python `file`

yarikoptic (changes requested):
please answer some comments from the past reviews as well.  ran into some typos in the new changes
Cheers!

niceman/resource/base.py
mjtravers: For the merging, you are referring to PR #98? We should probably work on that together in TX

niceman/resource/docker_container.py
mjtravers: I thought decoupling the resource name from the Docker container name may be useful. Probably just complicates more than its utility. I will remove

niceman/resource/docker_container.py
mjtravers: Just raising the exception if the command exit code comes back other than zero. Are there other scenarios I am not thinking of?

niceman/resource/base.py
```diff
@@ -367,13 +368,33 @@ def get_updated_env(self, custom_env):
 
     @classmethod
     def _generate_id(cls):
+        """Utility class method to generate a UUID.
+        
+        Returns
+        -------
+        string
+            Newly crated UUID
```
kyleam: s/crated/created/

niceman/tests/fixtures.py
```diff
@@ -22,6 +24,39 @@ def niceman_cfg_path(request):
...
+                file = open(self.path, "w")
+                file.write(content)
+                file.close()
+        def __repr__(self):
+            return self.path
+
+    fid, path = tempfile.mkstemp()
+    temp_file = FixtureTempFile(fid, path)
+    yield temp_file
+    os.remove(path)
```
kyleam: Thoughts/my own confusions:

  * I'm probably missing something, but I don't see the advantage of
    this structure over returning the result of
    `tempfile.NamedTemporaryFile`.

  * Why open and close each call?

  * Why implement `__call__` rather than sticking to the more familiar
    `.write` for a file objects?

  * Is the `__repr__` method made use of in the current tests?

  * I would have expected this to have pytest's default function scope
    rather than module scope.

kyleam (commented):
In addition to these minor comments, there is a "subdirectores" typo
in niceman/resource/session.py, but I can't mark it because I can't
convince GitHub to load the diff.

niceman/tests/fixtures.py
mjtravers: I was focused on mkstemp() and missed NamedTemporaryFile, which is basically doing what I coded up. The __call__ was to simplify file operations but passing back the established NamedTemporaryFile makes more sense. Python noob mistake, excellent guidance. Will rework and push

mjtravers: UPDATE: The simplicity of setting up a temporary file with NamedTemporaryFile makes the use of a fixture overkill. I removed the temp_file fixture in favor of using tempfile.NamedTemporaryFile directly.

niceman/interface/tests/test_exec.py
```diff
@@ -0,0 +1,54 @@
...
+            'exec',
+            'mkdir',
+            path,
+            '--name', 'testing-container',
+            '--config', niceman_cfg_path
+        ])
+
+        session = ResourceManager.factory(config).get_session()
+
+        assert session.exists(path) == True
```
yarikoptic: no need for `== True`

yarikoptic (changes requested):
Let's get those docker tests running on travis and merge the beast

niceman/tests/fixtures.py
yarikoptic: I am lost why we need another fixture in addition to existing `tmpdir` fixture in pytest (also remember we still have nose-style `@with_tempfile` with cute features suchas `content` option)

kyleam: > I am lost why we need another fixture

I believe it's already be removed in e84402f6.

niceman/resource/tests/test_shell.py
```diff
@@ -48,56 +50,95 @@ def test_shell_class():
...
+        'type': 'shell'
+    }
+    resource = ResourceManager.factory(config)
+
+    status = resource.create()
+    print("====================", status)
+    assert re.match('\w{8}-\w{4}-\w{4}-\w{4}-\w{12}$', status['id']) is not None
+
+    assert type(resource.connect()) == Shell
+    assert resource.delete() == None
```
yarikoptic: generally `is`/`is not` is advisable over `==`/`!=` when comparing to None (or any other singleton)

niceman/resource/tests/test_session.py
```diff
@@ -46,4 +78,273 @@ def test_get_updated_env():
...
     assert get_updated_env({'a': 1}, {'a': None}) == {}
     assert get_updated_env({'a': 1, 'b': 2}, {'a': None}) == {'b': 2}
-    assert get_updated_env({'a': 1, 'b': 2}, {'a': None, 'b': 3}) == {'b': 3}
\ No newline at end of file
+    assert get_updated_env({'a': 1, 'b': 2}, {'a': None, 'b': 3}) == {'b': 3}
+
+
+def test_get_local_session():
+    # get_local_session(env={'LC_ALL': 'C'}, pty=False, shared=False)
+    return
```
yarikoptic: this test isn't finished seems to me ;-)

niceman/resource/tests/test_shell.py
yarikoptic: I will replace all of those in one sweep

niceman/tests/utils.py
```diff
@@ -403,6 +403,32 @@ def newfunc(*args, **kwargs):
...
+    func
+        Decorator function
+    
+    Raises
+    ------
+    SkipTest
+    """
+    def decorator(func):
+        stdout, _ = Runner().run(['docker', 'ps'])
+        if container_name not in stdout:
```
yarikoptic: very weak check since could be a partial match. Please TODO @mjtravers  - make matching by the value of the second column

niceman/resource/ssh.py
```diff
@@ -46,15 +56,16 @@ class SSH(Resource):
     _ssh = attr.ib(default=None)
 
     def connect(self):
+        """Open a connection to the environment resource.
         """
-        Open a connection to the environment resource.
-        """
-        self._ssh = SSHClient(
+        self._ssh = paramiko.SSHClient()
+        self._ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
yarikoptic: TODO (file an issue): we shouldn't blindly login to unknown hosts

yarikoptic: didn't look how difficult or possible via paramiko but we need make sure that it works coherently with existing system/user ssh configuration, i.e. that existing known_hosts used by user's `ssh` command are used, which seems to be not the case in simplest paramiko invocation:

```
$> python -c 'from paramiko import SSHClient; ssh = SSHClient().connect("127.0.0.1", username=None)'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/paramiko/client.py", line 361, in connect
    server_key)
  File "/usr/lib/python2.7/dist-packages/paramiko/client.py", line 648, in missing_host_key
    raise SSHException('Server %r not found in known_hosts' % hostname)
paramiko.ssh_exception.SSHException: Server '127.0.0.1' not found in known_hosts
```
- make sure that ssh keys authentication (possibly provided by ssh-agent) would be used

niceman/resource/ssh.py
```diff
@@ -110,93 +121,159 @@ class Ssh(SSH):
...
+        except paramiko.AuthenticationException:
+            raise exception.SSHAuthException(self.ssh._username,
+                self.ssh._host)
+        except paramiko.SSHException as e:
+            msg = e.args[0]
+            raise exception.SSHError(msg)
+        except socket.error:
+            raise exception.SSHConnectionError(self.ssh._host, self.ssh._port)
+        except EOFError:
+            raise exception.SSHConnectionError(self.ssh._host, self.ssh._port)
```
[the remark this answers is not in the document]
yarikoptic: note also that pycharm was warning you highlighting that `exception.` isn't defined here -- names were imported into the namespace... will fix up for this.

yarikoptic (commented):
Left some comments around, will address some minor ones in upcoming PR.  There are some comments which should be generalized into separate issues upon reading.  
Meanwhile I will merge this since it grew into a monster and we cannot hold it longer growing in the shadows ;)

PR #130, opened 2017-10-31
BF: docker fixture

### LINKED ISSUE #124: SSH and Conda tests failing
Using Debian 9.

Commit 4ae06f9 is okay.

Commit 2e53e36 introduces the error in [error1.txt](https://github.com/ReproNim/niceman/files/1395800/error1.txt).

Commit d4bdac7 introduces the error in [error2.txt](https://github.com/ReproNim/niceman/files/1395802/error2.txt).

chaselgrove:
Creates a fixture to start and stop docker containers for test_ssh_class().

Fixes #124 .

@mjtravers, I've written this so an existing docker container is used if it is running.

@rbuccigrossi, what magic did you use to get the get_conda_test_dir fixture to work with skip_if_no_network?  When I added the @skip_if_no_network decorator to my fixture like you did to get_conda_test_dir, it wouldn't run.


yarikoptic:
Just call skip_if_no_network () within your fixture, then it would raise skip test. Unfortunately out custom decorators don't play nicely yet with pytest fixtures

niceman/resource/tests/test_ssh.py
```diff
@@ -15,12 +15,51 @@
...
+    stdout = po.communicate()[0]
+    po.wait()
+    # stdout is a bytes object in Python 3, so we need to make the 
+    # search string a bytes object as well
+    if b'0.0.0.0:49000->22/tcp' in stdout:
+        stop_container = False
+    else:
+        args = ['docker', 
+                'run', 
+                '-d',
```
yarikoptic: Should it have --rm to not breed containers?

niceman/resource/tests/test_ssh.py
```diff
@@ -15,12 +15,51 @@
...
+        args = ['docker', 
+                'run', 
+                '-d', 
+                '-p', 
+                '49000:22', 
+                'rastasheep/ubuntu-sshd:14.04']
+        po = subprocess.Popen(args, stdout=subprocess.PIPE)
+        stdout = po.communicate()[0]
+        returncode = po.wait()
+        assert returncode == 0, 'error starting docker container'
```
yarikoptic: You could use niceman.cmd.Runner to simplify all of the above to one line

niceman/resource/tests/test_ssh.py
```diff
@@ -15,12 +15,51 @@
...
+    """pytest fixture for tests needing a running docker container
+
+    on setup, this fixture ensures that a docker container that maps 
+    host port 49000 to container port 22 is running and starts one if necessary
+
+    on teardown, this fixture stops the docker container if it was started by 
+    the fixture
+    """
+    if os.environ.get('NICEMAN_TESTS_NONETWORK'):
+        raise SkipTest("Skipping since no network settings")
```
yarikoptic: just call `skip_if_no_network()`

niceman/resource/tests/test_ssh.py
```diff
@@ -15,12 +15,51 @@
...
+from ...tests.utils import assert_in
 from ..base import ResourceManager
 
+import pytest
+from nose import SkipTest
+import subprocess
 
-@skip_if_no_network
-def test_ssh_class():
+@pytest.fixture
```
yarikoptic: add `scope="module"` to it

chaselgrove:
You want this somewhere else, too, say niceman/tests/utils.py?

niceman/resource/tests/test_ssh.py
```diff
@@ -17,10 +17,42 @@
...
+    stdout, _ = Runner().run(['docker', 'ps'])
+    if '0.0.0.0:49000->22/tcp' in stdout:
+        stop_container = False
+    else:
+        args = ['docker', 
+                'run', 
+                '-d', 
+                '--rm', 
+                '-p', 
+                '49000:22',
```
yarikoptic: wouldn't it cause trouble for those who run tests locally and have sshd running already on port 22?  why not to just retain/reuse port 49000 or some other privileged, but unlikely to be used port (2223?)?

chaselgrove: I don't understand -- 22 is the container port and 49000 is the host port, so it doesn't matter if the host is running sshd.

yarikoptic: ah, I took it wrong direction probably

chaselgrove:
Can now be merged.

niceman/resource/tests/test_ssh.py
```diff
@@ -17,11 +17,44 @@
...
+        container_id = stdout.strip()
+        stop_container = True
+    yield
+    if stop_container:
+        Runner().run(['docker', 'stop', container_id])
+    return
+
+def test_ssh_class(setup_docker):
+
+    skip_ssh()
```
yarikoptic: 1. just move `skip_ssh` inside the `setup_docker`.  This way we wouldn't need to call it in every function which uses setup_docker
2. please rename setup_docker to `setup_ssh` (since it is not per se about (any) docker but rather about establishing ability to test ssh).  

And then it will be done (although ideally, the fixture should provide those parameters as host, user, password, port ;) but if you aren't up for this challenge -- we could do that later)

niceman/tests/utils.py
```diff
@@ -382,18 +382,25 @@ def newfunc(*args, **kwargs):
...
-    return newfunc
+
+    if func:
+        @wraps(func)
+        def newfunc(*args, **kwargs):
+            if on_windows:
+                raise SkipTest("SSH currently not available on windows.")
+            check_and_raise()
+            return func(*args, **kwargs)
+        return newfunc
```
yarikoptic: now that there is no decorator, this code isn't tested ... suboptimal. ideally there should be a test (with `setup_docker`/`setup_ssh` fixture) which has inside another function which is `@skip_ssh` decorated, and calls it and verifies that it returns desired result.  This way we would assure at least that decorator works somewhat

niceman/resource/tests/test_ssh.py
chaselgrove: These two options don't play well together.  It made more sense to me to make this a general docker fixture (see above), but it's dragging on a bit and at this point I'd rather just close this and move on to something more productive.  Let's discuss on Monday with Matt as well (who was working on the ssh bit).

yarikoptic: More productive is just to get it done... I could do it if you like - it is just few lines changes. but could you explain what you meant by "don't play well together"?

yarikoptic: Refactoring into a generic docker fixtures probably should be done whenever we have the use case for that. ATM it is just the SSH over

yarikoptic:
ok, I will merge it as is and in an hour, unless you stop me since it is WiP @chaselgrove, I will submit a followup PR addressing my comments

PR #131, opened 2017-11-02
Addressing code review comments for debian identification (#115)

rbuccigrossi:


niceman/support/external_versions.py
```diff
@@ -69,7 +72,8 @@ class ExternalVersions(object):
 
     CUSTOM = {
         'cmd:annex': _get_annex_version,
-        'cmd:git': _get_git_version
+        'cmd:git': _get_git_version,
+        'cmd:apt-cache': _get_apt_cache_version
```
yarikoptic: insignificant "styling" comment:  python allow trailing `,` between items of the list, dict, or tuple.  So I usually add it to minimize the diff next time (since I wouldn't need to add it to preceding line)

niceman/support/distributions/tests/test_debian.py
```diff
@@ -285,6 +285,10 @@ def test_parse_apt_cache_policy_source_info():
  500 http://security.ubuntu.com/ubuntu xenial-security/restricted amd64 Packages
      release v=16.04,o=Ubuntu,a=xenial-security,n=xenial,l=Ubuntu,c=restricted,b=amd64
      origin security.ubuntu.com
+ 500 http://debproxy:9999/debian/ jessie-backports/contrib Translation-en
+ 100 http://debproxy:9999/debian/ jessie-backports/non-free amd64 Packages
+     release o=Debian Backports,a=jessie-backports,n=jessie-backports,l=Debian Backports,c=non-free
+     origin debproxy
```
yarikoptic: I believe I added it on the commit but it doesn't show here -- could we have a test which tests that value is parsed correctly?

rbuccigrossi: That is a quote right from your bug report, but I cut out the exact portion that caused the problem.  I'm happy to add more if you wish...

yarikoptic: What I meant -- could the test verify correct parsing of this line -- that we are getting correct record with `origin="Debian Backports"` etc?

PS as for what "doesn't show here" -- it was about my comment, not the code ;)

yarikoptic:
I think we could merge as soon as wip label disappears ;-)

niceman/support/distributions/tests/test_debian.py
rbuccigrossi: Done ( 496efe8 )

rbuccigrossi:
@yarikoptic I removed the WIP tag. There's only one thing left in your original review, and that's the "sledgehammer test".  I'll do that as a separate pull request.  So if you're happy, merge away!

PR #134, opened 2017-11-03
Create LICENSE

satra:


LICENSE
```diff
@@ -0,0 +1,13 @@
+ Copyright 2017 ReproNim
```
yarikoptic: is there some kind of a ReproNim entity which could hold the copyright?

I would go for "ReproNim Developers" but even then not sure e.g. who is holding the copyright for work done by TCG folks -- might be TCG.

yarikoptic:
closing for now, as described, we have COPYING file and the majority of files in the project already have consistent header pointing to it ;)

PR #135, opened 2017-11-24
SSH fixture - skips within before initiating docker, renamed

yarikoptic:
continuation to #130 
immediate pros: no docker started unless actually needed

TODOs:
- [x] test for the decorator
- [x]  pass parameters for the connection via the fixture to avoid any hardcoded specs in the test code

### DIFF
--- a/niceman/resource/tests/test_ssh.py
+++ b/niceman/resource/tests/test_ssh.py
@@ -39,20 +39,20 @@ def setup_ssh():
     skip_if_no_network()
     skip_ssh()
     stdout, _ = Runner().run(['docker', 'ps'])
-    if '0.0.0.0:49000->22/tcp' in stdout:
+    params = {
+        'host': 'localhost',
+        'user': 'root',
+        'password': 'root',
+        'port': '49000',
+    }
+    if '0.0.0.0:{port}->22/tcp'.format(**params) in stdout:
         stop_container = False
     else:
-        params = {
-            'host': 'localhost',
-            'user': 'root',
-            'password': 'root',
-            'port': '49000',
-        }
-        args = ['docker', 
-                'run', 
-                '-d', 
-                '--rm', 
-                '-p', 
+        args = ['docker',
+                'run',
+                '-d',
+                '--rm',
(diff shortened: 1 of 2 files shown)


niceman/tests/test_tests_utils.py
```diff
@@ -518,3 +518,24 @@ def test_assert_is_subset_recur():
...
+
+        with patch.dict('os.environ', {'NICEMAN_TESTS_SSH': "1"}):
+            assert func(2) == 4
+    except SkipTest:
+        raise AssertionError("must have not skipped")
+
+    with patch.dict('os.environ', {'NICEMAN_TESTS_SSH': ""}):
+        try:
+            func(2)
+            assert "Must have thrown SkipTest"
```
chaselgrove: "Must have thrown SkipTest" evaluates to True, so you probably want raise AssertionError() again here.

So...who tests the tests' tests? :)

niceman/resource/tests/test_ssh.py
```diff
@@ -23,51 +23,57 @@
...
                 '--rm', 
                 '-p', 
-                '49000:22', 
+                '{port}:22'.format(**params),
                 'rastasheep/ubuntu-sshd:14.04']
         stdout, _ = Runner().run(args)
         container_id = stdout.strip()
         stop_container = True
-    yield
+    yield params
```
chaselgrove: params is not defined if a suitable container is already running (if the test on line 42 succeeds).  You might also want to use params['port'] in this test so 49000 only appears once.

niceman/tests/test_tests_utils.py
yarikoptic: ha -- thanks... it must have been late or smth ;)

the week link in the chain -- humans! ;)

niceman/resource/tests/test_ssh.py
yarikoptic: ho ho... thanks! fixing up!

yarikoptic:
@chaselgrove pushed fixes on both, thanks!

PR #137, opened 2017-11-29
ENH: Large test for debian package identification

rbuccigrossi:
This adds a new slow test that identifies packages for files in /usr/bin/ and /usr/lib/. To run it, pass the "--runslow" parameter to py.test.  The test itself makes sure that:

- Non-local origins exist
- Non-local origins have mandatory attributes
- That all files are identified (it fails and lists any files that are not identified)

### DIFF
--- a/niceman/distributions/tests/test_debian.py
+++ b/niceman/distributions/tests/test_debian.py
@@ -8,6 +8,8 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
 import os
+from os.path import islink
+from os.path import join, isfile
 
 from pprint import pprint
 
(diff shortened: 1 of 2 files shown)


niceman/distributions/tests/test_debian.py
```diff
@@ -54,6 +56,34 @@ def test_dpkg_manager_identify_packages():
         assert False, "A non-local origin must be found"
 
 
+@pytest.mark.slow
```
yarikoptic: I guess we would never know how long it takes on travis unless we run it there, right? ;)

yarikoptic: might also be a good time to try to profile our code since dpkg-query itself is quite fast and I don't remember if we otherwise analyzed our current implementation to make it any faster (might be doing the same thing over and over again).  The simplest initial step might be to use `@line_profile` helper for e.g. `identify_packages_from_files`

yarikoptic: BTW -- it did finish for me without crashing! test failed since we do not support yet diversions (now #138) and I had some leftover .distrib files for some reason.  if only it was a tiny bit faster ;)

yarikoptic:
Now it is a part of the blazingly fast #143

Issue #142, opened 2017-12-06
performance: probably need to bundle all apt-cache policy and show calls


yarikoptic:
ATM retracing invokes those per each package separately.  But it seems that there is a major startup penalty in those commands (when considering multiple invocations).  I guess the only way is to do two pass analysis -- collect the list of all packages to be considered, pass to those commands in a single invocation for each, and then analyze for per package information.  Should provide a notable speed up ;-)

rbuccigrossi:
From today's discussion:

  The best general approach is to update DebTracer::identify_distributions() to not call the base class identify_package_from_files(), but instead loop through the files and gather all the package names first (This is what CondaTracer does now). That way we can get the details of all the packages in one shot.

  We use a number of system calls (dpkg -s, apt-cache show, and apt-cache policy) that support multiple packages nicely. We also use "stat" on the <pkg>.list file to get the creation date (representing the install date of the package).  To get stat to display the date and the filename, we should switch the current call to something like:

```shell
stat -c "%n: %Y" filename1 filename2
```

This will allow us to parse both the filename and the create date timestamp for each file.

Issue #138, opened 2017-11-29
debian: support diversions


yarikoptic:
Some packages (and admins), in rare but occurring cases, divert (not just rename once, but so that future package updates also reflect that) a file aside.  So e.g.
```shell
$> dpkg-divert --list /usr/bin/x86_64-linux-gnu-strip*      
diversion of /usr/bin/x86_64-linux-gnu-strip to /usr/bin/x86_64-linux-gnu-strip.single by binutils-multiarch
2 10191.....................................:Wed 29 Nov 2017 12:39:58 AM EST:.
(git)hopa:~/proj/repronim/niceman[pr-137]git
$> dpkg-query -S /usr/bin/x86_64-linux-gnu-strip*
diversion by binutils-multiarch from: /usr/bin/x86_64-linux-gnu-strip
diversion by binutils-multiarch to: /usr/bin/x86_64-linux-gnu-strip.single
binutils-multiarch, binutils-x86-64-linux-gnu: /usr/bin/x86_64-linux-gnu-strip
diversion by binutils-multiarch from: /usr/bin/x86_64-linux-gnu-strip
diversion by binutils-multiarch to: /usr/bin/x86_64-linux-gnu-strip.single
2 10192.....................................:Wed 29 Nov 2017 12:41:32 AM EST:.
(git)hopa:~/proj/repronim/niceman[pr-137]git
$> dpkg-query -S /usr/bin/x86_64-linux-gnu-strip 
diversion by binutils-multiarch from: /usr/bin/x86_64-linux-gnu-strip
diversion by binutils-multiarch to: /usr/bin/x86_64-linux-gnu-strip.single
binutils-multiarch, binutils-x86-64-linux-gnu: /usr/bin/x86_64-linux-gnu-strip
```

rbuccigrossi:
Huh. How did you get the one file to appear in two different projects?  I have a case of a diversion that works on my system:

```shell
[butch@ubuntu ~/repronim] $ dpkg-query -S /usr/bin/pod2latex
diversion by libpod-latex-perl from: /usr/bin/pod2latex
diversion by libpod-latex-perl to: /usr/bin/pod2latex.bundled
libpod-latex-perl: /usr/bin/pod2latex
[butch@ubuntu ~/repronim] $ dpkg-query -S /usr/bin/pod2latex  /usr/bin/pod2latex.bundled
diversion by libpod-latex-perl from: /usr/bin/pod2latex
[...]

Issue #102, opened 2017-08-09
debian: support alternatives


yarikoptic:
dpkg-query doesn't detect those 
```shell
$> niceman retrace /usr/bin/jar
2017-08-09 16:23:38,763 [INFO   ] Tracing using <class 'niceman.distributions.debian.DebTracer'> 
2017-08-09 16:23:40,346 [INFO   ] DebTracer: 0 packages with 0 files, and 1 other files 
2017-08-09 16:23:40,346 [INFO   ] Tracing using <class 'niceman.distributions.conda.CondaTracer'> 
2017-08-09 16:23:40,346 [INFO   ] Tracing using <class 'niceman.distributions.vcs.VCSTracer'> 
2017-08-09 16:23:40,475 [INFO   ] VCSTracer: 0 packages with 0 files, and 1 other files 
# NICEMAN Environment Configuration File
# This file was created by NICEMAN 0.0.2 on 2017-08-09 16:23:40.477064
version: 0.0.1
files: [/usr/bin/jar]
(dev) 2 35648 [1].....................................:Wed 09 Aug 2017 04:23:40 PM EDT:.
(git)hopa:~/proj/repronim/niceman[enh-sessions]git
$> ls -l /usr/bin/jar
lrwxrwxrwx 1 root root 21 Mar  6  2015 /usr/bin/jar -> /etc/alternatives/jar*
(dev) 2 35649 [1].....................................:Wed 09 Aug 2017 04:23:47 PM EDT:.
(git)hopa:~/proj/repronim/niceman[enh-sessions]git
$> ls -l /etc/alternatives/jar 
lrwxrwxrwx 1 root root 41 Aug 28  2016 /etc/alternatives/jar -> /usr/lib/jvm/java-8-openjdk-amd64/bin/jar*
(dev) 2 35650 [1].....................................:Wed 09 Aug 2017 04:23:52 PM EDT:.
(git)hopa:~/proj/repronim/niceman[enh-sessions]git
$> niceman retrace /usr/lib/jvm/java-8-openjdk-amd64/bin/jar 
2017-08-09 16:24:02,811 [INFO   ] Tracing using <class 'niceman.distributions.debian.DebTracer'> 
2017-08-09 16:24:04,390 [INFO   ] DebTracer: 1 packages with 1 files, and 0 other files 
2017-08-09 16:24:04,391 [INFO   ] Tracing using <class 'niceman.distributions.conda.CondaTracer'> 
[...]

PR #143, opened 2017-12-14
WIP: Enhancement -  batch apt package commands

rbuccigrossi:
This branch bundles apt-cache, dpkg, and stat calls across multiple packages to speed up Debian identification. #142 

It reduces execution time of test_check_bin_packages() from 32 seconds to 2.1 seconds on my machine :)

One trick was to be able to carefully match up command results and the original list of packages. I end up placing the command results into a dictionary based on the package name (and package:arch for apt-cache and dpkg calls), and then loop through packages looking for matching results.

It's now ready for code review...

### DIFF
--- a/niceman/distributions/debian.py
+++ b/niceman/distributions/debian.py
@@ -244,6 +244,9 @@ class DebTracer(DistributionTracer):
         # found association
         if not packages:
             return
+
+        packages = self.get_details_for_packages(packages)
+
         # TODO: Depending on ID might be debian or ubuntu -- we might want to
         # absorb them all within DebianDistribution or have custom classes??
         # So far they seems to be largely similar
@@ -302,45 +305,63 @@ class DebTracer(DistributionTracer):
         # Create a named origin
         return name
 
-    def _create_package(self, name, architecture=None):
+    def get_details_for_packages(self, packages):
         # Find apt sources if not defined
         if not self._all_apt_sources:
             self._find_all_sources()
 
-        # Use dpkg -s <pkg> to get arch and version
-        architecture, version = self._get_pkg_arch_and_version(name,
-                                                               architecture)
-        if not version:
-            lgr.warning("Unable to query package %s" % name)
-            return None
+        # Store the package details as dicts so that we can easily add to them
+        pkg_details = [attr.asdict(pkg) for pkg in packages]
 
-        # Use apt-cache show <pkg> to get details
-        info = self._get_pkg_details(name, architecture, version)
-        if not info:
-            lgr.warning("Unable to get details for package %s" % name)
-            return None
+        new_packages = []
+        for pkg in pkg_details:
+            name = pkg.get("name")
+            architecture = pkg.get("architecture")
+            # Use dpkg -s <pkg> to get arch and version
+            architecture, version = self._get_pkg_arch_and_version(name,
+                                                                   architecture)
+            if not version:
+                lgr.warning("Unable to query package %s" % name)
+                return None
+            pkg["architecture"] = architecture
+            pkg["version"] = version
 
-        # Get install date from the modify time of the dpkg info file
-        install_date = self._get_pkg_install_date(name)
+            # Use apt-cache show <pkg> to get details
+            info = self._get_pkg_details(name, architecture, version)
+            if not info:
+                lgr.warning("Unable to get details for package %s" % name)
+                return None
+            pkg["source_name"] = info.get("Source_name")
+            pkg["source_version"] = info.get("Source_version")
+            pkg["size"] = info.get("Size")
+            pkg["md5"] = info.get("MD5sum")
+            pkg["sha1"] = info.get("SHA1")
+            pkg["sha256"] = info.get("SHA256")
 
-        # Now use "apt-cache policy pkg:arch" to get versions
-        ver_dict = self._get_pkg_versions_and_sources(name, architecture)
-        if not ver_dict:
-            lgr.warning("Was unable to get version table for %s" % name)
-            return None
+            # Get install date from the modify time of the dpkg info file
+            install_date = self._get_pkg_install_date(name)
+            pkg["install_date"] = install_date
 
+            # Now use "apt-cache policy pkg:arch" to get versions
+            ver_dict = self._get_pkg_versions_and_sources(name, architecture)
+            if not ver_dict:
+                lgr.warning("Was unable to get version table for %s" % name)
+                return None
+            pkg["versions"] = ver_dict
+
+        new_packages = []
+        for pkg in pkg_details:
+            new_pkg = DEBPackage(**pkg)
+            new_packages.append(new_pkg)
+
+        return new_packages
+
+    def _create_package(self, name, architecture=None):
+
+        # Store the details we currently know, we will populate the rest later
         return DEBPackage(
             name=name,
-            version=version,
             architecture=architecture,
-            source_name=info.get("Source_name"),
-            source_version=info.get("Source_version"),
-            size=info.get("Size"),
-            md5=info.get("MD5sum"),
-            sha1=info.get("SHA1"),
-            sha256=info.get("SHA256"),
-            install_date=install_date,
-            versions=ver_dict
         )
 
     def _find_all_sources(self):
@@ -427,7 +448,6 @@ class DebTracer(DistributionTracer):
             ['apt-cache', 'policy', query]
         )
         out = utils.to_unicode(out, "utf-8")
-        # dpkg -s uses the same output as apt-cache show pkg
         ver = parse_apt_cache_policy_pkgs_output(out)
         if not ver:
             return None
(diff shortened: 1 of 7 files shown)


niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+            p["architecture"] = r.get("Architecture")
+            p["version"] = r.get("Version")
+
+    @staticmethod
+    def create_lookup_from_apt_cache_show(cmd_results):
+        lookup_results = {}
+        for r in cmd_results:
+            lookup_results[r.get("Package")] = r
+            lookup_results["%s:%s" % (r.get("Package"),
+                                      r.get("architecture"))] = r
```
yarikoptic: getting lost in capitalizations  -- some times all upper case, some times all lower, and now mixed in ;)  any chance those all could be unified to e.g. lower case?

also note could be `"%(Package)s:%(Architecture)s" % r` or `"{Package}:{Architecture}".format(**r)` to avoid explicit `.get` and not end up with spurious None there (we better fail early in our testing than to discover those later)

niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+                           else "%s:%s=%s" % (name, architecture, version))
+
+        # Find out how many packages we can query at once
+        max_len = max([len(q) for q in queries])
+        num_queries = max((_MAX_LEN_CMDLINE - 13) // (max_len + 1), 1)
+
+        cmd_results = []
+        # Execute dpkg -s, parse, and collate the output
+        for subqueries in (queries[pos:pos + num_queries]
+                           for pos in range(0, len(queries), num_queries)):
```
yarikoptic: aha -- the same pattern -- so we at least need a helper, duplication is evil ;)

niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+        for p in pkg_dicts:
+            # Use "dpkg -s pkg" to get the installed version and arch
+            architecture = p.get("architecture")
+            name = p.get("name")
+            version = p.get("version")
+            queries.append("%s=%s" % (name, version) if not architecture
+                           else "%s:%s=%s" % (name, architecture, version))
+
+        # Find out how many packages we can query at once
+        max_len = max([len(q) for q in queries])
```
yarikoptic: minor (feel free to ignore, just nitpicking): `max(map(len, queries))`  or use `()` comprehension not a `[]` to avoid creating a possibly long list (not that it would make it much faster, but who knows??)

niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+            if not r:
+                lgr.warning("Was unable to run apt-cache show for %s" %
+                            p.get("name"))
+                continue
+            p["source_name"] = r.get("Source_name")
+            p["source_version"] = r.get("Source_version")
+            p["size"] = r.get("Size")
+            p["md5"] = r.get("MD5sum")
+            p["sha1"] = r.get("SHA1")
+            p["sha256"] = r.get("SHA256")
```
yarikoptic: if those should be defined, better imho to use `[]` not `get`, again to catch problems early. I would use `get` more sparingly -- really in the cases where I expect that there might be no that key.

And since I don't like duplication, I would have done it like (again -- just nitpicking, please ignore unless you like it too ;))
```python
for f in ('Source_name', 'Source_version', 'Size', ...):
   # only some fields needs custom renaming
   p[{'MD5sum': 'md5'}.get(f, f)] = r[f]
```
which might be worthwhile of a little helper `map_fields(from, to, mapping={})`

niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+            queries.append(self._pkg_name_to_dpkg_list_file(p.get("name")))
+
+        # Find out how many packages we can query at once
+        max_len = max([len(q) for q in queries])
+        num_queries = max((_MAX_LEN_CMDLINE - 13) // (max_len + 1), 1)
+
+        cmd_results = {}
+        # Execute dpkg -s, parse, and collate the output
+        for subqueries in (queries[pos:pos + num_queries]
+                           for pos in range(0, len(queries), num_queries)):
```
yarikoptic: oops -- I see it again! ;)

niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+                           else "%s:%s" % (name, architecture))
+
+        # Find out how many packages we can query at once
+        max_len = max([len(q) for q in queries])
+        num_queries = max((_MAX_LEN_CMDLINE - 13) // (max_len + 1), 1)
+
+        cmd_results = {}
+        # Execute dpkg -s, parse, and collate the output
+        for subqueries in (queries[pos:pos + num_queries]
+                           for pos in range(0, len(queries), num_queries)):
```
yarikoptic: oy, it is haunting me!

niceman/distributions/debian.py
```diff
@@ -368,93 +365,212 @@ def _find_all_sources(self):
...
+
+        # Find out how many packages we can query at once
+        max_len = max([len(q) for q in queries])
+        num_queries = max((_MAX_LEN_CMDLINE - 13) // (max_len + 1), 1)
+
+        cmd_results = []
+        # Execute dpkg -s, parse, and collate the output
+        for subqueries in (queries[pos:pos + num_queries]
+                           for pos in range(0, len(queries), num_queries)):
+            try:
```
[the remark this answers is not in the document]
yarikoptic: actually it shouldn't  be the modification to execute_command, but rather a dedicated helper (since we want to generate here), either `_session.execute_command_in_batches` or a generic `call_in_batches(self._session.execute_command, ['dpkg', '-s', None], args=...)` (do not like `_batches` so a better name is welcome!)

niceman/distributions/debian.py
rbuccigrossi: I updated the code to make our calls to get() explicit when we expect a value to be optional.  This allowed me to simplify some cases when I knew values should be set. I also used the shorter version to format the strings.

niceman/distributions/debian.py
rbuccigrossi: I pushed the lowercase and md5 to the parsing support function, simplifying the code here.

niceman/distributions/debian.py
rbuccigrossi: I also made the parsing support function translate all keys to lower case to simplify the mapping.

niceman/distributions/debian.py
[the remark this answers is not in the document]
rbuccigrossi: The key issue is that sometimes I parse the results into arrays and sometimes into dictionaries. Meanwhile, I need to handle CommandErrors in different ways. 

So instead of creating call_in_batches(), I wrote batch_process_list(proc_func, proc_list, batch_len, start_val), which takes a processing function as an argument and iteratively calls that on the proc_list split into batch_len. The processing function is of the form f(batch, prev_value) -> result_value, and is written to take the results previous call as input (starting with start_val). 

This gave me the flexibility to change how I handle command output and exceptions.

I look forward to your review of the result (good or bad :).

niceman/distributions/debian.py
rbuccigrossi: Done (now in utils.py get_cmd_batch_len() )

niceman/distributions/debian.py
rbuccigrossi: Helper added.

niceman/distributions/debian.py
rbuccigrossi: And again.

niceman/distributions/debian.py
rbuccigrossi: Now to see if my cure is worse than the original ailment...

niceman/distributions/debian.py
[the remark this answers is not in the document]
rbuccigrossi: I describe the helper below...

yarikoptic:
wow it is nice and fast now, if not only #102 and #138 .  Also reflected upon VCS tracing in #147 .
Thanks for the buzz in email.  Indeed, with all the callbacks feels a bit non Python -- I would have done it "closer" to the execute_command, thus avoiding also magical numbers in the `get_cmd_batch_len` call.  But that could be RFed later.

I am also confused by codecov reports/behavior, e.g. reporting that some `test_` tests aren't covered: https://codecov.io/gh/ReproNim/niceman/compare/d2132a5c42472117aafbe1ceffa500f65af9c5a5...ad6704a29e15600e2dcd89b2e430cc59509f2ef4/src/niceman/tests/test_utils.py  
A dummy one in #145 though doesn't show this behavior... so I wonder if that could be anyhow related to the introduced tagging?

let me dig a bit into this before merging...

PR #144, opened 2017-12-20
WIP: environment diff

chaselgrove:
Difference of environments/verification of environment.


chaselgrove:
Are we verifying files in any way?  If a required environment needs a file, it's currently just listed by file name, so a provided file could well be different from the one that was originally used.  Do we have plans to keep hashes of the extra files or something similar?

niceman/interface/diff.py
```diff
@@ -0,0 +1,150 @@
...
+            args=("--env",),
+            doc="NICEMAN environment specification", 
+            metavar='env',
+            constraints=EnsureStr()),
+        req=Parameter(
+            args=("--req",),
+            metavar="req",
+            doc="NICEMAN requirements", 
+            constraints=EnsureStr())
+    )
```
yarikoptic: what use-cases do you envision while having those as --arguments?
The simplest use case I see, is very similar to a regular 'diff' command -- `diff [options] spec1 spec2`
Seeing it this way also allows you to simplify code -- consider `specs`  to be a list! 
`__call__(*specs)`
and avoid code duplication you have now via list comprehensions and/or a simple loop.

PS Thinking this way could eventually also lead to another generalization: arbitrary number of specs!  e.g. I collected a bunch of specs in a directory.  Are they different? how?  but that is for later
PSS It seems that my mind is tuned to making everyone do `diff`, so you could see also a work of a new students who decided to help out with nibabel (while learning the "trade"):  `nib-ls diff` https://github.com/yarikoptic/nibabel/pull/3 , where diff'ing multiple niftis is on his agenda as well

niceman/interface/tests/test_diff.py
```diff
@@ -0,0 +1,41 @@
...
+def test_diff_different():
+    with swallow_logs(new_level=logging.DEBUG) as log:
+        args = ['diff', 
+                '--env', 
+                'niceman/interface/tests/files/xeyes_no_x11-apps.yaml', 
+                '--req', 
+                'niceman/interface/tests/files/xeyes.yaml'
+                ]
+        main(args)
+        assert_in('needed packages: 1', log.lines)
```
yarikoptic: from this test I see that you are aiming for a directionality, i.e. "satisfy" relationship, not just a pure diff (how two specs are different).
I think, to support that mode we should add `--satisfy` flag (later might reconsider for `--mode=(diff|satisfy|...smthelse)`.  But otherwise keep it as `spec1 spec2` and adjust documentation that "later is checked to satisfy previous", which would also scale up for multiple specs (again -- collected a bunch of them and want to see that next ones were always "enhanced" (new stuff added) but not really changed/upgraded -- i.e. that I could still satisfy requirements of the previous environment)

Makes sense?

niceman/interface/tests/test_diff.py
```diff
@@ -0,0 +1,41 @@
...
+xeyes_yaml = opj(dirname(__file__), 'files', 'xeyes.yaml')
+xeyes_no_x11_yaml = opj(dirname(__file__), 'files', 'xeyes_no_x11-apps.yaml')
+
+def test_diff_same():
+    with swallow_logs(new_level=logging.DEBUG) as log:
+        args = ['diff', 
+                '--env', 
+                'niceman/interface/tests/files/xeyes.yaml', 
+                '--req', 
+                'niceman/interface/tests/files/xeyes_no_x11-apps.yaml'
```
yarikoptic: with such long 'spec' files it is hardly possible to visually see what is different.  For such minimalistic test here -- why not to hand craft two tiny specs which would clearly demonstrate what you are trying to test here?

niceman/interface/diff.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @staticmethod
+    def __call__(env, req):
+
+        if not env:
+            raise InsufficientArgumentsError("env undefined")
+
+        if not req:
+            raise InsufficientArgumentsError("req undefined")
```
yarikoptic: when you refactor it to be a list -- just verify that `len(specs)==2`. also make a unittest to trigger that check

niceman/interface/tests/test_diff.py
yarikoptic: another `smthelse` here will be something like `compatible`: given a number of specs to determine if they are not in conflict. i.e. that they are all just a subset of a larger union of them all. And if union is not possible (e.g. in one Debian package X is version 1 and in the other version 2), show for which components, in which specs there is a difference.

niceman/interface/diff.py
```diff
@@ -0,0 +1,115 @@
...
+    def __call__(prov1, prov2):
+
+        env_1 = NicemanProvenance(prov1).get_environment()
+        env_2 = NicemanProvenance(prov2).get_environment()
+
+        for (dist_type, pkg_type) in ((DebianDistribution, "Debian package"), 
+                                      (CondaDistribution, "Conda package")):
+
+            dist_1 = env_1.get_distribution(dist_type)
+            pkgs_1 = { p._cmp_id: p for p in dist_1.packages }
```
kyleam: To try to test out this command, I ran `niceman retrace /usr/bin/diff > deb0.yml`.  Then I copied the file to deb1.yml and manually changed the version of the program.  Running `niceman diff deb0.yml deb1.yml` I get the following error

```
'NoneType' object has no attribute 'packages' [diff.py:__call__:68] (AttributeError)
```

In this case, this happens when get_distributions returns None for the conda dist_type.

chaselgrove: Good catch; thanks.

I fixed this in commit d52ecb7.  The semantics I'm using are that the distributions are nothing more than containers for packages (and are uninteresting in themselves), so the report only shows the differences in contained packages.  Try with:

    niceman diff niceman/interface/tests/files/empty.yaml niceman/interface/tests/files/diff_1.yaml

What do you think?

kyleam: LGTM

chaselgrove:
Yes, let's merge this one.

PR #153, opened 2018-01-10
ENH: New generator execute_command_batch() to batch arguments of shell commands

rbuccigrossi:
A (hopefully) more pythonic implementation of shell command batching using a generator "execute_command_batch()".  I admit that it makes the implementation easier to follow...

### DIFF
--- a/niceman/distributions/debian.py
+++ b/niceman/distributions/debian.py
@@ -33,7 +33,8 @@ from niceman.support.distributions.debian import \
     get_spec_from_release_file
 
 # Pick a conservative max command-line
-from niceman.utils import get_cmd_batch_len, execute_command_batch
+from niceman.utils import get_cmd_batch_len, execute_command_batch, \
+    cmd_err_filter
 
 # To parse output of dpkg-query
 import re
@@ -357,19 +358,14 @@ class DebTracer(DistributionTracer):
 
     def _get_packagefields_for_files(self, files):
         # Call dpkg query in batches
-        exec_gen = execute_command_batch(self._session,
-                                         ['dpkg-query', '-S'],
-                                         files,
-                                         CommandError)
+        exec_gen = execute_command_batch(
+            self._session, ['dpkg-query', '-S'], files,
+            cmd_err_filter('no path found matching pattern'))
         # Parse and accumulate stat results in a dict
         file_to_package_dict = {}
         for (out, _, exc) in exec_gen:
             if exc:
-                stderr = utils.to_unicode(exc.stderr, "utf-8")
-                if 'no path found matching pattern' in stderr:
-                    out = exc.stdout  # One file not found, so continue
-                else:
-                    raise exc  # some other fault -- handle it above
+                out = exc.stdout  # One file not found, so continue
             # Now go through the output and assign packages to files
             for outline in out.splitlines():
                 # Parse package name (architecture) and path
@@ -532,19 +528,14 @@ class DebTracer(DistributionTracer):
         queries = [self._pkg_name_to_dpkg_list_file(p["name"])
                    for p in pkg_dicts]
         # Call stat in batches
-        exec_gen = execute_command_batch(self._session,
-                                         ['stat', '-c', '%n: %Y'],
-                                         queries,
-                                         CommandError)
+        exec_gen = execute_command_batch(
+            self._session, ['stat', '-c', '%n: %Y'], queries,
+            cmd_err_filter('No such file or directory'))
         # Parse and accumulate stat results in a dict
         results = {}
         for (out, _, exc) in exec_gen:
             if exc:
-                stderr = utils.to_unicode(exc.stderr, "utf-8")
-                if 'No such file or directory' in stderr:
-                    out = exc.stdout  # One file not found, so continue
-                else:
-                    raise exc  # some other fault -- handle it above
+                out = exc.stdout  # One file not found, so continue
             # Parse the output and store by filename
             for outlines in out.splitlines():
                 (fname, ftime) = outlines.split(": ")
--- a/niceman/distributions/debian.py
+++ b/niceman/distributions/debian.py
@@ -31,7 +31,7 @@ from niceman.support.distributions.debian import \
     get_spec_from_release_file
 
 # Pick a conservative max command-line
-from niceman.utils import get_cmd_batch_len, batch_process_list
+from niceman.utils import get_cmd_batch_len, execute_command_batch
 
 # To parse output of dpkg-query
 import re
@@ -354,11 +354,21 @@ class DebTracer(DistributionTracer):
         yield dist, remaining_files
 
     def _get_packagefields_for_files(self, files):
-        # Define a function to process batches of files and return
-        # a dictionary that maps files to packages
-        def proc_dpkg_query(subfiles, file_to_package_dict):
-            out = self._run_dpkg_query(subfiles)
-
+        # Call dpkg query in batches
+        exec_gen = execute_command_batch(self._session,
+                                         ['dpkg-query', '-S'],
+                                         files,
+                                         CommandError)
+        # Parse and accumulate stat results in a dict
+        file_to_package_dict = {}
+        for (out, _, exc) in exec_gen:
+            if exc:
+                stderr = utils.to_unicode(exc.stderr, "utf-8")
+                if 'no path found matching pattern' in stderr:
+                    out = exc.stdout  # One file not found, so continue
+                else:
+                    raise exc  # some other fault -- handle it above
+            out = utils.to_unicode(out, "utf-8")
             # Now go through the output and assign packages to files
             for outline in out.splitlines():
                 # Parse package name (architecture) and path
@@ -379,12 +389,7 @@ class DebTracer(DistributionTracer):
                 lgr.debug("Identified file %r to belong to package %s",
                           found_name, pkg)
                 file_to_package_dict[found_name] = pkg
-            return file_to_package_dict
-
-        # Determine how big our batches can be
-        num_files = get_cmd_batch_len(files, 13)
-        # Now process our file list and return the file to package mapping
-        return batch_process_list(proc_dpkg_query, files, num_files, {})
+        return file_to_package_dict
 
     def _get_apt_source_name(self, src):
         # Create a unique name for the origin
@@ -457,20 +462,6 @@ class DebTracer(DistributionTracer):
                     archive_uri=src_vals.get("archive_uri"))
 
     def _get_pkgs_arch_and_version(self, pkg_dicts):
-        # Define a function to process batches of packages and return
-        # an array of parsed dpkg -s output
-        def proc_dpkg_cmd(subqueries, cmd_results):
-            try:
-                out, _ = self._session.execute_command(
-                    ['dpkg', '-s'] + subqueries
-                )
-                out = utils.to_unicode(out, "utf-8")
-                # dpkg -s uses the same output as apt-cache show pkg
-                cmd_results += parse_apt_cache_show_pkgs_output(out)
-            except CommandError as _:
-                raise  # Currently raise any exception we encounter
-            return cmd_results
-
         # Convert package names to name:arch format
         queries = []
         for p in pkg_dicts:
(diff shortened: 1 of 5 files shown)


niceman/distributions/debian.py
```diff
@@ -457,31 +462,22 @@ def _find_all_sources(self):
...
-        results = batch_process_list(proc_dpkg_cmd, queries, num_files, [])
+        # Call "dpkg -s" in batches
+        exec_gen = execute_command_batch(self._session, ['dpkg', '-s'],
+                                         queries)
+        # Parse and accumulate "dpkg -s" results in an array
+        results = []
+        for (out, _, _) in exec_gen:
+            out = utils.to_unicode(out, "utf-8")
+            # dpkg -s uses the same output as apt-cache show pkg
+            results += parse_apt_cache_show_pkgs_output(out)
```
yarikoptic: all good, but if you want to become a really pythonic fella you could express yourself via a list comprehension.
```python
results = [
   parse_apt_cache_show_pkgs_output(utils.to_unicode(out, "utf-8"))
   for (out, _, _) in exec_gen
]
```
and then a very pythonic fella would have made it into a generator ( `()` instead of `[]`) to not even incur temporary list construction if  `self.create_lookup_from_apt_cache_show` could swallow it ;)   Benefit is not as much as shorter code as making it clear right away that result is a list which contains those items.

some obnoxious pythonista who came from functional languages could use 
```python
results = map(
  lambda out: parse_apt_cache_show_pkgs_output(utils.to_unicode(outs[0], "utf-8")),
  exec_gen
)
```
but imho it is ulgyish due to need of using lambda here ... if only we decoded outputs right away within execut_command_batch, then it could be just 
```python
results = map(parse_apt_cache_show_pkgs_output, exec_gen)
```
nice and clear.  Note: map in python2 returns a list and in python3 a generator.  but you could use `from six.moves import map` to guarantee it being a generator ;-)

niceman/distributions/debian.py
```diff
@@ -505,28 +501,18 @@ def create_lookup_from_apt_cache_show(cmd_results):
...
-        results = batch_process_list(proc_apt_cache_show_cmd, queries,
-                                     num_files, [])
+        # Call "apt-cache show" in batches
+        exec_gen = execute_command_batch(self._session, ['apt-cache', 'show'],
+                                         queries)
+        # Parse and accumulate "apt-cache show" results in an array
+        results = []
+        for (out, _, _) in exec_gen:
+            out = utils.to_unicode(out, "utf-8")
+            results += parse_apt_cache_show_pkgs_output(out)
```
yarikoptic: yet another place where a `map` would look beautiful ;-)

niceman/distributions/debian.py
```diff
@@ -543,35 +529,32 @@ def proc_apt_cache_show_cmd(subqueries, cmd_results):
...
-        def proc_stat_cmd(subqueries, cmd_results):
-            try:
-                out, _ = self._session.execute_command(
-                    ['stat', '-c', '%n: %Y'] + subqueries
-                )
-            except CommandError as exc:  # file not found
+        # Convert package names to dpkg list filenames
+        queries = []
+        for p in pkg_dicts:
+            queries.append(self._pkg_name_to_dpkg_list_file(p["name"]))
```
yarikoptic: use list (or generator) comprehension please

niceman/distributions/debian.py
```diff
@@ -543,35 +529,32 @@ def proc_apt_cache_show_cmd(subqueries, cmd_results):
...
+        results = {}
+        for (out, _, exc) in exec_gen:
+            if exc:
                 stderr = utils.to_unicode(exc.stderr, "utf-8")
                 if 'No such file or directory' in stderr:
                     out = exc.stdout  # One file not found, so continue
                 else:
-                    raise  # some other fault -- handle it above
+                    raise exc  # some other fault -- handle it above
+            out = utils.to_unicode(out, "utf-8")
```
yarikoptic: more I see of the same lines, more I am confident that probably should be a part of the `execute_command_batch` command

niceman/distributions/debian.py
kyleam: > you could express yourself via a list comprehension

If I'm reading it right, I think it'd take a little more work than
your example.  `results +=` is extending with a list rather than
appending an element, so there'd need to be flattening somewhere so
that your list comp didn't give a list of lists.

kyleam: The itertools docs also give `chain.from_iterable(listOfLists)` as a
snippet of flattening one level of nesting.

niceman/tests/test_utils.py
```diff
@@ -435,12 +422,40 @@ def test_hashable_dict():
...
+    # First let's do a simple test to count the args
+    args = list(map(str, range(1, 101)))
+    cmd_gen = execute_command_batch(session, [], args, None)
+    for (out, _, _) in cmd_gen:
+        assert out == 100
+    # Now let's raise an exception but not list it as handled
+    cmd_gen = execute_command_batch(session, ["ValueError"], args, None)
+    with pytest.raises(ValueError):
+        for (_, _, _) in cmd_gen:
+            pass
```
kyleam: This for loop could be reduced to `next(cmd_gen)`.

niceman/utils.py
```diff
@@ -1014,6 +986,50 @@ def get_cmd_batch_len(arg_list, cmd_len):
...
+    packages, but risk creating command-lines that are too long. This
+    function is a generator that will call execute_command but with batches
+    of arguments (to stay within the command-line length limit) and yield the
+    results.
+
+    Parameters
+    ----------
+    session
+      Session object that implements the execute_command() member
+    command : array
```
kyleam: I think `sequence` (preferably) or `list` should be used rather than
`array` (here and a few other places) to avoid confusion with arrays
from the stdlib or numpy.

niceman/tests/test_utils.py
yarikoptic: good call! it would indeed should be sufficient here!
but in general: `next` would give just a single element, while the loop would go through all. so if it was desired,  `list` instead of `next` should be used ;)

kyleam: > but in general

Sure, we only know `next` is sufficient in this particular case
because we constructed the one item list two lines above.  But from a
readability/intention standpoint, I agree it'd be better to use
`list`.

niceman/utils.py
rbuccigrossi: I started with "list" but then pycharm threw some PEP errors with my use of len().  I'll try sequence...

niceman/tests/test_utils.py
rbuccigrossi: Done.

niceman/utils.py
rbuccigrossi: Done.

niceman/distributions/debian.py
rbuccigrossi: Yup (this forced me to update my test_execute_command_batch because my mock returned an int....  Shame on me for making a mock that didn't match the function signature :/

niceman/distributions/debian.py
rbuccigrossi: So I had two cases: a sequence of dicts, and a sequence of lists.  To flatten the sequence of dicts I used: 

results = {k: v for d in results for k, v in d.items()}

niceman/distributions/debian.py
rbuccigrossi: Done.

niceman/distributions/debian.py
rbuccigrossi: Yeah, I have a couple of these.  Thanks!

rbuccigrossi:
This should be ready for a final review and merge :)

niceman/distributions/debian.py
```diff
@@ -354,11 +356,20 @@ def identify_distributions(self, files):
...
+                                         CommandError)
+        # Parse and accumulate stat results in a dict
+        file_to_package_dict = {}
+        for (out, _, exc) in exec_gen:
+            if exc:
+                stderr = utils.to_unicode(exc.stderr, "utf-8")
+                if 'no path found matching pattern' in stderr:
+                    out = exc.stdout  # One file not found, so continue
+                else:
+                    raise exc  # some other fault -- handle it above
```
yarikoptic: Eh -- my bad I think -- we wouldn't be to reraise the exception this way -- it would loose original exception throwing location:
```shell
$> nl -b a /tmp/1.py
     1	
     2	def fun():
     3	  raise Exception("buga")
     4	
     5	try:
     6	   fun()
     7	except Exception as exc:
     8	   vexc = exc
     9	
    10	# now we are here "reraising" it and information will be local
    11	raise vexc

$> python /tmp/1.py
Traceback (most recent call last):
  File "/tmp/1.py", line 11, in <module>
    raise vexc
Exception: buga
```
correct way would  be for us  to return exception info so it could be "properly" reraised:
```shell
$> nl -b a /tmp/1proper.py 
     1	import sys        
     2	from six import reraise
     3	
     4	def fun():
     5	  raise Exception("buga")
     6	
     7	try:
     8	   fun()
     9	except Exception as exc:
    10	   exc_info = sys.exc_info()
    11	
    12	# now we are here "reraising" it and information will be local
    13	reraise(*exc_info)

$> python /tmp/1proper.py 
Traceback (most recent call last):
  File "/tmp/1proper.py", line 13, in <module>
    reraise(*exc_info)
  File "/tmp/1proper.py", line 8, in <module>
    fun()
  File "/tmp/1proper.py", line 5, in fun
    raise Exception("buga")
Exception: buga
```
The `[1]` of that tuple is the exception itself, so you could access its attributes (e.g., `.stderr`).
feel free to postpone this to a separate PR

niceman/distributions/debian.py
```diff
@@ -584,28 +564,17 @@ def _pkg_name_to_dpkg_list_file(name):
...
+        queries = ["%(name)s:%(architecture)s" % p for p in pkg_dicts]
+        # Call apt-cache policy in batches
+        exec_gen = execute_command_batch(self._session,
+                                         ['apt-cache', 'policy'],
+                                         queries)
+        # Parse results into a single generator
+        results = (parse_apt_cache_policy_pkgs_output(out)
+                   for (out, _, _) in exec_gen)
+        # Combine sequence of dicts
+        results = {k: v for d in results for k, v in d.items()}
```
yarikoptic: IMHO it is worth a separate helper function with a description/test of the expected operation when (for some reason) the same `k` happens in multiple entries.  ATM it just replaces with a "new" value.
In our case here it would be unexpected and should throw an Exception (e.g., a `RuntimeError("Have obtained a new value %r for %r where previous known was %r" % (v, k, results[k]))`).  I know that it shouldn't happen, but again -- those damn computers some times raise a fight against us and we better do not miss their sneaky spies (probably I am reading too much of sci-fi recently, currently Avogadro Corp trilogy)

niceman/distributions/debian.py
rbuccigrossi: It appears that raising an exception properly with past history changed from Python 2 to 3. However, if I update command_batch to be more explicit in what exceptions we accept (specifically by passing in a filter-like function), then we can have the generator throw those we don't accept.  The generator will then be able to use the no argument raise keyword to pass the exception gracefully.

Let me give this a go and see what you think.

rbuccigrossi: I'm still not resolving all the issues, but I committed the fix to address the handling of exceptions.

kyleam: > It appears that raising an exception properly with past history changed from Python 2 to 3.

Doesn't `six.reraise` let use work around this?

rbuccigrossi: Nice.  I didn't see that.  My current approach avoids the issue (as long as ya'll are OK with it).

niceman/distributions/debian.py
rbuccigrossi: Done.  (And trust no one. :)

niceman/utils.py
```diff
@@ -986,7 +988,47 @@ def get_cmd_batch_len(arg_list, cmd_len):
     return max((_MAX_LEN_CMDLINE - cmd_len) // (max_len + 1), 1)
 
 
-def execute_command_batch(session, command, args, expected_exceptions=None):
+def join_sequence_of_dicts(seq):
+    """
+    Joins a sequence of dicts into a single dict, trows RuntimeError if dupe
```
yarikoptic:
```
From WordNet (r) 3.0 (2006) [wn]:

  dupe
      n 1: a person who is tricked or swindled [syn: {victim}, {dupe}]
      v 1: fool or hoax; 
```

rbuccigrossi: "dupe" = short for "duplicate" -- The joys of trying to fit things in 80 characters :)

rbuccigrossi: The Oxford English dictionary does recognize "dupe" as short for "duplicate": https://en.oxforddictionaries.com/definition/dupe#dupe-2

kyleam: Another option: End the summary sentence at dict and then add

```
Raises
------
RuntimeError if a duplicate key is encountered.
```

to the end of the docstring.

rbuccigrossi: @kyleam for the win!  Updated...

Issue #151, opened 2018-01-08
conda retracing might crash


yarikoptic:
```shell
$> git describe
v0.0.5

$> niceman retrace  ~/anaconda3/envs/py3.4.2/bin/elbadmin      
2018-01-08 09:01:16,507 [INFO   ] Tracing using <class 'niceman.distributions.debian.DebTracer'> 
2018-01-08 09:01:16,991 [INFO   ] DebTracer: 0 packages with 0 files, and 1 other files 
2018-01-08 09:01:16,991 [INFO   ] Tracing using <class 'niceman.distributions.conda.CondaTracer'> 
2018-01-08 09:01:17,005 [INFO   ] Detected conda /home/yoh/anaconda3/envs/py3.4.2 
2018-01-08 09:01:18,252 [WARNING] Could not retrieve conda environment export from path /home/yoh/anaconda3/envs/py3.4.2: CommandError: command '/home/yoh/anaconda3/bin/conda-env export -p /home/yoh/anaconda3/envs/py3.4.2' failed with exitcode 2.
Failed to run u'/home/yoh/anaconda3/bin/conda-env export -p /home/yoh/anaconda3/envs/py3.4.2' under None. Exit code=2. out= err=usage: conda-env [-h] {attach,create,export,list,remove,upload,update} ...
conda-env: error: unrecognized arguments: -p /home/yoh/anaconda3/envs/py3.4.2
 [cmd.py:run:297] 
2018-01-08 09:01:18,349 [ERROR  ] 'NoneType' object is not iterable [conda.py:_get_conda_pip_package_details:191] (TypeError) 
```

yarikoptic:
oh right, here there is the `-p` issue we talked about.  We already issue a warning, so I think it should just proceed with an additional explanation (outdated conda), and without crashing

PR #157, opened 2018-01-13
BF: Detect and gracefully handle old conda versions

rbuccigrossi:
To address #151 - we check stderr for the warning on "-p", ask users to use Conda 4.3.19 or greater if the error occurs, and better handle the result so we don't crash (and detect as much as possible)


yarikoptic:
Can't merge on the phone since the new code is not test covered ... should be trivial to test with a mock or two ;-)

rbuccigrossi:
I have unit tests for the particular exceptions...  I think you'll have more to review with the unit tests than my original fix :)

yarikoptic:
awesome, thanks!

PR #159, opened 2018-01-15
RF: Refactor setup_ssh into a generic get_docker_fixture

yarikoptic:
Since @mjtravers  stopped by with a question, we have decided to make it happen so it could be reused in other tests.

The only side-effect we see ATM is how to skip for a specific types of tests -- be careful to use decorators such as `skip_ssh` around the generated fixture since would result in skipping the entire tests module.  Otherwise should be quite flexible


yarikoptic:
@rbuccigrossi what do you think about this one?

rbuccigrossi:
@yarikoptic I'm not clear as to the goal of my assignment.  In terms of understanding the code, the only thing that made me think hard was something that both you and Matt did: use the behavior of yield to tear down the docker when you were done. 

So if I'm understanding the code correctly, the assumption of the generator is that the caller will use the docker before calling next again. That is a "cute" way to ensure cleanup (unless there is an exception).

kyleam:
> the only thing that made me think hard was something that both you and Matt did: use the behavior of yield to tear down the docker when you were done.

Yeah, that's [one of the ways][ps] to specify teardown code with
pytest.  The teardown will happen even if an exception occurs in the
test code, but it won't be executed if an exception occurs in the
setup code (to ensure that, you can use `addfinalizer`).

[ps]: https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

yarikoptic:
thanks @kyleam for the clarification and @rbuccigrossi for the feedback.  since no objections/recommendations, will merge

yarikoptic:
ah, and if the "assignment" was not clear @rbuccigrossi  -- it was "to provide code review/feedback" ;) also, a way to make you aware of new functionality in the codebase

rbuccigrossi:
Thank you... will do.

PR #165, opened 2018-01-23
Do not store None values for url or pushurl

### LINKED ISSUE #105: git tracer includes meaningless pushurl: null in generated output spec

yarikoptic:
Ideally we might just make up a Remote(SpecObject) class, instead of
ad-hoc dict there and then it would not require custom checking there

Closes #105

### DIFF
--- a/niceman/interface/tests/test_retrace_packages.py
+++ b/niceman/interface/tests/test_retrace_packages.py
@@ -38,9 +38,15 @@ def test_identify_myself():
     distributions, files = identify_distributions([__file__, '/nonexisting-for-sure'])
     assert len(distributions) == 1
     assert distributions[0].name == 'git'
-    assert distributions[0].packages[0].files == [__file__]
+    repo = distributions[0].packages[0]
+    assert repo.files == [__file__]
 
     assert files == ['/nonexisting-for-sure']
+    # there should be None rurl's or pushurl's
+    for remote in repo.remotes.values():
+        for f in 'url', 'pushurl':
+            if f in remote:
+                assert remote[f] is not None  # we do not store None's
 
 
 @with_tempfile(mkdir=True)
--- a/niceman/distributions/vcs.py
+++ b/niceman/distributions/vcs.py
@@ -360,12 +360,12 @@ class GitRepoShim(GitSVNRepoShim):
         # which remotes contain this commit, so we could provide this
         # possibly valuable information
         if not hexsha:  # just initialized
-            return []
+            return {}
         remote_branches = self._run_git(
             'branch -r --contains %s' % hexsha,
             expect_fail=True)
         if not remote_branches:
-            return []
+            return {}
         containing_remotes = set(x.split('/', 1)[0] for x in remote_branches)
         remotes = {}
         for remote in self._run_git('remote').splitlines():
--- a/niceman/distributions/vcs.py
+++ b/niceman/distributions/vcs.py
@@ -87,7 +87,7 @@ class GitRepo(VCSRepo):
     hexsha = attr.ib(default=None)
     describe = attr.ib(default=None)
     tracked_remote = attr.ib(default=None)
-    remotes = attr.ib(default=attr.Factory(list))
+    remotes = attr.ib(default=attr.Factory(dict))
 
 # Probably generation wouldn't be flexible enough
 #GitDistribution = get_vcs_distribution(GitRepo, 'git', 'Git')
--- a/niceman/distributions/vcs.py
+++ b/niceman/distributions/vcs.py
@@ -372,10 +372,12 @@ class GitRepoShim(GitSVNRepoShim):
             rec = {}
             for f in 'url', 'pushurl':
                 try:
-                    rec[f] = self._run_git('config remote.%s.%s' % (remote, f)
-                                           , expect_fail=True
-                                           #, expect_stderr=True
-                                           )
+                    v = self._run_git('config remote.%s.%s' % (remote, f)
+                                      , expect_fail=True
+                                      #, expect_stderr=True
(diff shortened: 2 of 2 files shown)


niceman/interface/tests/test_retrace_packages.py
```diff
@@ -38,9 +38,15 @@ def test_identify_myself():
     distributions, files = identify_distributions([__file__, '/nonexisting-for-sure'])
     assert len(distributions) == 1
     assert distributions[0].name == 'git'
-    assert distributions[0].packages[0].files == [__file__]
+    repo = distributions[0].packages[0]
+    assert repo.files == [__file__]
 
     assert files == ['/nonexisting-for-sure']
+    # there should be None rurl's or pushurl's
+    for remote in repo.remotes.values():
```
kyleam: Here's the failure:

```
=================================== FAILURES ===================================
_____________________________ test_identify_myself _____________________________

args = (), kwargs = {}

    @wraps(func)
    def newfunc(*args, **kwargs):
        if cond:
            raise SkipTest(msg if msg else "condition was True")
>       return func(*args, **kwargs)

niceman/tests/utils.py:381:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    @skip_if(not lexists(opj(dirname(__file__), pardir, pardir, pardir, '.git')))
    def test_identify_myself():
        distributions, files = identify_distributions([__file__, '/nonexisting-for-sure'])
        assert len(distributions) == 1
        assert distributions[0].name == 'git'
        repo = distributions[0].packages[0]
        assert repo.files == [__file__]

        assert files == ['/nonexisting-for-sure']
        # there should be None rurl's or pushurl's
>       for remote in repo.remotes.values():
E       AttributeError: 'list' object has no attribute 'values'

niceman/interface/tests/test_retrace_packages.py:46: AttributeError
```

I'm guessing this is because `GitRepoShim.remotes` [returns an
empty][r] list when there are no remotes configured.  Shouldn't it
return an empty dict instead?

[r]: https://github.com/ReproNim/niceman/blob/1b4fe1a1060b6d60c836c2b3e7b297d744282005/niceman/distributions/vcs.py#L367

yarikoptic: THANKS @kyleam ! pushed the fix

kyleam: Code-wise, I think that's correct, but, since remotes aren't
configured on travis, I suppose your test isn't actually getting
executed.

yarikoptic: they used to be I believe... anyways -- I need to RF (later) all these tests to operate on custom git repos, where I know for sure what is going on

kyleam: And the same failure after 3e79a5a :/

The next placed I'd guess a list would be coming from is [here].  Hard
to debug since I've yet to reproduce it locally.  I'll set up a local
directory without remotes and see if I can trigger it.

[here]: https://github.com/ReproNim/niceman/blob/3e79a5abd7946fc914978786e63b57f6673b26e5/niceman/distributions/vcs.py#L90

kyleam: OK, I can reproduce locally in a repo with no remotes and

```diff
diff --git a/niceman/distributions/vcs.py b/niceman/distributions/vcs.py
index 7c5834a..6b3f1e5 100644
--- a/niceman/distributions/vcs.py
+++ b/niceman/distributions/vcs.py
@@ -87,7 +87,7 @@ class GitRepo(VCSRepo):
     hexsha = attr.ib(default=None)
     describe = attr.ib(default=None)
     tracked_remote = attr.ib(default=None)
-    remotes = attr.ib(default=attr.Factory(list))
+    remotes = attr.ib(default=attr.Factory(dict))

 # Probably generation wouldn't be flexible enough
 #GitDistribution = get_vcs_distribution(GitRepo, 'git', 'Git')
```

can fix it.  Then we should be back to the case where tests are
passing but the assert statements aren't getting executed.

yarikoptic: d'oh -- I should have spotted it... THANKS! pushed the "fix".  BTW you should be able to push to my branch, so do not hesitate next time if such obvious fixes emerge

kyleam: Sure, will do.  My overall grasp of how parts in vcs.py interact
(particularly the shims) is shaky enough that my suggestion wasn't
obviously correct to me.

kyleam: A bit more info about why this test is a noop on travis:

The travis clone operations boil down to

```
$ git clone --depth=50 https://github.com/ReproNim/niceman.git ReproNim/niceman
$ git fetch origin +refs/pull/165/merge:
$ git checkout -qf FETCH_HEAD
```

So at this point, we're in a detached head for the merge between the
PR tip and the master.  `GitSVNRepoShim.remotes` returns an empty dict
if `git branch -r --contains HEAD` doesn't return any branch, and it
never does on travis because no remote branches contain the PR/master
merge.

Of course, this isn't true locally, so the assert in this PR's test is
executed.

> I need to RF (later) all these tests to operate on custom git repos, where I know for sure what is going on

Right, I'd say this PR is good to go then.  Tests pass locally and
remotely.  Meaningful tests are only executed locally ATM, but they'll
be executed on travis as well once the RF mentioned above is complete.

yarikoptic:
with @kyleam blessing will merge this one since the world better with it than without ;)  otherwise -- more work to be done indeed!

Issue #163, opened 2018-01-23
conda tracer will drop pip package locations, depending on pip/conda versions


kyleam:
The output of `pip list --format=legacy` looks like this:

```
pkg1 (X.Y.Z)
pkg2 (X.Y.Z, /path/to/local/repo)
```

where the second entry is a package that was installed with
`--editable`.  The conda tracer recognizes these paths and adds them
as `origin_location` attributes.

Unfortunately, the path isn't always included in `conda-env export`
output that the tracer processes:

  * the exact list format that `conda-env export` processes underneath
    depends on the conda version.  Starting with conda v4.4.0
    (2017-12-20), it uses `pip list --format=json` unconditionally.
    As of v4.4.3, it uses json as long as pip is at version 9 (older
    versions of pip don't support `--format`).  This matters because...

  * `pip list --format=json` does not include the location information
    for pip v9.  In the development branch, it is available when the
    `-v` flag is added: 760ae144 (Made list command to respond to '-v'
    option, 2017-04-02).

Instead of relying on the pip packages in the `conda-env export`
output, the conda tracer should call `pip list --format=legacy`
directly so that locations are always included with editable packages.
(The virtualenv tracer in #162 has code to do this, so that should be
pulled out for reuse in the conda tracer.)

PR #162, opened 2018-01-22
ENH: virtualenv tracing

kyleam:
This adds basic support for tracing virtualenvs.

WIP bits:

  * [x] This is on top of #153 (merged now) because it uses batched `pip show`
    calls. It should be rebased once that is merged.

    Instead of this PRs diff, here's the more useful
    [comparison](https://github.com/ReproNim/niceman/compare/c7783b623e67ec2af15b37c0bfa660eefa5ed3c7...kyleam:enh-venv-tracer).

  * [x] More tests are needed for the tracers as well as the added helpers.

Here's an example of the output:

```
$ niceman retrace ~/src/python/pyout/venv-py3/lib/python3.5/site-packages/blessings/tests.py
2018-01-22 13:41:54,826 [INFO   ] Tracing using <class 'niceman.distributions.debian.DebTracer'>
2018-01-22 13:41:54,897 [INFO   ] DebTracer: 0 packages with 0 files, and 1 other files
2018-01-22 13:41:54,897 [INFO   ] Tracing using <class 'niceman.distributions.conda.CondaTracer'>
2018-01-22 13:41:54,934 [INFO   ] CondaTracer: 0 packages with 0 files, and 1 other files
2018-01-22 13:41:54,934 [INFO   ] Tracing using <class 'niceman.distributions.venv.VenvTracer'>
2018-01-22 13:41:56,954 [INFO   ] VenvTracer: 15 packages with 1 files, and 0 other files
2018-01-22 13:41:57,234 [INFO   ] Tracing using <class 'niceman.distributions.vcs.VCSTracer'>
# NICEMAN Environment Configuration File
# This file was created by NICEMAN 0.0.5 on 2018-01-22 13:41:57.235499
version: 0.0.1
distributions:
- name: venv
  path: /usr/bin/virtualenv
  venv_version: 15.1.0
  environments:
  - name: venv-py3
    path: /home/kyle/src/python/pyout/venv-py3
    python_version: 3.5.3
    packages:
    - {name: wheel, version: 0.30.0}
    - {name: pkg-resources, version: 0.0.0}
    - {name: pytest-timeout, version: 1.2.1}
    - {name: pbr, version: 3.1.1}
    - {name: pyout, version: 0.1.0, origin_location: /home/kyle/src/python/pyout}
    - {name: py, version: 1.5.2}
    - {name: pytest, version: 3.3.2}
    - name: blessings
      version: 1.6.1
      files: [lib/python3.5/site-packages/blessings/tests.py]
[...]

PR #168, opened 2018-02-01
ENH+RF: Share 'pip {list,show}' code between venv and conda tracers

kyleam:
Adjust the conda tracer to reuse the batched `pip show` from the venv
tracer (#162).

Note that this is not ready to be merged because it depends on #162.
Currently this PR is against a temporary branch rather than master so
that the diff doesn't include the commits from #162.

### DIFF
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -21,7 +21,7 @@ from .base import SpecObject
 from .base import DistributionTracer
 from .base import Package
 from .base import TypedList
-from .piputils import pip_show
+from .piputils import pip_show, pip_packages
 from niceman.dochelpers import exc_str
 from niceman.utils import PathRoot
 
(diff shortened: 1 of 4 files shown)


kyleam:
Underneath the batching is done using @rbuccigrossi's
`execute_command_batch` from #153, which works nicely.  Thanks :)

kyleam:
The two commits just pushed broaden the scope of this issue a bit in
order to fix #163.

niceman/distributions/conda.py
```diff
@@ -157,57 +157,18 @@ def _get_conda_package_details(self, conda_path):
...
-                            "export from path %s: %s", conda_path,
-                            exc_str(exc))
-                continue
-
+    def _get_conda_pip_package_details(self, conda_path):
+        pip = conda_path + "/bin/pip"
+        try:
+            pkgs = list(pip_packages(self._session, pip))
+        except Exception as exc:
+            lgr.warning("Could determine pip packages for %s: %s",
```
yarikoptic: I guess "Could **not** determine..."?

niceman/distributions/venv.py
```diff
@@ -72,26 +71,14 @@ def _create_package(self, **package_fields):
...
-            for path in info["Files"]:
-                full_path = os.path.normpath(
-                    os.path.join(info["Location"], path))
-                file_to_pkg[full_path] = pkg
-        return packages, file_to_pkg
+        pip = venv_path + "/bin/pip"
+        try:
+            pkgs = list(pip_packages(self._session, pip))
+        except Exception as exc:
+            lgr.warning("Could determine pip packages for %s: %s",
```
yarikoptic: the same "Could **not**"

niceman/distributions/piputils.py
```diff
@@ -38,6 +42,50 @@ def parse_pip_show(out):
...
 
 
+def _pip_batched_show(session, which_pip, pkgs):
+    cmd = [which_pip, "show", "-f"]
+    batch = execute_command_batch(session, cmd, pkgs)
+    entries = (stacked.split("---") for stacked, _, _ in batch)
+
+    for pkg, entry in zip(pkgs, itertools.chain(*entries)):
+        info = parse_pip_show(entry)
+        yield pkg, info
```
yarikoptic: I would have appreciated a unittest for this function which mocks away `execute_command_batch` and demonstrate that it does work e.g. when output for multiple pkgs in a batch is output (on some dummified but still legit).  That could allow others comprehend on what is going on here (how pip show -f output should look like etc), if things start to break (e.g. `pip show -f` format changes or smth odd in it, like someone using `---` in the Summary or a url for some reason ;-))

niceman/distributions/conda.py
kyleam: guh, thanks

yarikoptic:
the `---` will be dealt with later, merging

kyleam:
> the --- will be dealt with later, merging

Addressed with #180.

Issue #139, opened 2017-11-30
Instantiation of conda environments from the spec


yarikoptic:
Now that we collect all needed (?) information, time to instantiate them. For starters 1:1 in terms of the location/path, but later we would need a config setting for paths remapping (e.g. if bucc traces on his box, I should still be able to instantiate on mine).  When we get there, we will see what other possible pieces it could potentially touch upon

kyleam:
Closed by #174.

PR #174, opened 2018-02-05
ENH: Conda generation from spec

rbuccigrossi:
This pull request generates conda environments from specs. It properly sets the conda and python versions from the environments. In addition I combined the install and detection tests into a single round trip.  I believe it is ready, so I'm removing the "WIP" from the title.

The key test (as well as the large Debian detection test for /usr/bin/* files) is now considered an "integration" test and requires "--integration" on the py.test line to run. I did add "--integration" to the py.test call on .travis.yml so that the integration tests are run.

This pull request addresses #139 .


niceman/distributions/conda.py
```diff
@@ -92,13 +92,43 @@ def install_packages(self, session=None):
...
+        export = "name: %s\n" % name
+        # Collect channels
+        export += "channels:\n"
+        for c in env.channels:
+            export += "- %s\n" % c["name"]
+        export += "dependencies:\n"
+        # Collect packages (separating pip installed packages)
+        pip_export = ""
+        for p in env.packages:
+            if p.get("installer") is None:
```
kyleam: I think you'd more commonly see this check as `"installer" in p` or as
`p["install"]` with a `KeyError` catch.

niceman/distributions/conda.py
```diff
@@ -92,13 +92,43 @@ def install_packages(self, session=None):
...
+                export += "- %s=%s=%s\n" % (p["name"], p["version"],
+                                            p["build"])
+            if p.get("installer") == "pip":
```
kyleam: The only "installer" that's present in the environments is "pip",
correct?  Should we add an `else` branch for the `p["installer"] != "pip"`
case that logs a warning if another key is encountered?

So something like

```python
if "installer" not in p:
    ...
elif p["installer"] == "pip":
    ...
else:
    ...
```

niceman/distributions/conda.py
```diff
@@ -92,13 +92,43 @@ def install_packages(self, session=None):
...
+        export += "channels:\n"
+        for c in env.channels:
+            export += "- %s\n" % c["name"]
+        export += "dependencies:\n"
+        # Collect packages (separating pip installed packages)
+        pip_export = ""
+        for p in env.packages:
+            if p.get("installer") is None:
+                export += "- %s=%s=%s\n" % (p["name"], p["version"],
+                                            p["build"])
```
kyleam: Purely a personal preference, but with `.format()`, this could be

```python
export += "- {name}={version}={build}".format(**p)
```

niceman/distributions/conda.py
```diff
@@ -92,13 +92,43 @@ def install_packages(self, session=None):
...
+
         return
 
+    @staticmethod
+    def create_conda_export(env):
+        # TODO: The environment name should be discovered on retrace
+        name = os.path.basename(os.path.normpath(env.path))
+        export = "name: %s\n" % name
+        # Collect channels
+        export += "channels:\n"
```
kyleam: Depending on the number of packages that are present, the final export
string may be a result of a good number of string concatenations.  An
alternative to the string concatenation would be to build up a list of
strings (sans the new lines) and then `"\n".join(export_lines)`.

niceman/distributions/conda.py
```diff
@@ -92,13 +92,43 @@ def install_packages(self, session=None):
...
-        # for package in self.provenance['packages']:
-        #     session.add_command(['conda',
-        #                            'install',
-        #                            package['name']])
+        # TODO: Install Conda
+#        print("We got" + str(environment))
+        # Loop through non-root packages, creating the conda-env export
+        # file to import
+        for env in self.environments:
+            if env.path == self.path:
```
yarikoptic: so we never populate packages into the root env?

niceman/distributions/conda.py
```diff
@@ -92,13 +92,43 @@ def install_packages(self, session=None):
...
+                continue
+            export_contents = self.create_conda_export(env)
+            print ("EXPORT\n" + export_contents)
```
yarikoptic: `lgr.debug` or it was intended to be displayed to the user?

niceman/distributions/conda.py
yarikoptic: IMHO it should check `version` and `build` on being defined.   would it tollerate `None` or should they be provided an alternative -- an empty string (i.e. `or ""`) so if not defined -- it might then look like `- pkg==` (is that legit?)

niceman/distributions/conda.py
rbuccigrossi: I've been spoiled with mutable strings. I'll make the switch.

niceman/distributions/conda.py
rbuccigrossi: I made the change but then I realized: "installer" actually is in p, but set to None...  So I'm not checking for existence but actual definition (we have a number of mandatory fields in our objects that can be set to None).

kyleam: Ah, sorry.  Should have looked more closely.  Since it's mandatory, you can just use the getitem access without worrying about a KeyError then, right?

niceman/distributions/conda.py
rbuccigrossi: OK, this has led me to delve into the conda-env code: https://github.com/conda/conda-env/blob/develop/conda_env/env.py .  Clearly I can learn some lessons from them (making the file exactly as they do).  It does look however that pkg== is legit from their construction, but I'll test it out :)

niceman/distributions/conda.py
rbuccigrossi: It's for me until I get things working :)

niceman/distributions/conda.py
rbuccigrossi: Not yet...  (I'm working inside out...)

niceman/distributions/conda.py
rbuccigrossi: OK, pkg== is an invalid package specification. So I need to be more nuanced in building the package name.  (Time for a helper function.)

niceman/distributions/conda.py
```diff
@@ -94,13 +96,99 @@ def install_packages(self, session=None):
             Environment sub-class instance.
         """
 
-        # TODO: Need to figure out a graceful way to install conda before we can install packages here.
-        # for package in self.provenance['packages']:
-        #     session.add_command(['conda',
-        #                            'install',
-        #                            package['name']])
+        if not self.path:  # Permit dummy tests
+            return
```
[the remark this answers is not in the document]
rbuccigrossi: I claim it is the right action but bad comment.  We should permit empty conda resource entries...

niceman/distributions/conda.py
```diff
@@ -94,13 +96,99 @@ def install_packages(self, session=None):
...
+        try:
+            # Install Conda
+            # See if Conda root path exists and if not, install Conda
+            if not session.isdir(self.path):
+                # TODO: Support OSX and non x86_64 (and move to helper func)
+                miniconda_url = "https://repo.continuum.io/miniconda/"
+                if self.python_version.startswith("2"):
+                    miniconda_url += "Miniconda2-latest-Linux-x86_64.sh"
+                else:
+                    miniconda_url += "Miniconda3-latest-Linux-x86_64.sh"
```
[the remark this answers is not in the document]
rbuccigrossi: Absolutely.  My intention is to move this to an isolated function that creates the URL from the Conda and Python version, and if it is not available then fall back to the latest  :)

niceman/distributions/conda.py
```diff
@@ -94,13 +96,99 @@ def install_packages(self, session=None):
...
-        #     session.add_command(['conda',
-        #                            'install',
-        #                            package['name']])
+        if not self.path:  # Permit dummy tests
+            return
+
+        if not session:
+            session = get_local_session()
+
+        # Make a temporary directory for our installation files
```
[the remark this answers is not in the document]
rbuccigrossi: I created a session.mktmpdir() member

niceman/resource/session.py
```diff
@@ -563,6 +573,12 @@ def mkdir(self, path, parents=False):
         if not self.isdir(path):
             raise CommandError(cmd='mkdir', msg="Failed to create directory")
 
+    def mktmpdir(self):
+        path, err = self.execute_command(["mktemp", "-d"])
+        if err:
+            raise SessionRuntimeError("mktmpdir had error output: %s" % err)
```
kyleam: Do you know of a case where `mktemp -d` has an exit code of 0 and stderr?  Otherwise, I think `execute_command` will throw a `CommandError` before this is reached.

niceman/distributions/conda.py
```diff
@@ -121,15 +152,101 @@ def install_packages(self, session=None):
         ----------
         session : object
             Environment sub-class instance.
+
+        Raises
+        ------
+        ValueError
+            Unexpected conda platform or python version
+        CommandError
+            If unexpected error in install commands occur
```
kyleam: s/occur/occurs/

niceman/distributions/tests/test_conda.py
```diff
@@ -20,108 +20,212 @@
...
-    if sys.platform.startswith('darwin'):
-        miniconda_sh = "Miniconda2-latest-MacOSX-x86_64.sh"
-    elif sys.platform.startswith('linux'):
-        miniconda_sh = "Miniconda2-latest-Linux-x86_64.sh"
-    else:
-        raise ValueError("Conda test not supported with platform %s " %
-                         sys.platform)
-
-    pymod_dir = os.path.join(test_dir, "minimal_pymodule")
-    create_pymodule(pymod_dir)
```
kyleam: [mostly thinking out loud]

Dropping this means that we're not testing that editable packages are detected when we trace conda environments. I'm guessing this is necessary because we're now doing an instantiate then trace test, and we can't instantiate editable packages (at least with conda instantiation alone).

rbuccigrossi: Ah...  but I can squeeze in the creation of the minimal_pymodule after the instantiate (I do something similar in order to test the detection of an empty conda environment).  I'll look into that.

rbuccigrossi: I was able to add the minimal_pymodule test back in.

niceman/resource/session.py
rbuccigrossi: No, I don't.  I did model this after the "read" command two lines above.  But I'm happy to simplify the code (mkdir is also simpler)

niceman/distributions/conda.py
rbuccigrossi: 1 instance replaced.

PR #180, opened 2018-02-14
BF: piputils: Be more selective when splitting show output

kyleam:
When multiple packages are passed to 'pip show', the entries for each package are separated by a line that contains the marker "---".  We split the output string on "---" to get the entries.  That, however, will break if an entry's field contains "---".  Instead use a regular expression to restrict the split to lines that consist only of "---".

Re: https://github.com/ReproNim/niceman/pull/168#discussion_r168196705


yarikoptic:
perfect score ;-)

PR #181, opened 2018-02-14
RF: multipass through Tracers

yarikoptic:
From our pair-coding session with @kyleam earlier today:
- [x] RF code to use sets for files to be tracked, instead of `list->set->list->...` 
- [ ] pass back into the list of files paths which might still be of interest for tracing by other tracers
     - [x] venv - `origin_location` (excluding the path under the venv)
     - [ ] ...conda?
        moved to #189 
- [x] actually support/do multiple passes while enhancing information in previously collected distributions/packages
- ~~~unify multiple distribution entries in the retrace output~~~ moved to #184

### DIFF
--- a/niceman/utils.py
+++ b/niceman/utils.py
@@ -1202,4 +1202,5 @@ def is_subpath(path, directory):
     """
     return not os.path.relpath(path, directory).startswith(os.path.pardir)
 
+
 lgr.log(5, "Done importing niceman.utils")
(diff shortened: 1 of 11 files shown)


niceman/utils.py
```diff
@@ -1193,4 +1194,13 @@ def _walk_up(path):
             yield path
             path = os.path.dirname(path)
 
+
+def is_subpath(path, directory):
+    """Test whether `path` is below (or is itself) `directory`.
+
+    Symbolic links are not resolved before the check.
+    """
+    return not os.path.relpath(path, directory).startswith(os.path.pardir)
```
rbuccigrossi: Out of curiosity, why not:
`return os.path.normpath(directory).startswith(os.path.normpath(path))
`
?

rbuccigrossi: Wow... mine would fail if the norm path doesn't end with slash because

 /a/foo.txt
 /a/foo.txt2

would be considered subdirectories.

PR #183, opened 2018-02-15
BF: vcs: Replace 'lstrip' with 'strip' for older Git version

### LINKED ISSUE #182: test_git_repo_remotes fails
test_git_repo_remotes fails for me, ultimately with:

> E       AttributeError: 'NoneType' object has no attribute 'splitlines'
> 
> niceman/distributions/vcs.py:368: AttributeError

The command in question is:

```
        remote_branches = self._run_git(
            ["for-each-ref", "--contains", hexsha,
             # refs/remotes/<remote>/<name> => <remote>/<name>
             "--format=%(refname:lstrip=2)",
             "refs/remotes"],
           expect_fail=True).splitlines()

```

Why do you (@kyleam ?) unconditionally process the return value of _run_git() with expect_fail=True?  That implies you're _not_ expecting failure. :)

kyleam:
#182 reports a failure in `test_git_repo_remotes`.  The error in that issue is an `AttributeError` that occurs because on failure `expect_fail=True` will result in None being returned, which we try to call `splitlines` on.

We, however, shouldn't actually expect an error there.  We know we're in a Git directory at that point because we have a hash.  Plus, `test_git_repo_remotes` is passing on Travis, so there has to be some difference causing the failure seen in #182.  I _think_ that is the fact that `lstrip` is a relatively recent addition to `git for-each-ref` (v2.13.0).

@chaselgrove, could you verify 1) your local Git version is pre v2.13.0 and 2) that `test_git_repo_remotes` now passes for you locally?

Closes #182.

### DIFF
--- a/niceman/distributions/vcs.py
+++ b/niceman/distributions/vcs.py
@@ -364,8 +364,7 @@ class GitRepoShim(GitSVNRepoShim):
             ["for-each-ref", "--contains", hexsha,
              # refs/remotes/<remote>/<name> => <remote>/<name>
              "--format=%(refname:strip=2)",
-             "refs/remotes"],
-            expect_fail=True).splitlines()
+             "refs/remotes"]).splitlines()
 
         if not remote_branches:
             return {}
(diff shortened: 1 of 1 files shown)


kyleam:
I've verified that if I use Git v2.11.0 I see the error from #182.

chaselgrove:
Yes, I'm using git 2.11.0.

Tests pass for me on kyleam:bf-iss182.

Thanks!

PR #185, opened 2018-02-19
Fix duplicate pip entry in conda trace; add "editable" attribute for conda/pip

kyleam:
This PR fixes two issues:

   * In 54da3757 I introduced a bug in conda traces that caused a duplicate pip listing for each conda package.

   * Starting with 6762a746, I've abused "origin_location" to mean any location, but originally it was only used for locations of editable packages.  I've renamed "origin_location" to "location" and added a field "editable" that is True if the package is editable.

WIP because

  - [X] Needs more tests, particularly for the "editable" attribute.  My hesitation here is that I'm not sure of what the best way is to set up that test.

     Update: I've now added an editable package to the venv tests.

### DIFF
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -14,15 +14,15 @@ from collections import defaultdict
 import attr
 import yaml
 
-from niceman.distributions import Distribution
+from niceman.distributions import Distribution, piputils
+from niceman.dochelpers import exc_str
+from niceman.utils import PathRoot
 
 from .base import SpecObject
 from .base import DistributionTracer
 from .base import Package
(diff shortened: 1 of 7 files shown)


kyleam:
The test failures would be fixed by #187.

kyleam:
I've rebased this to deal with conflicts.  One remaining issue that we discussed in the call is how to handle the test python module.  Currently it is just a helper function (in tests.utils) that is used to in the code that sets up the cached testing environments for venv and conda.  A fixture (at the module or session level) doesn't fit well with the current setup because the venv and conda test directories are cached across tests runs.

PR #186, opened 2018-02-22
BF+ENH: Support latest conda release and find conda environment names

rbuccigrossi:
This does two things:

1. Addresses #187 - The latest release of conda does not put bin/conda in the environments. Therefore this pull request changes how we search for root paths for each environment.  We also remove the assumption that there is a conda and python version for each conda environment.
2. We now determine the name of the conda environment from the path (an issue identified in #174). An alternate approach is to use "conda-env info", but since we know the conda root path, we can save a shell call by analyzing the paths instead.

### DIFF
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -345,10 +345,11 @@ class CondaTracer(DistributionTracer):
                 packages.append(package)
 
             # Give the distribution a name
-            if (len(conda_paths)) > 1:
-                env_name = 'conda_env-%d' % idx
+            # Determine name from path (Alt approach: use conda-env info)
+            if os.path.normpath(conda_path) == os.path.normpath(root_path):
+                env_name = "root"
             else:
-                env_name = 'conda_env'
+                env_name = os.path.basename(os.path.normpath(conda_path))
 
             # Keep track of found package count
             found_package_count += len(packages)
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -246,31 +246,14 @@ class CondaTracer(DistributionTracer):
         return self._env_path_root(path)
 
     def _is_conda_env_directory(self, path):
-        try:
-            _, _ = self._session.execute_command(
-                'ls -ld %s/conda-meta'
-                % path
-            )
-        except Exception as exc:
-            lgr.debug("Did not detect conda env at the path %s: %s", path,
-                      exc_str(exc))
-            return False
-        return True
+        return self._session.exists('%s/conda-meta' % path)
 
     def _get_conda_root_path(self, path):
         return self._root_path_root(path)
 
     def _is_conda_root_directory(self, path):
-        try:
-            _, _ = self._session.execute_command(
-                'ls -ld %s/bin %s/envs %s/conda-meta'
-                % (path, path, path)
-            )
-        except Exception as exc:
-            lgr.debug("Did not detect conda root at the path %s: %s", path,
-                      exc_str(exc))
-            return False
-        return True
+        return all(map(self._session.exists, ('%s/%s' % (path, d) for d in
+                                              ('bin', 'envs', 'conda-meta'))))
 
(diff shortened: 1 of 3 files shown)


niceman/distributions/conda.py
```diff
@@ -345,19 +363,18 @@ def identify_distributions(self, paths):
                 packages.append(package)
 
             # Give the distribution a name
-            if (len(conda_paths)) > 1:
-                env_name = 'conda_env-%d' % idx
+            # Determine name from path (Alt approach: use conda-env info)
+            if os.path.normpath(conda_path) == os.path.normpath(root_path):
+                env_name = "root"
```
kyleam: I suppose that one advantage of the alternative approach is that we wouldn't end up with the "root"/"base" mismatch for newer conda versions.

niceman/distributions/conda.py
```diff
@@ -345,19 +363,18 @@ def identify_distributions(self, paths):
...
             else:
-                env_name = 'conda_env'
+                env_name = os.path.basename(os.path.normpath(conda_path))
```
kyleam: I'm guessing the `normpath` call is to deal with trailing slashes so that "base/" still returns "base"?

rbuccigrossi: Yes, I was afraid that the paths given to us could have lots of weirdness (like ../ or ending slashes), so by normalizing I can make it more resilient.

kyleam: Makes sense, and certainly helps for the trailing slash.  Not that it matters, but I wasn't able to think up a way where `basename` would return a different value because of "../" or any other weirdness in the path leading to the base name.

niceman/distributions/conda.py
```diff
@@ -269,7 +283,7 @@ def identify_distributions(self, paths):
 
         # First, loop through all the files and identify conda paths
         for path in paths:
-            conda_path = self._get_conda_path(path)
+            conda_path = self._get_conda_env_path(path)
```
yarikoptic: so here if we have `paths=['/a/b', '/a/c']` where `/a` is not a conda path... I wondered if we could manage to avoid checking the same parent paths over and over again.

kyleam: > avoid checking the same parent paths over and over again

Beyond the caching that `PathRoot` does?

Not directly related, but I'm not sure I see the advantage of having the `_get_conda_env_path` and `_get_conda_root_path` since they are marked as internal and are just direct calls to the respective `PathRoot` objects.

rbuccigrossi: I have to admit that I found it easier to read.  However, it does hide the fact that PathRoot caches the results.

niceman/distributions/conda.py
```diff
@@ -243,17 +242,32 @@ def _get_conda_info(self, conda_path):
...
+
+    def _get_conda_root_path(self, path):
+        return self._root_path_root(path)
+
+    def _is_conda_root_directory(self, path):
+        try:
+            _, _ = self._session.execute_command(
+                'ls -ld %s/bin %s/envs %s/conda-meta'
+                % (path, path, path)
+            )
```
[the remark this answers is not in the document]
rbuccigrossi: I like using session.exists because it would be faster for local sessions (though we don't have ShellSession.exists() local function defined (I'll do that now).

niceman/distributions/conda.py
```diff
@@ -243,17 +242,32 @@ def _get_conda_info(self, conda_path):
...
 
-    def _is_conda_directory(self, path):
+    def _is_conda_env_directory(self, path):
         try:
             _, _ = self._session.execute_command(
-                'ls -ld %s/bin/conda %s/conda-meta'
-                % (path, path)
+                'ls -ld %s/conda-meta'
+                % path
             )
```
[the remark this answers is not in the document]
rbuccigrossi: I like it, and by defining exists() for the local shell session, this is faster (in that case)

Issue #179, opened 2018-02-14
Fix weak check in skip_if_no_docker_container decorator


mjtravers:
in file tests/utils.py

```
+    def decorator(func):
+        stdout, _ = Runner().run(['docker', 'ps'])
+        if container_name not in stdout:
```
very weak check since could be a partial match. Please TODO @mjtravers - make matching by the value of the second column

Issue #178, opened 2018-02-14
Fix PY3 conflict with interactive shell


mjtravers:
In resource/ssh.py

```
        except Exception as e:
            raise exception.SSHError(str(e))
```
unfortunately I guess due to the size of the diff codecov extension no longer annotates the covered lines, but I think here it it wouldn't work in python3 and would need to analyze the str() of the exception... indeed:

```
$> python3 -c 'from paramiko import SFTPError; ex=SFTPError("buga"); print("buga" in ex)'  
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: argument of type 'SFTPError' is not iterable
3 6967 ->1.....................................:Wed 07 Feb 2018 04:38:05 PM EST:.
hopa:~
$> python3 -c 'from paramiko import SFTPError; ex=SFTPError("buga"); print("buga" in str(ex))'
True
```

mjtravers:
Fixed in PR #264

Issue #177, opened 2018-02-14
SSH connections should validate host and work with user's system settings


mjtravers:
In resources/ssh.py we shouldn't blindly login to unknown hosts.

Also:
didn't look how difficult or possible via paramiko but we need make sure that it works coherently with existing system/user ssh configuration, i.e. that existing known_hosts used by user's ssh command are used, which seems to be not the case in simplest paramiko invocation:

```
$> python -c 'from paramiko import SSHClient; ssh = SSHClient().connect("127.0.0.1", username=None)'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/dist-packages/paramiko/client.py", line 361, in connect
    server_key)
  File "/usr/lib/python2.7/dist-packages/paramiko/client.py", line 648, in missing_host_key
    raise SSHException('Server %r not found in known_hosts' % hostname)
paramiko.ssh_exception.SSHException: Server '127.0.0.1' not found in known_hosts
```

* make sure that ssh keys authentication (possibly provided by ssh-agent) would be used

yarikoptic:
@memartone Please have a look at https://github.com/paramiko/paramiko/blob/master/demos/demo.py#L41 which seems to provide quite a complete example for establishing an interactive session, while also trying to use SSHAgent if available.  It worked for me, besides only not picking up custom `port` setting for the host in my `.ssh/config` (yet to check if that is complexity/mess of my config or something else)

mjtravers:
Addressed in PR #264

PR #188, opened 2018-02-22
BF - Refactored SSH interactive shell and docker check decorator

mjtravers:
This is a small commit addressing the issues: #177, #178, and #179 .

Part of the time on this was spent investigating Elasticluster. Elasticluster provides a command line Python script that starts and stops clusters of machines on cloud providers including Google, AWS, and OpenStack as well as provisioning them from Ansible files and setting up cluster pipeline managers like SLURM and SGE.

We should discuss at a status meeting on if and how we could leverage Elasticluster.

The code commit updates:

- Improve the docker decorator check by looking for container Id rather than name.
- Load the system known_hosts when creating an SSH connection.
- Replace the SSH interactive shell code with a much simplified version inspired by what is done in Elasicluster

### DIFF
--- a/niceman/resource/ssh.py
+++ b/niceman/resource/ssh.py
@@ -11,11 +11,6 @@
 import attr
 import uuid
 from pipes import quote
-import socket
-import termios
-import tty
-import sys
-import select
 import paramiko
 import os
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
we moved away from this approach for now

PR #190, opened 2018-02-22
Merge of yarikoptic/rf-two-pass

kyleam:
I'll make sure tests pass with the conflict resolution and then push to origin/master.


yarikoptic:
you should be able to just push to my branch on my remote for that PR

kyleam:
> you should be able to just push to my branch on my remote for that PR

Sure, but why would I prefer that way?  Wouldn't that involve either merging master into the topic or push master:yarikoptic/rf-two-pass?  I don't see either of those as better.

PR #191, opened 2018-02-28
Cosmetic followups to #185

kyleam:
This PR makes a few cosmetic changes to code from #185 that @yarikoptic and I discussed.  I think the first commit (f553802c) is an obvious improvement, but I find the changes in the other two commits (4208d472, 9c962a6e) to be less readable than what's replaced.

Labeling as "WiP" because I think either the last two commits should be dropped or replaced with other suggestions.


kyleam:
Updated to drop 9c962a6e and to use @yarikoptic's suggestion of group names in the regexp (replacing 4208d472 with e7daa1b7).  Removing the WiP label since I think these two changes are good to merge.

Issue #195, opened 2018-03-01
conda information should include architecture


yarikoptic:
there are x86 (32bit) and x86_64 (64 bit, currently what we assume) miniconda builds available: see https://repo.continuum.io/miniconda/  but our spec seems to be lacking architecture information about conda

kyleam:
Closed by #196.

PR #196, opened 2018-03-05
ENH: Detect conda architecture (platform) #195

rbuccigrossi:
Since we already run conda --info, detecting the platform was surprisingly easy.  This addresses #195 .

### DIFF
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -27,17 +27,6 @@ from .base import TypedList
 import logging
 lgr = logging.getLogger('niceman.distributions.conda')
 
-# Provides the conda platform mapping from comda/models/enum.py
-# Note that these are python prefixes (both 'linux2' and 'linux' from python
-# map to 'linux' in conda.
-python_to_conda_platform_map = {
-    'darwin': 'osx',
-    'linux': 'linux',
-    'openbsd': 'openbsd',
-    'win': 'win',
-    'zos': 'zos',
-}
-
 
 def get_conda_platform_from_python(py_platform):
     """
@@ -45,15 +34,25 @@ def get_conda_platform_from_python(py_platform):
 
     Parameters
     ----------
-    py_platform : basestring
+    py_platform : str
         The python platform string
 
     Returns
     -------
-    basestring
+    str
         The conda platform string
 
     """
+    # Provides the conda platform mapping from conda/models/enum.py
+    # Note that these are python prefixes (both 'linux2' and 'linux' from
+    # python map to 'linux' in conda.)
+    python_to_conda_platform_map = {
+        'darwin': 'osx',
+        'linux': 'linux',
+        'openbsd': 'openbsd',
+        'win': 'win',
+        'zos': 'zos',
+    }
     for k in python_to_conda_platform_map:
         if py_platform.startswith(k):
             return python_to_conda_platform_map[k]
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -68,6 +68,7 @@ class CondaDistribution(Distribution):
     path = attr.ib(default=None)
     conda_version = attr.ib(default=None)
     python_version = attr.ib(default=None)
+    platform = attr.ib(default=None)
     environments = TypedList(CondaEnvironment)
 
     def initiate(self, environment):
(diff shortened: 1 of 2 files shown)


niceman/distributions/tests/test_conda.py
```diff
@@ -81,6 +81,9 @@ def test_conda_manager_identify_distributions(get_conda_test_dir):
         "/sbin/iptables",
         os.path.join(test_dir, "minimal_pymodule")}
 
+    assert distributions.platform in ["linux-64", "osx-64"], \
```
yarikoptic: this would fail whenever we run tests on 32bit, etc whenever we get there to install 32bit conda in 32bit env (time will come).  So why not to make it more specific if e.g. in this whole setup we assume that conda should match our native env, it should match one to one (after mapping) then to output of the local python's `platform.architecture()`

rbuccigrossi: Good point.  I found the python to conda platform mapping that conda does in its source code: https://github.com/conda/conda/blob/301133cb8c804730c218ad8e47f116a8b1af51fc/conda/models/enums.py .  So I created a function to map python platforms to conda platforms and then updated our unit test to make sure that the detected platform is what's expected for the current environment.

niceman/distributions/conda.py
```diff
@@ -27,6 +27,37 @@
 import logging
 lgr = logging.getLogger('niceman.distributions.conda')
 
+# Provides the conda platform mapping from comda/models/enum.py
```
kyleam: s/comda/conda/

niceman/distributions/conda.py
```diff
@@ -27,6 +27,37 @@
...
+# Note that these are python prefixes (both 'linux2' and 'linux' from python
```
kyleam: missing the trailing ")"

niceman/distributions/conda.py
```diff
@@ -27,6 +27,37 @@
...
+}
+
+
+def get_conda_platform_from_python(py_platform):
+    """
+    Converts a python platform string to a corresponding conda platform
+
+    Parameters
+    ----------
+    py_platform : basestring
```
kyleam: s/basestring/str/, since basestring doesn't exist in Python 3.

Or "string", I guess.  Our docstrings are inconsistent:

```
$ git grep " : str\b" | wc -l
14
$ git grep " : string" | wc -l
40
```

"string" is the clear winner (especially since a few of those "str"s are mine), but I prefer "str" because it matches the actual type.  "str" is also consistent with our preference for "bool" over "boolean" and "dict" over "dictionary".

```
$ git grep " : bool\b" | wc -l
15
$ git grep " : boolean" | wc -l
2
$ git grep " : dict\b"  | wc -l
15
$ git grep " : dictionary"  | wc -l
2
```

Another point in favor of "str": The [numpy documentation][0] also uses it.

[0]: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

niceman/distributions/conda.py
```diff
@@ -27,6 +27,37 @@
 import logging
 lgr = logging.getLogger('niceman.distributions.conda')
 
+# Provides the conda platform mapping from comda/models/enum.py
+# Note that these are python prefixes (both 'linux2' and 'linux' from python
+# map to 'linux' in conda.
+python_to_conda_platform_map = {
```
kyleam: Since this is only used in the function below, why not make it local to that function?

niceman/distributions/conda.py
rbuccigrossi: Thank you...

niceman/distributions/conda.py
rbuccigrossi: And again...

niceman/distributions/conda.py
rbuccigrossi: pycharm suggested basestring, but I prefer str myself.

niceman/distributions/conda.py
rbuccigrossi: That was an artifact of the order in which I wrote it.  Originally I just had the dict, but I realized that the function made the dict much more useful as I was writing the test....  I'll combine it.

Issue #194, opened 2018-03-01
conda retracing should be robust to having no custom packages in the environment


yarikoptic:
currently blows:
```shell
$> ~/anaconda-4.4.0-3.6/envs/datalad/bin/conda env list  
# conda environments:
#
datalad                  /home/yoh/anaconda-4.4.0-3.6/envs/datalad
root                  *  /home/yoh/anaconda-4.4.0-3.6

(dev) 2 12825.....................................:Thu 01 Mar 2018 05:49:08 PM EST:.
(git)hopa:~/proj/repronim/niceman[master]git
$> ls -l ~/anaconda-4.4.0-3.6/envs/datalad/conda-meta 
total 4
-rw------- 1 yoh yoh 113 Feb  1 14:30 history
(dev) 2 12826.....................................:Thu 01 Mar 2018 05:49:24 PM EST:.
(git)hopa:~/proj/repronim/niceman[master]git
$> niceman retrace ~/anaconda-4.4.0-3.6/envs/datalad/bin/conda            
2018-03-01 17:49:30,480 [INFO   ] Entering iteration #1 over Tracers 
2018-03-01 17:49:31,060 [INFO   ] DebTracer: 0 envs with 1 other files remaining 
2018-03-01 17:49:31,887 [WARNING] Could not retrieve conda-meta files in path /home/yoh/anaconda-4.4.0-3.6/envs/datalad: CommandError: command 'ls /home/yoh/anaconda-4.4.0-3.6/envs/datalad/conda-meta/*.json' failed with exitcode 2.
Failed to run u'ls /home/yoh/anaconda-4.4.0-3.6/envs/datalad/conda-meta/*.json' under None. Exit code=2. out= err=ls: cannot access '/home/yoh/anaconda-4.4.0-3.6/envs/datalad/conda-meta/*.json': No such file or directory
 [cmd.py:run:297] 
2018-03-01 17:49:31,888 [ERROR  ] 'NoneType' object is not iterable [conda.py:_get_conda_package_details:131] (TypeError) 
```

kyleam:
Fixed by #197.

PR #197, opened 2018-03-05
BF: Gracefully handle no custom packages in a conda environment #194

rbuccigrossi:
This detects and gracefully handles when there are no custom packages installed in a conda environment (so that our search for *.json is empty).  This addresses #194 .

### DIFF
--- a/niceman/distributions/conda.py
+++ b/niceman/distributions/conda.py
@@ -16,6 +16,7 @@ import yaml
 
 from niceman.distributions import Distribution, piputils
 from niceman.dochelpers import exc_str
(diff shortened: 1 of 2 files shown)


niceman/distributions/conda.py
```diff
@@ -16,6 +16,7 @@
 
 from niceman.distributions import Distribution, piputils
 from niceman.dochelpers import exc_str
+from niceman.support.exceptions import CommandError
```
kyleam: With eabf08c, this import is unused.

rbuccigrossi: Good eye!

PR #198, opened 2018-03-13
ENH: Added a distributions module for Docker to allow tracing and ins…

mjtravers:
…talling of Docker images to an environment running the Docker engine.

I experimented with Docker Hub and discovered that images pushed to the hub are still accessible after all the tags assigned to them are deleted. At least they have been available for the past few days since I have removed the tags from a test image.

However, if a repo is deleted, then the saved images are lost.

For this branch, the retrace and install operate at the image level. I will additionally look into retracing within a Docker container itself as discussed at the last status meeting.

### DIFF
--- a/niceman/distributions/docker.py
+++ b/niceman/distributions/docker.py
@@ -1,0 +1,169 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the niceman package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Support for Docker distribution(s)."""
+
+import attr
+import json
+import logging
+
+lgr = logging.getLogger('niceman.distributions.docker')
+
+from .base import Package
+from .base import Distribution
+from .base import DistributionTracer
+from .base import TypedList
+from .base import _register_with_representer
+from ..dochelpers import borrowdoc
+from ..support.exceptions import CommandError
+
+
+@attr.s(slots=True, frozen=True, cmp=False, hash=True)
+class DockerImage(Package):
+    """Docker image information"""
+    id = attr.ib()
+    # Optional
+    architecture = attr.ib(default=None)
+    operating_system = attr.ib(default=None)
+    docker_version = attr.ib(default=None)
+    repo_digests = attr.ib(default=None)
+    repo_tags = attr.ib(default=None)
+    created = attr.ib(default=None)
+
+_register_with_representer(DockerImage)
+
+@attr.s
+class DockerDistribution(Distribution):
+    """
+    Class to provide commands to Docker engine.
+    """
+
+    images = TypedList(DockerImage)
+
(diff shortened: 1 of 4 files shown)


niceman/distributions/docker.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Parameters
+        ----------
+        session : object
+            The Session to work in
+        """
+
+        # Raise niceman.support.exceptions.CommandError exception if Docker 
+        # engine is not to be found.
+        session.execute_command(['docker', 'info'])
```
yarikoptic: is this needed to initiate smth or just a mere "check" that docker is there and accessible?

niceman/distributions/docker.py
```diff
@@ -0,0 +1,169 @@
...
+class DockerImage(Package):
+    """Docker image information"""
+    id = attr.ib()
+    # Optional
+    architecture = attr.ib(default=None)
+    operating_system = attr.ib(default=None)
+    docker_version = attr.ib(default=None)
+    repo_digests = attr.ib(default=None)
+    repo_tags = attr.ib(default=None)
+    created = attr.ib(default=None)
```
yarikoptic: we will need an information about the hub from which it came (if possible)

niceman/distributions/docker.py
```diff
@@ -0,0 +1,169 @@
...
+        ----------
+        session : object
+            Session to work in
+        """
+
+        for image in self.images:
+
+            # First, look in local Docker engine.
+            try:
+                session.execute_command(['docker', 'image', 'inspect',
```
kyleam: `session` is an optional argument.  If nothing is passed, wouldn't this result in an AttributeError?

niceman/distributions/docker.py
```diff
@@ -0,0 +1,169 @@
...
+                    msg="Unable to locate Docker image {}".format(image.id))
+
+_register_with_representer(DockerDistribution)
+
+
+class DockerTracer(DistributionTracer):
+    """Docker image tracer
+
+    If a Docker engine is not found running in the session, the files
+    are quietly passed on to thep next tracer.
```
kyleam: s/thep/the/

niceman/distributions/docker.py
```diff
@@ -0,0 +1,169 @@
...
+
+    HANDLES_DIRS = False
+
+    @borrowdoc(DistributionTracer)
+    def identify_distributions(self, files):
+        if not files:
+            return
+
+        images = []
+        remaining_files = []
```
kyleam: You end up converting this to a set at the end, so why not start off with a set?

niceman/distributions/docker.py
mjtravers: This is just a check to make sure Docker is accessible. No other function. It will raise a CommandError exception if Docker can not be found.

niceman/distributions/docker.py
mjtravers: The hub (or in Docker speak, the registry and repository) where the image can be found is contained in the repo_digests property. For example, here is the digest for a local docker registry I am running on my local machine:

localhost:5000/my-ubuntu@sha256:615701b8ab7d454db039f412e50f22daa2af04b1561e08672e758cdac2af22cb

If there is no HTTP prefix to the digest, it is (by convention) defaulted to hub.docker.com

The hash after the @ symbol is a hash on the image manifest file located in the registry. The manifest contains, among other things, the name and tag of the image and the hashes of the file system layers needed to reconstruct the image.

mjtravers:
Ready for merging.

PR #201, opened 2018-04-03
ENH: vcs(git): Use symbolic-ref to get current branch

kyleam:
For the most part, symbolic-ref's output will be the same as that of
rev-parse --abbrev-ref, but it's a bit nicer because 1) it will return
the current branch when we're on a branch without any commits and 2)
it won't return "HEAD" when we're in a detached state.


yarikoptic:
one day I will know them all by heart! ;)

PR #204, opened 2018-04-10
ENH - Added resource for singularity containers.

mjtravers:
This commit adds a resource class for singularity containers. (..and sneaks in a fix in the testing of Docker retracing)

The retracing of singularity containers is on hold while we investigate how to map containers back to their original image in a repository.

The installation of singularity was added to the Travis test setup.

- [ ] remove workaround for too long names

### DIFF
--- a/niceman/distributions/singularity.py
+++ b/niceman/distributions/singularity.py
@@ -1,0 +1,136 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the niceman package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Support for Singularity distribution(s)."""
+
+import attr
+import json
+import logging
+
+lgr = logging.getLogger('niceman.distributions.singularity')
+
+from .base import Package
+from .base import Distribution
+from .base import DistributionTracer
+from .base import TypedList
+from .base import _register_with_representer
+from ..dochelpers import borrowdoc
+from ..utils import attrib, md5sum
+
+
+@attr.s(slots=True, frozen=True, cmp=False, hash=True)
+class SingularityImage(Package):
+    """Singularity image information"""
+    md5 = attrib(default=attr.NOTHING)
+    # Optional
+    bootstrap = attrib()
+    maintainer = attrib()
+    deffile = attrib()
+    schema_version = attrib()
+    build_date = attrib()
+    build_size = attrib()
+    singularity_version = attrib()
+    base_image = attrib()
+
+_register_with_representer(SingularityImage)
+
+
+@attr.s
+class SingularityDistribution(Distribution):
+    """
+    Class to provide commands to Singularity.
+    """
+
+    images = TypedList(SingularityImage)
+
+    def initiate(self, session):
+        """
+        Perform any initialization commands needed in the environment
+        environment.
+
+        Parameters
+        ----------
+        session : object
+            The Session to work in
+        """
+
+        # Raise niceman.support.exceptions.CommandError exception if
+        # Singularity is not to be found.
+        session.execute_command(['singularity', 'selftest'])
+
+    def install_packages(self, session):
+        """
+        Install the Singularity images associated to this distribution by the
+        provenance into the environment.
+
+        Parameters
+        ----------
+        session : object
+            Session to work in
+        """
+
+        # TODO: Currently we have no way to locate the image given the metadata
+
+_register_with_representer(SingularityDistribution)
+
+
+class SingularityTracer(DistributionTracer):
+    """Singularity image tracer
+
+    If a given file is not identified as a singularity image, the files
+    are quietly passed on to the next tracer.
+    """
+
+    HANDLES_DIRS = False
+
+    @borrowdoc(DistributionTracer)
+    def identify_distributions(self, files):
+        if not files:
+            return
+
+        images = []
+        remaining_files = set()
+
+        for file in files:
+            try:
+                # import pdb; pdb.set_trace()
+                image = json.loads(self._session.execute_command(['singularity',
+                    'inspect', file])[0])
+
+                images.append(SingularityImage(
+                    md5=md5sum(file),
+                    bootstrap=image['org.label-schema.usage.singularity.deffile.bootstrap'],
+                    maintainer=image['MAINTAINER'],
+                    deffile=image['org.label-schema.usage.singularity.deffile'],
+                    schema_version=image['org.label-schema.schema-version'],
+                    build_date=image['org.label-schema.build-date'],
+                    build_size=image['org.label-schema.build-size'],
+                    singularity_version=image['org.label-schema.usage.singularity.version'],
+                    base_image=image['org.label-schema.usage.singularity.deffile.from']
+                ))
+            except Exception as exc:
+                lgr.debug(exc)
+                remaining_files.add(file)
+
+        if not images:
+            return
+
+        dist = SingularityDistribution(
+            name="singularity",
+            images=images
+        )
+
+        yield dist, remaining_files
+
+    @borrowdoc(DistributionTracer)
+    def _get_packagefields_for_files(self, files):
+        return
+
+    @borrowdoc(DistributionTracer)
+    def _create_package(self, **package_fields):
+        return
--- a/niceman/distributions/singularity.py
+++ b/niceman/distributions/singularity.py
@@ -19,7 +19,7 @@ from .base import Distribution
 from .base import DistributionTracer
 from .base import TypedList
 from .base import _register_with_representer
-from ..dochelpers import borrowdoc
+from ..dochelpers import borrowdoc, exc_str
 from ..utils import attrib, md5sum
 
 
@@ -36,6 +36,7 @@ class SingularityImage(Package):
     build_size = attrib()
     singularity_version = attrib()
     base_image = attrib()
+    mirror_url = attrib()
 
 _register_with_representer(SingularityImage)
 
@@ -50,8 +51,7 @@ class SingularityDistribution(Distribution):
 
     def initiate(self, session):
         """
-        Perform any initialization commands needed in the environment
-        environment.
+        Perform any initialization commands needed in the environment.
 
         Parameters
         ----------
@@ -96,26 +96,33 @@ class SingularityTracer(DistributionTracer):
         images = []
         remaining_files = set()
 
-        for file in files:
+        for file_path in files:
             try:
-                # import pdb; pdb.set_trace()
                 image = json.loads(self._session.execute_command(['singularity',
-                    'inspect', file])[0])
+                    'inspect', file_path])[0])
 
                 images.append(SingularityImage(
-                    md5=md5sum(file),
-                    bootstrap=image['org.label-schema.usage.singularity.deffile.bootstrap'],
-                    maintainer=image['MAINTAINER'],
-                    deffile=image['org.label-schema.usage.singularity.deffile'],
-                    schema_version=image['org.label-schema.schema-version'],
-                    build_date=image['org.label-schema.build-date'],
-                    build_size=image['org.label-schema.build-size'],
-                    singularity_version=image['org.label-schema.usage.singularity.version'],
-                    base_image=image['org.label-schema.usage.singularity.deffile.from']
+                    md5=md5sum(file_path),
+                    bootstrap=image.get(
+                        'org.label-schema.usage.singularity.deffile.bootstrap'),
+                    maintainer=image.get('MAINTAINER'),
+                    deffile=image.get(
+                        'org.label-schema.usage.singularity.deffile'),
+                    schema_version=image.get('org.label-schema.schema-version'),
+                    build_date=image.get('org.label-schema.build-date'),
+                    build_size=image.get('org.label-schema.build-size'),
+                    singularity_version=image.get(
+                        'org.label-schema.usage.singularity.version'),
+                    base_image=image.get(
+                        'org.label-schema.usage.singularity.deffile.from'),
+                    mirror_url=image.get(
+                        'org.label-schema.usage.singularity.deffile.mirrorurl')
                 ))
             except Exception as exc:
+                lgr.debug("Probably %s is not a Singularity image: %s",
+                    file_path, exc_str(exc))
                 lgr.debug(exc)
-                remaining_files.add(file)
+                remaining_files.add(file_path)
 
         if not images:
             return
--- a/niceman/resource/singularity.py
+++ b/niceman/resource/singularity.py
@@ -81,8 +81,10 @@ class Singularity(Resource):
             lgr.info('Resource {} already exists.'.format(self.name))
         else:
             # Start the container instance.
+            # NOTE: Logging stdout and stderr hangs the run call, so we
+            # disable the logging in the run call below.
             self._runner.run(['singularity', 'instance.start', self.image,
-                self.name])
+                self.name], log_stdout=False, log_stderr=False)
 
         # Update status
         self.id = self.name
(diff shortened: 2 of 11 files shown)


niceman/resource/singularity.py
```diff
@@ -0,0 +1,240 @@
...
+            lgr.info('Resource {} already exists.'.format(self.name))
+        else:
+            self.image_file = "{}.simg".format(str(uuid.uuid4().hex)[:16])
+            if self.image.startswith('shub://') or \
+                    self.image.startswith('docker://'):
+
+                self._runner.run(['singularity', 'pull', '--name',
+                    self.image_file, self.image])
+            else:
+                shutil.copyfile(self.image, self.image_file)
```
kyleam: Hmm, so the code above is to work around the instance.start's "Label string too long" error?  If so, should we avoid copying self.image if its name is short enough?

niceman/resource/singularity.py
```diff
@@ -0,0 +1,240 @@
...
+        Stop a running container.
+        """
+        lgr.info("Singularity does not provide a stop feature.")
+
+    @borrowdoc(Resource)
+    def get_session(self, pty=False, shared=None):
+        if not self.get_instance_info():
+            self.connect()
+
+        if pty and shared is not None and not shared:
```
kyleam: I'm confused about the values that `shared` can take.  `Resource.get_session` makes it seem like it can be None or a string.  Based on that, I'd guess that the intent of `shared is not None and not shared` is to check that it's not an empty string.  But grepping the code base, I see that `shared=False` is passed, so perhaps the meaning of False should be documented and then this could be just `shared is
False`?

niceman/resource/singularity.py
```diff
@@ -0,0 +1,240 @@
...
+        try:
+            stdout, _ = self._runner.run(['singularity', 'instance.list'])
+        except CommandError:
+            return None
+
+        # Parse stdout to find the running instance.
+        # The output of instance.list is a table with columns:
+        #   DAEMON NAME, PID, and CONTAINER IMAGE
+        # Daemon names are unique on each server.
+        rows = stdout.split('\n')
```
kyleam: nitpick: splitlines (which will take care of the need to handle a trailing blank line below)

niceman/resource/singularity.py
```diff
@@ -0,0 +1,240 @@
...
+        for row in rows:
+            items = row.split()
```
kyleam: [thinking aloud]

This assumes that no fields have spaces.  singularity instance.start doesn't accept spaces for `<instance name>` and even though it accepts a `<container path>` argument with spaces, it doesn't seem to actually be able to handle it.  So this assumption seems safe.  (Plus instance.list doesn't seem to have an option for using something like a nul-separate or json output, so I don't think there is really much we can do.)

niceman/resource/singularity.py
```diff
@@ -0,0 +1,240 @@
...
+            if len(items) != 3:
+                # Skip header and trailing blank lines.
```
kyleam: With the s/split('\n')/splitlines/ change above, you could remove this check and just start the iteration at the second item.

niceman/resource/singularity.py
```diff
@@ -0,0 +1,240 @@
...
+        stdout, stderr = self._runner.run(['singularity', 'exec',
+            'instance://' + self.name] + command)
+
+        return (stdout, stderr)
+
+    @borrowdoc(Session)
+    def put(self, src_path, dest_path, uid=-1, gid=-1):
+        dest_dir, _ = os.path.split(dest_path)
+        self.mkdir(dest_dir, parents=True)
+        cmd = 'cat {} | singularity exec instance://{} tee {} > /dev/null'
```
kyleam: Isn't `cat IN | tee OUT >/dev/null` the same as `cat IN | cat >OUT`?

niceman/resource/singularity.py
mjtravers: I thought about doing that but opted for keeping the code and process as simple as possible by treating all image names the same. It is assumed that this work around will be removed once the bug is fixed in the singularity code base.

niceman/resource/singularity.py
mjtravers: I believe the shared=False in resource.session.get_local_session() is an error. The parameter is defined as thus in base.Resource.get_session:

shared : string, optional
            Shared session identifier (the default is None)

niceman/resource/singularity.py
mjtravers: Yes, I am assuming no spaces in the column data.

niceman/resource/singularity.py
mjtravers: Yes, I can use slicing [1:]. Much more pythony!

niceman/resource/singularity.py
mjtravers: It is. Will change

niceman/resource/singularity.py
mjtravers: Making change

niceman/resource/singularity.py
kyleam: So are you guarding against an empty string?  If not, the `and not shared` can be dropped.

mjtravers: Right, an empty string would not be a valid identifier. So I think that is what the *and not shared* is there to fail.

niceman/resource/singularity.py
kyleam: eh, yeah, you'd have to do something ugly like `/bin/sh -c 'cat ... >OUT'`.  Sorry for the noise.

mjtravers: No worries at all! As you can see from my having to re-edit my response that I had forgotten why I had done the tee thing in the first place.  :)

yarikoptic:
Btw singularity issue with long names is about to be resolved upstream https://github.com/singularityware/singularity/pull/1461 and already included in NeuroDebian (don't remember if I uploaded already)

mjtravers:
This PR addresses Issue #207

yarikoptic:
for retrace we would want for now just the date from `inspect` output + md5sum since that is the one used by `singularity pull --hash`

niceman/distributions/singularity.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Class to provide commands to Singularity.
+    """
+
+    images = TypedList(SingularityImage)
+
+    def initiate(self, session):
+        """
+        Perform any initialization commands needed in the environment
+        environment.
```
kyleam: duplicate "environment"

niceman/distributions/singularity.py
```diff
@@ -0,0 +1,136 @@
...
+    def identify_distributions(self, files):
+        if not files:
+            return
+
+        images = []
+        remaining_files = set()
+
+        for file in files:
+            try:
+                # import pdb; pdb.set_trace()
```
kyleam: nit: Please remove leftover debugging bits.

niceman/distributions/singularity.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @borrowdoc(DistributionTracer)
+    def identify_distributions(self, files):
+        if not files:
+            return
+
+        images = []
+        remaining_files = set()
+
+        for file in files:
```
kyleam: `file` is overriding a builtin.   Perhaps`file_` or `fname`?

niceman/distributions/singularity.py
```diff
@@ -0,0 +1,136 @@
...
+                    bootstrap=image['org.label-schema.usage.singularity.deffile.bootstrap'],
+                    maintainer=image['MAINTAINER'],
+                    deffile=image['org.label-schema.usage.singularity.deffile'],
+                    schema_version=image['org.label-schema.schema-version'],
+                    build_date=image['org.label-schema.build-date'],
+                    build_size=image['org.label-schema.build-size'],
+                    singularity_version=image['org.label-schema.usage.singularity.version'],
+                    base_image=image['org.label-schema.usage.singularity.deffile.from']
+                ))
+            except Exception as exc:
```
kyleam: Can you catch `CommandError` here to be more specific? (I know other tracers catch `Exception`, but I think those should eventually be switched over as well.)

niceman/distributions/singularity.py
```diff
@@ -0,0 +1,136 @@
...
+        for file in files:
+            try:
+                # import pdb; pdb.set_trace()
+                image = json.loads(self._session.execute_command(['singularity',
+                    'inspect', file])[0])
+
+                images.append(SingularityImage(
+                    md5=md5sum(file),
+                    bootstrap=image['org.label-schema.usage.singularity.deffile.bootstrap'],
+                    maintainer=image['MAINTAINER'],
```
kyleam: There may not be a 'MAINTAINER' key.  For example, with

```
BootStrap: shub
From: GodloveD/busybox

%runscript
    echo "This is what happens when you run the container..."

%post
echo "Hello from inside the container"
```

```bash
%> singularity inspect s.simg
{
    "org.label-schema.usage.singularity.deffile.bootstrap": "shub",
    "org.label-schema.usage.singularity.deffile": "Singularity",
    "org.label-schema.schema-version": "1.0",
    "org.label-schema.usage.singularity.deffile.from": "GodloveD/busybox",
    "org.label-schema.build-date": "Fri,_06_Jul_2018_18:19:38_-0400",
    "org.label-schema.usage.singularity.deffile.mirrorurl": "https://www.busybox.net/downloads/binaries/1.26.1-defconfig-multiarch/busybox-x86_64",
    "org.label-schema.usage.singularity.version": "2.5.1-dist",
    "org.label-schema.build-size": "699MB"
}
```

Tracing this image will result in a key error.  On INFO logging, it silently fails because of the "Exception instead of CommandError" issue mentioned above.

Also, it looks like this one has a "deffile.mirrorurl" field.  That might be worth recording if it's present.

niceman/distributions/singularity.py
```diff
@@ -0,0 +1,136 @@
...
+                    deffile=image['org.label-schema.usage.singularity.deffile'],
+                    schema_version=image['org.label-schema.schema-version'],
+                    build_date=image['org.label-schema.build-date'],
+                    build_size=image['org.label-schema.build-size'],
+                    singularity_version=image['org.label-schema.usage.singularity.version'],
+                    base_image=image['org.label-schema.usage.singularity.deffile.from']
+                ))
+            except Exception as exc:
+                lgr.debug(exc)
```
kyleam: `exc_str(exc)` is preferred, I think

niceman/distributions/singularity.py
kyleam: Hmm, I think my suggestion is misguided.  While catching a narrower exception would be better, in the case of the runner, it may raise a non-CommandError.  The case I'm aware of is `FileNotFoundError` when `shell=False` and the name in the command from the first argument isn't found, but perhaps there are other cases as well.

niceman/resource/tests/test_singularity.py
```diff
@@ -0,0 +1,73 @@
...
+import logging
+import re
+from ..singularity import Singularity, SingularitySession
+from ...tests.utils import skip_if_no_singularity, skip_if_no_network, \
+    swallow_logs, assert_in
+
+
+@skip_if_no_network
+@skip_if_no_singularity
+def test_singularity_resource_class():
```
kyleam: This test should use a temporary directory so that it doesn't leave the singularity image in the current directory.

yarikoptic:
@kyleam test failure seems to relate more to the recently merged refactoring than to changes here, or what am I missing?
```

=================================== FAILURES ===================================
_______________________ test_assert_Xwd_unchanged[chdir] _______________________
func = <built-in function chdir>
    @pytest.mark.parametrize("func", [os.chdir, chpwd],
                             ids=["chdir", "chpwd"])
    def test_assert_Xwd_unchanged(func):
        orig_cwd = os.getcwd()
        orig_pwd = getpwd()
    
        @assert_cwd_unchanged
        def do_chdir():
            func(os.pardir)
    
        with pytest.raises(AssertionError) as cm:
            do_chdir()
    
        eq_(orig_cwd, os.getcwd(),
>           "assert_cwd_unchanged didn't return us back to cwd %s" % orig_cwd)
niceman/tests/test_tests_utils.py:253: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
a = '/tmp/tmp8BjI82', b = '/home/travis/build/ReproNim/niceman'
msg = "assert_cwd_unchanged didn't return us back to cwd /tmp/tmp8BjI82"
    def assert_equal(a, b, msg=None):
>       assert a == b, msg or "{!r} != {!r}".format(a, b)
E       AssertionError: assert_cwd_unchanged didn't return us back to cwd /tmp/tmp8BjI82
niceman/tests/utils.py:46: AssertionError
----------------------------- Captured stderr call -----------------------------
2018-07-09 18:27:51,117 [WARNING] <function do_chdir at 0x7f4155351e60> changed cwd to /tmp. Mitigating and changing back to /home/travis/build/ReproNim/niceman 
============== 1 failed, 198 passed, 29 skipped in 205.41 seconds ==============
```

yarikoptic:
@mjtravers can't we install singularity on travis so those tests run?  I think we have singularity-container package for trusty on travis: http://neuro.debian.net/pkgs/singularity-container.html

kyleam:
> @kyleam test failure seems to relate more to the recently merged refactoring than to changes here, or what am I missing?

Hmm, yeah, that's confusing given that tests on master pass.  I'll have a look.

> @mjtravers can't we install singularity on travis so those tests run? I think we have singularity-container package for trusty on travis: http://neuro.debian.net/pkgs/singularity-container.html

Yep, [here's][0] a simple .travis.yml that uses it.

[0]: https://github.com/kyleam/mlb-rundiff/blob/2ef9d1b379446bedd9fb42b747fd18d670a3601f/.travis.yml#L9-L14

mjtravers:
@kyleam @yarikoptic master is failing for me when running the tests. I am going to see if I can get this figured out this morning.

    def test_skip_if_no_network():
        cleaned_env = os.environ.copy()
        cleaned_env.pop('NICEMAN_TESTS_NONETWORK', None)
        # we need to run under cleaned env to make sure we actually test in both conditions
        with patch('os.environ', cleaned_env):
            @skip_if_no_network
            def somefunc(a1):
                return a1
            eq_(somefunc.tags, ['network'])
            with patch.dict('os.environ', {'NICEMAN_TESTS_NONETWORK': '1'}):
>               assert_raises(pytest.skip.Exception, somefunc, 1)

../niceman/niceman/tests/test_tests_utils.py:433: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../niceman/niceman/tests/utils.py:364: in newfunc
    check_and_raise()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def check_and_raise():
        if os.environ.get('NICEMAN_TESTS_NONETWORK'):
            pytest.skip("Skipping since no network settings",
>                       allow_module_level=True)
E           TypeError: skip() got an unexpected keyword argument 'allow_module_level'

../niceman/niceman/tests/utils.py:359: TypeError

kyleam:
> E TypeError: skip() got an unexpected keyword argument 'allow_module_level'

Upgrading your pytest 3.3.0 to should fix this one.

kyleam:
I'm haven't been able to trigger the `test_assert_Xwd_unchanged[chdir]` error locally and still don't have any idea why it would fail in this branch but not on master.

niceman/resource/tests/test_singularity.py
```diff
@@ -0,0 +1,81 @@
...
+
+    with swallow_logs(new_level=logging.DEBUG) as log:
+
+        # Set working directory to a scratch directory since we will be creating
+        # Singularity image files during testing.
+        tempdir = tempfile.mkdtemp()
+        os.chdir(tempdir)
```
yarikoptic: Use `with chpwd(tempdir)` context manager here to make sure you come back "correctly".  This is what caused the test failure of `test_assert_Xwd_unchanged`

PR #218, opened 2018-04-29
DOC:  Improve documentation and reporting of config locations

kyleam:
This PR

  * Mentions the allowed config locations in `MissingConfigFileError`s message so that the user knows how to fix the problem.

  * Mentions `$PWD/.niceman/config` in the locations list.

  * Corrects the documented location of the config file in `XDG_CONFIG_HOME`.

  * Changes the `XDG_CONFIG_HOME` location from `$XDG_CONFIG_HOME/niceman/niceman.cfg` to `$XDG_CONFIG_HOME/niceman/config` for consistency with other names.

### DIFF
--- a/niceman/config.py
+++ b/niceman/config.py
@@ -24,7 +24,7 @@ LOCATIONS_DOC = """
     1. '/etc/niceman/niceman.cfg'
     2. 'niceman/config' in all directories defined by $XDG_CONFIG_DIRS
        (by default: /etc/xdg/)
-    3. 'niceman/niceman.cfg' in $XDG_CONFIG_HOME (by default: ~/.config/)
+    3. 'niceman/config' in $XDG_CONFIG_HOME (by default: ~/.config/)
     4. '.niceman/config', relative to the current directory
     5. 'niceman.cfg' in the current directory""".lstrip()
 
--- a/niceman/config.py
+++ b/niceman/config.py
@@ -24,7 +24,7 @@ locations_doc = """
     1. '/etc/niceman/niceman.cfg'
     2. 'niceman/config' in all directories defined by $XDG_CONFIG_DIRS
        (by default: /etc/xdg/)
-    3. 'niceman.cfg' in $XDG_CONFIG_HOME (by default: ~/.config/)
+    3. 'niceman/niceman.cfg' in $XDG_CONFIG_HOME (by default: ~/.config/)
     4. '.niceman/config', relative to the current directory
     5. 'niceman.cfg' in the current directory""".lstrip()
 
--- a/niceman/config.py
+++ b/niceman/config.py
@@ -20,7 +20,7 @@ from os.path import join as opj
 
 from .support.configparserinc import SafeConfigParserWithIncludes
 
-locations_doc = """
+LOCATIONS_DOC = """
     1. '/etc/niceman/niceman.cfg'
     2. 'niceman/config' in all directories defined by $XDG_CONFIG_DIRS
        (by default: /etc/xdg/)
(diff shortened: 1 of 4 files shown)


kyleam:
Hrm, the `__doc__` assignment in 1e4c207 breaks in py2.  I wasn't aware that there was a py2/py3 difference here.

niceman/config.py
```diff
@@ -31,14 +40,8 @@ class ConfigManager(SafeConfigParserWithIncludes, object):
...
-    called later on.  Files are read and parsed in the following
-    order:
-
-    1. '/etc/niceman/niceman.cfg'
-    2. 'niceman/config' in all directories defined by $XDG_CONFIG_DIRS
-       (by default: /etc/xdg/)
-    3. 'niceman.cfg' in $XDG_CONFIG_HOME (by default: ~/.config/)
-    4. 'niceman.cfg' in the current directory
+    called later on.  Files are read and parsed in the order described by
+    `LOCATIONS_DOC`.
```
yarikoptic: well, you could embed the value here.  We do similar in https://github.com/datalad/datalad/blob/master/datalad/config.py#L29

niceman/config.py
```diff
@@ -126,7 +129,7 @@ def _get_default_file_candidates(self):
         # XDG user config
         home_cfg_base_path = self.dirs.user_config_dir
         if os.path.isabs(home_cfg_base_path):
-            cfg_file_candidates.append(opj(home_cfg_base_path, 'niceman.cfg'))
+            cfg_file_candidates.append(opj(home_cfg_base_path, 'config'))
```
yarikoptic: while at it -- may be it should be unified? i.e. as it was in the code here and just docs adjusted?

niceman/config.py
kyleam: I don't think that's true.  Did you note my attempt to do that in 1e4c207 and then the followup 6c89d18?  I think the core issue is that, in py2, the `__doc__` of a class isn't writeable (AFAICS datalad's _where_reload_doc is only decorating functions).

Am I missing something?

niceman/config.py
kyleam: Sorry, I don't understand your comment.

How could it be unified by just adjusting the docs?  If we're going to unify to niceman.cfg, wouldn't $XDG_CONFIG_DIRS/niceman/config and $PWD/.niceman/config need to be changed?

Also, this change is a separate commit, so feel free to just revert 37cdc86 if you want to stay with $XDG_CONFIG_HOME/niceman/niceman.cfg.

niceman/config.py
yarikoptic: Jesus... forgot about that:  I was certain that it should work since in PyMVPA we do compile `__doc__` for "classes" on the fly.  But looking at the code I realized that we are either doing it "per instance" or within metaclass (apparently could do it there, for the record - `AttributesCollector`).  So - my bad

niceman/config.py
yarikoptic: yes, sorry -- missed those as well.  But what do you think -  worth having them consisent?

niceman/config.py
kyleam: > Jesus... forgot about that

I was totally unaware of the issue until I saw the py2 failures in this PR :]

niceman/config.py
kyleam: Hmm, yeah, I'm fine either way.  NAME/config looks right to me because of things like ".config/git/config", but I see the point of "let's just use one name".

My only strong preference is that we use the same pattern for $XDG_CONFIG_HOME, $XDG_CONFIG_DIRS, and $PWD/.niceman.

kyleam: OK, well since you seem to prefer niceman.cfg and I have no preference and others haven't weighed in, I've changed all the locations to use niceman.cfg.

kyleam:
The failing test is unrelated and now fixed in master.

Issue #212, opened 2018-04-13
docker retracer shouldn't crash tracing if there is no docker


yarikoptic:
```shell
(niceman-dev) [yhalchen@discovery niceman]$ niceman retrace /usr/bin/python2
2018-04-13 13:18:56,494 [INFO   ] Entering iteration #1 over Tracers 
2018-04-13 13:18:56,501 [INFO   ] DebTracer: 0 envs with 1 other files remaining 
2018-04-13 13:18:56,527 [INFO   ] CondaTracer: 0 envs with 1 other files remaining 
2018-04-13 13:18:56,548 [INFO   ] VenvTracer: 0 envs with 1 other files remaining 
2018-04-13 13:18:56,573 [INFO   ] VCSTracer: 0 envs with 1 other files remaining 
2018-04-13 13:18:56,586 [ERROR  ] Failed to start ['docker', 'image', 'inspect', u'/usr/bin/python2']'': [Errno 2] No such file or directory [subprocess.py:_execute_child:1025] 
2018-04-13 13:18:56,588 [ERROR  ] [Errno 2] No such file or directory [subprocess.py:_execute_child:1025] (OSError) 
```
I guess there should first be a test if tracer is "available" for an environment with a single warning about which aren't, and then they would be excluded from a list of tracers.  Assigning to both @kyleam and @mjtravers for a healthy collaboration/competition ;)

yarikoptic:
and then there should be a general handling of unexpected behavior of a tracer - e.g. if it thought it would work but crashes - issue a warning and proceed to the next one, without causing the entire process to collapse.  I've made a dummy `docker` script which just echos empty `{}`, which is legit JSON, so it now crashes with
```shell
(niceman-dev) [yhalchen@discovery niceman]$ niceman --dbg retrace /usr/bin/python2
2018-04-13 13:22:41,978 [INFO   ] Entering iteration #1 over Tracers 
2018-04-13 13:22:41,986 [INFO   ] DebTracer: 0 envs with 1 other files remaining 
2018-04-13 13:22:42,010 [INFO   ] CondaTracer: 0 envs with 1 other files remaining 
2018-04-13 13:22:42,032 [INFO   ] VenvTracer: 0 envs with 1 other files remaining 
2018-04-13 13:22:42,056 [INFO   ] VCSTracer: 0 envs with 1 other files remaining 
Traceback (most recent call last):
[...]

PR #219, opened 2018-05-01
BF - Cleaned up docker crashing and put in code snippet to handle ker…

mjtravers:
…nel seccomp flag when creating containers

This commit addresses:

1. Docker crashing for issue #212 
2. Adding a flag for turning off seccomp when starting up containers so the rztrace binary can work
3. Removed throwing an exception in retrace if the inspected Docker image does not have a RepoDigest

### DIFF
--- a/niceman/distributions/docker.py
+++ b/niceman/distributions/docker.py
@@ -123,11 +123,10 @@ class DockerTracer(DistributionTracer):
                 image = json.loads(self._session.execute_command(['docker',
                     'image', 'inspect', file])[0])[0]
 
-                # Fail if the Docker image has not been pushed to a repository.
-                # If the image is not in a repository, then it can't be
-                # expected to be reasonably discovered by others to reproduce.
+                # Warn user if the image does not have any RepoDigest entries.
                 if not image['RepoDigests']:
-                    raise CommandError(msg='No Docker repos found')
+                    lgr.warning("The Docker image '{}' does not have any \
+repository IDs associated with it".format(file))
 
                 images.append(DockerImage(
                     id=image['Id'],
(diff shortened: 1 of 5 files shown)


niceman/distributions/docker.py
```diff
@@ -123,11 +123,10 @@ def identify_distributions(self, files):
...
                     'image', 'inspect', file])[0])[0]
 
-                # Fail if the Docker image has not been pushed to a repository.
-                # If the image is not in a repository, then it can't be
-                # expected to be reasonably discovered by others to reproduce.
+                # Warn user if the image does not have any RepoDigest entries.
                 if not image['RepoDigests']:
-                    raise CommandError(msg='No Docker repos found')
+                    lgr.warning("The Docker image '{}' does not have any \
+repository IDs associated with it".format(file))
```
kyleam: Very minor, but just a note.  With logging messages, it's preferable to use something like `lgr.warning("a %s", foo)` because that avoids unnecessarily formatting the string when the logger isn't enabled for that level.

niceman/resource/docker_container.py
```diff
@@ -42,6 +42,8 @@ class DockerContainer(Resource):
         doc="Docker base image ID from which to create the running instance")
     engine_url = attrib(default='unix:///var/run/docker.sock',
         doc="Docker server URL where engine is listening for connections")
+    seccomp = attrib(default='on',
+        doc="Disable kernel secure computing mode when creating the container")
```
kyleam: With the doc saying "Disable kernel secure computing mode ..." I don't think it's clear that `default=on` means the code does *not* disable it.

Maybe make `seccomp_unconfined` the attribute (defaulting to `False`)?

niceman/distributions/docker.py
mjtravers: Thanks for the tip! I've made the change.

niceman/resource/docker_container.py
mjtravers: Yes, your suggestion is more clear and intuitive. Made the change.

PR #227, opened 2018-05-16
debian: Improve warning about multiple packages found by dpkg-query -S

### LINKED ISSUE #224: handle "dpkg-query line has multiple packages" without issuing a warning
ATM we issue a **long** warning listing many packages which renders output unusable.  We should get back to the issue and see how it could be resolved correctly (deduce which package should be included in the spec, or include multiple if applicable)

kyleam:
* Don't warn about  "multiple packages" when the path is a directory because the Debian tracer doesn't handle directories and these warnings flood the screen.
   * Fix "multiple package" test to remove false positives.

Fixes #224.

### DIFF
--- a/niceman/distributions/debian.py
+++ b/niceman/distributions/debian.py
@@ -8,6 +8,7 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Support for Debian(-based) distribution(s)."""
 import os
+import re
 
 import itertools
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


kyleam:
> I've restarted master build https://travis-ci.org/ReproNim/niceman/builds/377849365 to see if it is also present in master

Thanks.  master does seem to be failing with the same error.

kyleam:
> master does seem to be failing with the same error.

I've rebased the branch (no content changes) to trigger a rebuild with the fixed master.

PR #235, opened 2018-05-24
BF - Replaced the Docker image tag for the SSH testing container with…

mjtravers:
… its RepoDigest.

This commit removes Kyle's "apt-get update" and replaces the SSH Docker image tag with its RepoDigest hash.


yarikoptic:
Let's see if that digest will remain available for much longer ;-)

PR #236, opened 2018-05-26
Make default=None  the default in attrib; use attrib consistently

### LINKED ISSUE #234: make our niceman/resource/base.py:attrib to default to default=None and use it throughout

kyleam:
Closes #234.

### DIFF
--- a/niceman/base.py
+++ b/niceman/base.py
@@ -9,6 +9,7 @@
 """Base classes"""
 
 import attr
+from .utils import attrib
 
 
 @attr.s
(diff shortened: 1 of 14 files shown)


yarikoptic:
Thank you!

kyleam:
I guess codecov is acting up since it didn't add a report here?

kyleam:
@yarikoptic @mjtravers any objections to these changes, particularly b836df19?

yarikoptic:
no objections... that NOTHING was sticking out for me, but it was better than default everywhere so I think it is the way to go.  I have restarted travis build with the hope that may be it awakens codecov

yarikoptic:
ok, I will merge before we start digging into conflict resolutions too much (about to fixup conflicts for #139 )

PR #247, opened 2018-06-26
TST: Drop nose dependency, remove yield tests

### LINKED ISSUE #82: convert nose style generator tests to pytest parametric
those warnings like
```
WC1 /home/yoh/proj/repronim/niceman/niceman/tests/test_api.py yield tests are deprecated, and scheduled to be removed in pytest 4.0
```
are annoying so we better off just converting them.  I believe that `@pytest.mark.parametrize` should be the way to do it, example of similar conversion:
https://github.com/duecredit/duecredit/pull/103/files#diff-da29be34342f39369f63abe4a138618aR137

kyleam:
This PR

- [x] removes nose as a dependency
- [x] replaces yield tests by using `pytest.mark.parametrize` (closes #82)
- [x] various test cleanup


kyleam:
Hmm, looks like there are some pytest compatibility issues:

```
____________ ERROR collecting niceman/interface/tests/test_exec.py _____________
niceman/interface/tests/test_exec.py:26: in <module>
    scope='module'
niceman/tests/utils.py:435: in newfunc
    check_and_raise()
niceman/tests/utils.py:427: in check_and_raise
    allow_module_level=True)
E   TypeError: skip() got an unexpected keyword argument 'allow_module_level'

```

I was running with pytest-3.5.1 locally.

yarikoptic:
wow.  Thanks for doing that!  It would make it harder to copy some functionality including tests from datalad (yet to look/comment to #248) but a "right thing to do", so thanks! ;-)

yarikoptic:
Let's get into that bright future without nose! ;-)

PR #248, opened 2018-06-26
BF: dochelpers: Avoid FutureWarning about non-empty regexp match

kyleam:
This applies the same fix as DataLad's 8a87175b9.

### DIFF
--- a/niceman/dochelpers.py
+++ b/niceman/dochelpers.py
@@ -145,8 +145,10 @@ def _split_out_parameters(initdoc):
 
 
 __re_params = re.compile('(?:\n\S.*?)+$')
-__re_spliter1 = re.compile('(?:\n|\A)(?=\S)')
+__re_spliter1 = re.compile('\n(?=\S)')
 __re_spliter2 = re.compile('[\n:]')
+
+
 def _parse_parameters(paramdoc):
     """Parse parameters and return list of (name, full_doc_string)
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
I wondered if we should just copy entire dochelpers.py from datalad. Note that originally I've created a branch `3rd` purpose of which is to contain those pristine copies of materials borrowed, so they could be merged into master to incorporate our changes.  Not sure if it is still applicable but you could try.

kyleam:
> Note that originally I've created a branch 3rd purpose of which is to contain those pristine copies of materials borrowed, so they could be merged into master to incorporate our changes

Thanks for reminding me about that.  I'll give it a try.

kyleam:
> Note that originally I've created a branch 3rd purpose

FTR, at this point origin/3rd is essentially a really old master (it contains a lot of niceman-specific commits), with one additional non-master commit on top (adding StarCluster files).  3rd-repronimed, on the other hand, looks right.

yarikoptic:
ah, could be, I guess I might have messed up somewhere

kyleam:
> 3rd-repronimed, on the other hand, looks right.

Well, looks better, but, as the name suggests, that's of course already repronimed a bit, so not what we're after.  But looking closer, the 3rd on your remote does look good.  I'll tag the current origin/3rd for posterity then update it to match yarikoptic/3rd, with the dochelper changes on top.

PR #251, opened 2018-06-28
NF: Add initial support for "installing" Git repositories 

kyleam:
This adds basic support for instantiating Git repositories listed in a
GitDistribution object.  If a repository isn't present, we clone it
from one of the recorded sources.  Otherwise, if the recorded revision
isn't present, we fetch a remote that is marked as containing that
revision.  GitDistribution.install_packages' main goal is to check
out the same revision, but it also sets up the recorded remotes and
checks outs the recorded branch if possible.

~Right now the tests only cover the simpler cases.~

---

- [x] extend test coverage

### DIFF
--- a/niceman/distributions/tests/test_vcs.py
+++ b/niceman/distributions/tests/test_vcs.py
@@ -10,7 +10,7 @@ import os
 
 import attr
 
-from niceman.cmd import Runner
+from niceman.cmd import GitRunner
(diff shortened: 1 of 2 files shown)


kyleam:
OK, this is now ready for review/testing.

kyleam:
The failures are an unrelated docker issue (#254).

PR #253, opened 2018-06-29
BF: conda install: Work with packages and channels as SpecObject's

kyleam:
Conda's installation code originally handled packages and channels as
dicts because NicemanProvenance stopped converting dicts to objects at
that depth.  This is no longer the case as of 3b8a9884 (RF+BF: read
spec recursively, 2018-05-07).


yarikoptic:
Thanks @kyleam !

PR #256, opened 2018-07-09
TST: docker: Don't assume that tags map stably to image IDs

### LINKED ISSUE #254: Recent docker test failures
Seems like the hardcoded alpine ID is stale:

```
______________________________ test_docker_trace _______________________________
    @skip_if_no_network
    @skip_if_no_docker_engine
    def test_docker_trace():
        client = docker.Client()
        client.pull('alpine:3.6')

        tracer = DockerTracer()

        files = ['alpine:3.6', 'non-existant-image']
        dist, remaining_files = next(tracer.identify_distributions(files))
        assert dist.name == 'docker'
>       assert dist.images[0].id == 'sha256:77144d8c6bdce9b97b6d5a900f1ab85da' + \
            '325fe8a0d1b0ba0bbff2609befa2dda'
E       AssertionError: assert 'sha256:da579...3637cce406126' == 'sha256:77144d...f2609befa2dda'
E         - sha256:da579b235e92c861790f69d621d4601f54892bef262fa3fc38e3637cce406126
E         + sha256:77144d8c6bdce9b97b6d5a900f1ab85da325fe8a0d1b0ba0bbff2609befa2dda
niceman/distributions/tests/test_docker.py:32: AssertionError
```

([source](https://travis-ci.org/ReproNim/niceman/jobs/400948220))

kyleam:
As recent test failures show (gh-254), this isn't the case.

Break apart test_docker_trace into two tests.  The first pulls by a
tag and traces the image, making no assumptions about the actual IDs.
The second pulls by the digest so that we can make assertions with
known IDs.

Also, temporarily disable assertions about the tags of installed
"packages" because these repo tags won't be present if another local
image is already has with that tag (e.g., an updated alpine@3.6).

This approach should eliminate failures due to changing IDs as long as
the untagged repo digests remain available on DockerHub.  However, we
should eventually generate/upload our own test images so that we can
control the image tags.

Re: #203
Closes #254.

### DIFF
--- a/niceman/distributions/tests/test_docker.py
+++ b/niceman/distributions/tests/test_docker.py
@@ -38,6 +38,13 @@ def test_docker_trace():
     assert dist.images[0].created == '2018-01-09T21:10:38.538173323Z'
     assert 'non-existent-image' in remaining_files
 
+
+@skip_if_no_network
+@skip_if_no_docker_engine
+def test_docker_trace_local_image():
+    client = docker.Client()
+    client.pull('alpine:3.6')
+    tracer = DockerTracer()
     # Test tracing a local image not saved in a repository
     container = client.create_container(image='alpine:3.6',
         command='echo foo > test.txt')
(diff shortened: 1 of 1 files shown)


yarikoptic:
LGTM, thanks!

Issue #246, opened 2018-06-25
'SSHClient' object has no attribute '_host'


yarikoptic:
Tried to login but got above error:
```
$> niceman login -n discovery                                                                      
2018-06-25 14:11:50,386 [ERROR  ] 'SSHClient' object has no attribute '_host' [ssh.py:interactive_shell:210] (AttributeError) 
```

paramiko 2.4.0  scp 0.10.2 (if somehow related)

mjtravers:
This was fixed in PR #264

PR #264, opened 2018-07-16
Changed SSH login code to use low level Paramiko Transport class

mjtravers:
This commit is for Issue #246

### DIFF
--- a/niceman/resource/ssh.py
+++ b/niceman/resource/ssh.py
@@ -286,12 +286,15 @@ class SSHSession(POSIXSession):
         if uid > -1 or gid > -1:
             self.chown(dest_path, uid, gid, remote=False)
 
+    @borrowdoc(POSIXSession)
     def exists_command(self, path):
         return ['bash', '-c', '"test', '-e', path, '&&', 'echo', 'Found"']
 
+    @borrowdoc(POSIXSession)
     def isdir_command(self, path):
         return ['bash', '-c', '"test', '-d', path, '&&', 'echo', 'Found"']
 
+    @borrowdoc(POSIXSession)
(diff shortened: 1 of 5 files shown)


niceman/resource/ssh.py
```diff
@@ -169,13 +279,27 @@ def get(self, src_path, dest_path, uid=-1, gid=-1):
...
         if uid > -1 or gid > -1:
             self.chown(dest_path, uid, gid, remote=False)
 
+    @borrowdoc(POSIXSession)
+    def exists_command(self, path):
+        return ['bash', '-c', '"test', '-e', path, '&&', 'echo', 'Found"']
+
+    @borrowdoc(POSIXSession)
+    def isdir_command(self, path):
+        return ['bash', '-c', '"test', '-d', path, '&&', 'echo', 'Found"']
```
kyleam: This class, `SSHSession`, inherits from `POSIXSession`, and that already defines `exists_command` and `isdir_command` (as well as `get_mtime_command` below), so isn't overriding them here unnecessary?  I didn't spot any differences in the intended command, though I do think the variants defined here will always fail with a bash syntax error.

mjtravers: It's not the command that is troublesome but the *double-quotes* for the test command run in the bash shell. There is a difference in how the resulting command after the -c is passed to the execute_command() method which, in turn, is handled slightly differently by the docker, runner, and paramiko packages. I was unable to discover a common quoting pattern that all 3 execution handlers would handle without error in the remote shell.

kyleam: Ah, I see.  Thanks for the explanation.

Issue #263, opened 2018-07-11
add url to singularity definition


yarikoptic:
even though we can't trace it (as code comment in #204 notes), users could specify within their specs manually, then we could use it to 'pull' and if checksum is provided we could verify

mjtravers:
Added url in PR #265

PR #265, opened 2018-07-17
Added ability to retrace Singularity hub urls

mjtravers:
Commit addresses issue #263 

TODOs
- [x] add `path` field to the SingularityImage class
- [x] install image if `url` is known, if also `path` is known -- pull into that location


niceman/distributions/tests/test_singularity.py
```diff
@@ -46,3 +46,16 @@ def test_singularity_trace():
 
     # Clean up test files
     os.remove(image_file)
+
+    # Run tracer against a Singularity Hub url
+    files = ['shub:/GodloveD/lolcow']
```
kyleam: If this image is updated, this test will start to fail.  Pulling with the GitHub commit (appending "@a59d8de3121579fe9c95ab8af0297c2e3aefd827") would avoid that.

niceman/distributions/singularity.py
```diff
@@ -95,14 +100,30 @@ def identify_distributions(self, files):
...
         for file_path in files:
             try:
-                image = json.loads(self._session.execute_command(['singularity',
-                    'inspect', file_path])[0])
+                if file_path.startswith('shub:/'):
+                    # Correct file path for path normalization in retrace.py
+                    if not file_path.startswith('shub://'):
+                        file_path = file_path.replace('shub:/', 'shub://')
+                    temp_path = "{}.simg".format(uuid.uuid4())
+                    with chpwd(tempfile.gettempdir()):
```
yarikoptic: I think there should be some log message (lgr.info) to announce that we will download the thing to be "traced"

yarikoptic:
please
- [ ] add `path` field to the SingularityImage class
- [ ] install image if `url` is known, if also `path` is known -- pull into that location

yarikoptic:
see https://github.com/singularityware/singularity/issues/1775#issuecomment-406655734 about url to specify exact version

mjtravers:
All comments have been resolved.

PR #273, opened 2018-07-26
RF: satisfies, diff and alike

yarikoptic:
The idea in this PR to make comparison (diff) and satisfaction check logic to be implemented in a generic way, so only that semantic or custom handling would be assigned to the corresponding subclasses of SpecObject or their attributes.

- [x] `SpecObject.is_satisfied_by` has replaced the custom `DebianPackage.satisfies` without breaking any tests
- [x] RF `DebianDistribution.satisfies` so it could be replaced by the generic `is_satisfied_by`, test that previous/tested functionality works
- [ ] RF/ENH `is_satisfied_by` into a more generic `compare` method with a `mode` (or other name) kwarg with possible values of: `satisfied_by`, `identical_to` (anything else? better names?  later on we could come up with e.g. `upgraded_to` to check if a given env is an upgraded version of an existing one)
- [ ] ENH `TypedList` to also provide the `compare`, where `satisfied_by` would succeed if provided `other` is a subset where each entry `satisfied_by`, and then identical set with `identical_to`
- [ ] RF `files = attrib(default=attr.Factory(list))` into a dedicated `files = FileList()` which would probably be just an instance of a `TypedList(str)` to provide needed semantics for comparison, but better to have a dedicated `FileList` adapter happen we need to provide anything custom beyond `TypedList`
- `diff` then could benefit from all the above to be generic:
   - [ ] RF `diff` to provide both `identical_to` and `satisfied_by` semantic
   - [x] finish RFing output to be generic
   - [ ] RF to accept >2 specs, where for
         - `identical_to` it is a regular "all are identical to each other"
         - `satisfied_by` is an incremental satisfaction -- "each next one satisfies all previous ones"

Whatever is not done within this PR should be moved into separate issues/PRs


niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+
+    # Empty tuple would signal that all attribs will be used while establishing
+    # the "identity" of the specobject
+    _cmp_fields = tuple()
+    # Fields of the primary interest when showing diff
+    _diff_fields = tuple()
+
+    @property
+    def _cmp_id(self):
+        if not self._cmp_fields:
```
kyleam: Hmm, but an empty tuple would go down this path.  That seems to contradict above which says that an empty tuple means "all attribs establish identity".

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+            # Might need to be gone or some custom exception
+            raise RuntimeError(
+                "Cannot establish identity of %r since _cmp_fields "
+                "are not defined", self)
```
kyleam: This message won't be formatted for you.

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        fields = self._cmp_fields \
+            if self._cmp_fields \
+            else self._attr_names
```
kyleam: or `fields = self._cmp_fields or self._attr_names`

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        return tuple(getattr(self, a) for a in fields)
+
+    # TODO: make it "lazy" or may be there is already a helper in attrs?
+    @property
+    def _attr_names(self):
+        return tuple(f.name for f in self.__attrs_attrs__)
+
+    # TODO: code here is too generic and could be used for "is_identical_to"
+    #   kind of check, just logic could be adjusted
+    def is_satisfied_by(self, other):
```
kyleam: I like the name `is_statisfied_by`, though I think dropping the "is" (`satisfied_by`) would be better.

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        """does the other spec object satisfies the requirements of the current one
```
kyleam: s/satisfies/satisfy/

Stylistically, I'd prefer

    Does the spec object `other` satisfy the requirements of this one?

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        """
+        # TODO(OPT): might explicitly set to no op function if lgr level is too high
+        log_cmp = lambda *args: lgr.log(5, *args)
+        log_cmp("Comparing (satisfy) self=%r to other=%r", self, other)
+        if not isinstance(other, self.__class__):
+            log_cmp("is not a subclass")
+            return False
+
+        # ??? probably not appropriate for 'satisfy' since e.g. pkg could be
+        #     missing arch spec and thus be satisfied
```
kyleam: Is arch the single exception here?  In general, I would have assumed that if `other` doesn't have the same cmp_id, it shouldn't be able to satisfy a requirement.  Perhaps we can check if the cmp_id fields are equal, ignoring those that are missing?

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        # if self._cmp_id != other._cmp_id:
+        #     log_cmp("%r is different from %r since cmd_id differ",
```
kyleam: s/cmd_id/cmp_id/

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        #             self._cmp_id, other._cmp_id)
```
kyleam: I know this is commented out, but assuming it might be added back, at least somewhere: Your log message makes it seem like you're passing `self` and `other`, but then you give the `cmp_id`s as arguments.  IOW it's redundant to say "id1 and id2 are different since the ids are different".

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+
+        # TODO: any benefit from using ordered set may be to fail earlier etc?
+        # recently added OrderedSet to nibabel -- borrow from there
+        self_attrs = set(self._attr_names)
+        other_attrs = set(other._attr_names)
+        all_attrs = self_attrs.union(other_attrs)
+
+        # if `other` has some field set which is not set in `self` - that is Ok
+        # although (TODO) think about files!? (probably we would need custom
+        # satisfaction check for some fields like `files`)
```
kyleam: I'm not following the files bit.  If, say, a conda package was retraced with a particular file input, why should it matter?  Doesn't it just matter that the package is satisfied? I have version N of package X, then that file should be there.

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+        # if `self` has some field set which is not set in `other` we cannot
+        # say that it satisfies
+        # if field set in both but different, cannot be satisfied
+        for f in all_attrs:
+            def differ(msg, *args):
+                log_cmp("Difference %s.%s: " + msg, self, f, *args)
+                return False
+            if f not in self_attrs:
```
kyleam: Since `f` is used throughout a large block of text rather than a narrow scope like a comprehension, I'd prefer it have a more informative name.

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+            # TODO: could getattr be avoided if we just use actual attrs?
+            self_v = getattr(self, f)
+            other_v = getattr(other, f)
+            if self_v is None:
+                # we do not have it set at all, so other is ok regardless
+                continue
+
+            # TODO: may be some fields could be allowed to be not
+            # set (e.g. checksums etc) and should not trigger 'dis-satisfaction'
+            # if they are not defined!?  think/add
```
kyleam: If the sole purpose is reporting to the user (rather than some sort of programatic use), I think we should just report it and let the user decide if the discrepancy is something they care about.

niceman/distributions/base.py
```diff
@@ -40,6 +40,108 @@ def TypedList(type_):
...
+            #  Will actually be needed for anything non trivial, e.g. a list
+            #  of packages/repositories/whatnot!  So could be applicable even
+            #  if a value is a SpecObject
+            if isinstance(self_v, SpecObject):
+                if not self_v.is_satisfied_by(other_v):
+                    return differ("the other does not satisfy self")
+            else:
+                # For now (and otherwise) - just a simple identity check
+                if self_v != other_v:
+                    return differ("values differ %r != %r", self_v, other_v)
```
kyleam: "values differ" could be dropped.

niceman/distributions/vcs.py
```diff
@@ -79,22 +79,24 @@ class VCSRepo(SpecObject):
...
+    # @property
+    # def identifier(self):
+    #     try:
+    #         return getattr(self, self._identifier_attribute)
+    #     except AttributeError:
+    #         # raised if _identifier_attribute is not defined, but this means
+    #         # (to the caller) that identifier is not defined
+    #         msg = "%s instance has no attribute 'identifier'" % self.__class__
+    #         raise AttributeError(msg)
+    #
```
kyleam: IMO there's no reason to leave commented out code unless it comes with a TODO comment.

niceman/interface/diff.py
```diff
@@ -34,6 +34,13 @@ def __init__(self, cls):
         self.cls = cls
 
 
+def _make_plural(s):
+    """Poor man 'plural' version for now"""
+    if s.endswith('repository'):
+        return s.replace('repository', 'repositories')
+    else:
+        return s + 's'
```
kyleam: This is currently only used in one place.  Do you plan to use it elsewhere?  If not it doesn't seem worth having a separate function.

niceman/interface/diff.py
```diff
@@ -62,17 +69,22 @@ def __call__(prov1, prov2):
...
+                (DebianDistribution, "Debian package"),
+                (CondaDistribution, "Conda package"),
+                (GitDistribution, 'Git repository'),
+                (SVNDistribution, 'SVN repository')
+        ):
 
             dist_1 = env_1.get_distribution(dist_type)
             if dist_1:
-                pkgs_1 = { p._cmp_id: p for p in dist_1.packages }
+                pkgs_1 = {p._cmp_id: p for p in dist_1.packages}
```
kyleam: It'd be better if you'd do purely cosmetic changes like this and the one below in a separate commit.

niceman/interface/diff.py
```diff
@@ -81,9 +93,10 @@ def __call__(prov1, prov2):
     
             pkgs_only_1 = pkgs_1_s - pkgs_2_s
             pkgs_only_2 = pkgs_2_s - pkgs_1_s
-    
+
             if pkgs_only_1 or pkgs_only_2:
-                print(pkg_type + 's:')
+                print(_make_plural(pkg_type) + ':')
+
```
kyleam: When I run diff on a spec with Git repos, it fails because the code below assumes a `version` attribute.

niceman/interface/diff.py
```diff
@@ -104,48 +117,48 @@ def __call__(prov1, prov2):
...
+        #
+        #     for repo_id in pkgs_1_s.intersection(pkgs_2_s):
+        #         repo_1 = pkgs_1[repo_id]
+        #         repo_2 = pkgs_2[repo_id]
+        #         if repo_1.commit == repo_2.commit:
+        #             continue
+        #         print('%s pkgsitory %s:' % (repo_type, repo_id))
+        #         print('< %s (%s)' % (repo_1.commit, repo_1.path))
+        #         print('---')
+        #         print('> %s (%s)' % (repo_2.commit, repo_2.path))
```
kyleam: Is there any reason to leave this code commented out rather than deleting it?

niceman/interface/diff.py
chaselgrove: There were a number of tests that were failing; I've fixed these.

niceman/distributions/base.py
chaselgrove: I'm not sure I like the semantics of an empty _cmp_fields meaning use all of the attributes.  I'd rather require classes to define _cmp_fields explicitly and let the error be raised if they aren't.  This will mean one less special case we have to keep track of ("nothing means everything") and one more defense against a forgetful coder not thinking about _cmp_fields in a new class.

niceman/distributions/base.py
chaselgrove: I agree.  I'm happy to wait for a real case to show itself before trying to figure this one out.

niceman/distributions/base.py
chaselgrove: Fixed.

niceman/distributions/base.py
chaselgrove: +1

niceman/distributions/base.py
chaselgrove: Fixed.

niceman/distributions/vcs.py
chaselgrove: A comment later in the code referred to the block. :)

Gone.

niceman/interface/diff.py
chaselgrove: Probably left over to refer to during refactoring.  Gone now.

niceman/distributions/base.py
kyleam: @chaselgrove Sounds reasonable to me.  I've updated the comment to resolve the comment/code mismatch.

chaselgrove: Thanks.  Also removed the fallback to all attributes in the code.

niceman/distributions/base.py
chaselgrove: Fixed.

niceman/distributions/base.py
chaselgrove: Removed (see above).

niceman/distributions/base.py
kyleam: Ah, thanks, I hadn't spotted that.  So there was originally a code/comment mismatch *and* a code/code mismatch (the fallback was inaccessible due to the `not self._cmp_fields` guard).

niceman/distributions/base.py
chaselgrove: Changed.

niceman/distributions/base.py
```diff
@@ -50,6 +101,38 @@ def yaml_representer(dumper, data):
...
+        if not isinstance(other, self.__class__):
+            raise TypeError('incompatible specobject types')
+        for attr_name in self._satisfies_fields:
+            self_value = getattr(self, attr_name)
+            other_value = getattr(other, attr_name)
+            if self_value is None:
+                continue
+            if self_value != other_value:
+                return False
+        return True
```
yarikoptic: snippet accompanying removed code if you decide to roll back
```shell
$> git diff
diff --git a/niceman/distributions/base.py b/niceman/distributions/base.py
index 13f6f9d..7dbb07c 100644
--- a/niceman/distributions/base.py
+++ b/niceman/distributions/base.py
@@ -156,6 +156,11 @@ class SpecObject(object):
             if isinstance(self_v, SpecObject):
                 if not self_v.satisfied_by(other_v):
                     return differ("the other does not satisfy self")
+            elif isinstance(self_v, TypedList):
+                return all(
+                    any(self_v_e.satisfied_by(o) for o in other_v)
+                    for self_v_e in self_v
+                )
             else:
                 # For now (and otherwise) - just a simple identity check
                 if self_v != other_v:
```

yarikoptic:
seems, minor python3 issues, e.g.
```
  File "/home/travis/build/ReproNim/niceman/niceman/interface/diff.py", line 230
    print package.identity_string
                ^
SyntaxError: Missing parentheses in call to 'print'
making output directory...
```
and wonder what is your feel @chaselgrove about merging current state into master for the rename  (#359)

chaselgrove:
> seems, minor python3 issues, e.g.
> 
> ```
>   File "/home/travis/build/ReproNim/niceman/niceman/interface/diff.py", line 230
>     print package.identity_string
>                 ^
> SyntaxError: Missing parentheses in call to 'print'
> making output directory...
> ```

Fixed by a commit I hadn't pushed.
 
> and wonder what is your feel @chaselgrove about merging current state into master for the rename (#359)

Sure, let's merge and reopen the other points as needed.

yarikoptic:
Please try the diff on these two traces http://www.onerussian.com/tmp/niceman-traces-1.tgz  which were generated I believe from two installations of niceman, one was done with just '.' and another one '.[full]'

yarikoptic:
one of the development "use cases" is to use core functionality here to provide comparison/matching of the Distributions and Packages for the purpose of fixing those duplicates.  chicken-and-egg problem of a kind

yarikoptic:
fresh issue happens with the absence of unassociated files
```
> /home/yoh/proj/repronim/reproman/reproman/distributions/base.py(475)__init__()
-> raise TypeError('objects cannot both be None')
(Pdb) l
470  	    def __init__(self, a, b):
471  	        if not isinstance(a, (SpecObject, list, type(None))) \
472  	            or not isinstance(b, (SpecObject, list, type(None))):
473  	            raise TypeError('objects must be SpecObjects or None')
474  	        if not a and not b:
475  ->	            raise TypeError('objects cannot both be None')
476  	        if a and b and type(a) != type(b):
477  	            raise TypeError('objects must be of the same type')
478  	        self.cls = type(a) if a is not None else type(b)
479  	        self.a = a
480  	        self.b = b
(Pdb) up
> /home/yoh/proj/repronim/reproman/reproman/interface/diff.py(110)diff()
-> diffs.append({'diff': SpecDiff(env_1.files, env_2.files),
(Pdb) p env_1.files
[]
(Pdb) 
(dev3) 1 13710 ->1.....................................:Thu 16 May 2019 02:42:30 PM EDT:.
(git)hopa:~/proj/repronim/reproman[rf-diff]git
$> cat /tmp/sample2.spec
# ReproMan Environment Configuration File
# This file was created by ReproMan 0.1.0 on 2019-05-16 14:40:24.032705
version: 0.0.1
distributions:
- name: debian
  apt_sources:
  - name: apt_Debian_testing_main_0
    component: main
    archive: testing
    architecture: amd64
    codename: buster
    origin: Debian
    label: Debian
    site: http.debian.net
    archive_uri: http://http.debian.net/debian
    date: '2019-05-15 03:19:32+00:00'
  - name: apt_Debian_unstable_main_0
    component: main
    archive: unstable
    architecture: amd64
    codename: sid
    origin: Debian
    label: Debian
    site: http.debian.net
    archive_uri: http://http.debian.net/debian
    date: '2019-05-15 03:19:32+00:00'
  - name: apt__now__0
    archive: now
    archive_uri: /var/lib/dpkg/status
  packages:
  - name: login
    version: 1:4.5-1.1
    architecture: amd64
    source_name: shadow
    size: '750700'
    md5: 866c75edc8c06c95a7f5c343e61d30b4
    sha256: fa2d9a8686e5a1d0fef5d7f4c3a1bbd9cb69ee0fa506eec8c6c5528ababe7905
    versions:
      1:4.6-1.1:
      - apt_Debian_testing_main_0
      - apt_Debian_unstable_main_0
      - apt__now__0
    install_date: '2018-12-21 21:43:35+00:00'
    files:
    - /usr/bin/lastlog
  version: buster/sid
```

PR #274, opened 2018-08-01
Refactor ResourceManager and CLI resource references

kyleam:
This series revives aspects of @yarikoptic's gh-98, particulary the parts concerning the refactoring of `ResourceManager`.

The main parts:

-   [X] `--name` and `-id` have been consolidated into a single argument. On the command line, this argument is either a positional argument `RESOURCE` or an optional argument `--resource`, depending on the resource's role in that command.

    In the code, this "resource reference" always maps to the variable name `resref` so that it's clear that it's an identifier for a resource than a resource instance.

    If `name` and `id` for two different resources happen to collide, the `--resref-type` argument can be used to specify whether the reference is a name or ID.

    I'm not sure if I like these names (or whether the "resref" term should be exposed via the `--resref-type` option), so let me know if you have any other ideas.

-   [X] Callers' interactions with `ResourceManger` have been simplified because callers are no longer responsible for working with the inventory directly.

-   [X] `ResourceManager.get_resource_info` has been broken up so that a single method is no longer responsible for both creation and querying of resources.

-   [X] A single instance of the `ResourceManager` is used.

-   [X] `--config` is now a global option (i.e. `niceman --config FILE <command>`) rather than an option for individual commands.


niceman/resource/base.py
```diff
@@ -73,141 +119,84 @@ def factory(config):
...
-            config['type'] = type_
-        if id_:
-            config['id'] = id_
-
-        return config, inventory
-
-    @staticmethod
-    def get_config_manager(config_path=None):
-        """Returns the information stored in the niceman.cfg file.
+    def get_resource(self, resref, resref_type="auto"):
```
mjtravers: So, if I am reading this right, the default general configuration file (formerly known as niceman.cfg) is handled by the ConfigManager unless overridden with --config. What happens if --config is never passed at the command line?

kyleam: If --config isn't passed, ConfigManager looks for goes down its `/etc/niceman/niceman.cfg -> $XDG_CONFIG_DIRS -> $XDG_CONFIG_HOME -> $PWD/.niceman/niceman.cfg` processing, with settings from later files taking precedence.

yarikoptic:
There is no way back, only forward!

PR #281, opened 2018-08-15
BF: Put a stronger test in for testing if a particular docker contain…

mjtravers:
…er is running

### DIFF
--- a/niceman/tests/utils.py
+++ b/niceman/tests/utils.py
@@ -442,24 +442,27 @@ def skip_ssh(func=None):
 def skip_if_no_docker_container(container_name='testing-container'):
     """Test decorator that will skip a test if the Docker container the test is
     going to connect to is not running in the Docker engine.
-    
+
     Parameters
     ----------
     container_name : str
         Name of the container that needs to be running for the test to work.
-    
+
     Returns
     -------
     func
         Decorator function
-    
+
     Raises
     ------
     SkipTest
     """
     def decorator(func):
         stdout, _ = Runner().run(['docker', 'ps'])
-        if container_name not in stdout:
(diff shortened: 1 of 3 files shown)


niceman/tests/utils.py
```diff
@@ -442,24 +442,27 @@ def newfunc(*args, **kwargs):
...
     ------
     SkipTest
     """
     def decorator(func):
         stdout, _ = Runner().run(['docker', 'ps'])
-        if container_name not in stdout:
+        # Search the names of running containers in last column of stdout
+        matches = [line for line in stdout.splitlines()
+            if line.split()[-1] == container_name]
+        if len(matches) != 1:
```
kyleam: [I thought I left this comment, but now I don't see it.  Sorry if a near duplicate comes through.]

A downside of this approach is that it assumes name is in the last column, but that's not guaranteed to be stable across docker versions.  How about?

```diff
diff --git a/niceman/tests/utils.py b/niceman/tests/utils.py
index 97b1f6c8..7edaca26 100644
--- a/niceman/tests/utils.py
+++ b/niceman/tests/utils.py
@@ -443,8 +443,9 @@ def skip_if_no_docker_container(container_name='testing-container'):
     SkipTest
     """
     def decorator(func):
-        stdout, _ = Runner().run(['docker', 'ps'])
-        if container_name not in stdout:
+        stdout, _ = Runner().run(['docker', 'ps', '--quiet', '--filter',
+                                  'name=^/{}$'.format(container_name)])
+        if not stdout.strip():
             raise SkipTest("Docker container '{}' not running, skipping test \
                 {}".format(container_name, func.__name__))
         return func

```

Issue #295, opened 2018-09-11
AwsEc2 doesn't implement a get() method


yarikoptic:
and just copy to current directory under the same filename

ATM
```
$> niceman exec -r test-ec2-west-medium2 --internal get /tmp/strace-full.log         
2018-09-11 16:54:52,443 [ERROR  ] get() takes at least 3 arguments (2 given) [session.py:niceman_exec:266] (TypeError) 

$> niceman exec -r test-ec2-west-medium2 --internal get /tmp/strace-full.log /tmp/strace-full.log 
2018-09-11 16:55:02,586 [INFO   ] Executed the get command in the environment test-ec2-west-medium2 
```

kyleam:
Should be fixed.

PR #304, opened 2018-09-28
 ENH: resources: Teach get/put to infer destination basename 

kyleam:
This mostly addresses #295, but it looks like the particular example that issue gave still won't work because `AwsEc2` doesn't define a `get()` method.

---

- [x] fix failing tests

### DIFF
--- a/niceman/resource/session.py
+++ b/niceman/resource/session.py
@@ -300,7 +300,7 @@ class Session(object):
             mkdir = partial(self.mkdir, parents=True)
 
         dest_dir = op.dirname(dest_path)
-        if not exists(dest_dir):
+        if dest_dir and not exists(dest_dir):
             mkdir(dest_dir)
 
     def put(self, src_path, dest_path, uid=-1, gid=-1):
--- a/niceman/resource/session.py
+++ b/niceman/resource/session.py
@@ -38,7 +38,7 @@ class Session(object):
     def __attrs_post_init__(self):
         """
         Maintain both current and future session environments.
-        
+
         For persistent resources, we will save the environment information
         to make it available for sessions beyond the current one.
         """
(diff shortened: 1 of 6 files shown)


niceman/resource/session.py
```diff
@@ -267,31 +269,71 @@ def niceman_exec(self, command, args):
...
+            exists = op.exists
+            mkdir = os.makedirs
+            # Avoid a direct `abspath` call to preserve trailing slash.
+            dest_path = op.join(op.abspath(op.dirname(dest_path or "")),
+                                op.basename(dest_path or ""))
+        elif dest_path and op.isabs(dest_path):
+            exists = self.exists
+            mkdir = partial(self.mkdir, parents=True)
+        else:
+            raise ValueError("Remote destination paths must be absolute")
```
kyleam: I've assumed incorrectly here:

https://travis-ci.org/ReproNim/niceman/jobs/434716977#L1349

niceman/resource/session.py
```diff
@@ -267,31 +269,74 @@ def niceman_exec(self, command, args):
...
+        -------
+        The path to the destination, possibly adjusted to add the basename.
+        """
+        if local:
+            exists = op.exists
+            mkdir = os.makedirs
+        else:
+            exists = self.exists
+            mkdir = partial(self.mkdir, parents=True)
+
```
mjtravers: Interesting way to unclutter your if statements below by setting up your function calls here. Will remember the partial() trick.

yarikoptic:
looks good to me, added just a minor sugaring in 7ccdfee94c0d381a962ad4bc97f9768b864f069e

kyleam:
> looks good to me, added just a minor sugaring in 7ccdfee

and you addition looks good to me, thanks

PR #306, opened 2018-10-01
 TST: resource: Prevent proliferation of Singularity instances

kyleam:
```
Adjust the resource_session fixture to delete singularity instances so
that they don't pollute the user's list of running instances.  Without
this, each run of test_session_abstract_methods with
SingularitySession adds entries like

  $ singularity instance.list
  DAEMON NAME      PID      CONTAINER IMAGE
  70ee38d1a70      21382    /tmp/.singularity-runtime.EcWxJv4l/python:2.7
  e9eed6fafdb      20911    /tmp/.singularity-runtime.zeQVjAch/python:2.7
```

### DIFF
--- a/niceman/resource/tests/test_session.py
+++ b/niceman/resource/tests/test_session.py
@@ -213,7 +213,7 @@ def resource_session(request):
     request : object
         Pytest request object that contains the class to test against
 
-    Returns
+    Yields
     -------
     session object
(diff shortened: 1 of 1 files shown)


yarikoptic:
pushed minor doc fix, can be merged with it.

kyleam:
> pushed minor doc fix, can be merged with it.

Thanks for spotting that.

PR #309, opened 2018-10-16
ENH: Don't list resources if loading a known resource fails

### LINKED ISSUE #297: Confusing error message
should be more exception specific I guess
```
$> niceman create -t aws-ec2 test-ec2-west-small -b region_name=us-west-1 image=ami-059e7901352ebaef8 security_group=ssh-only key_name=yoh@hopa-west2 instance_type=t2.small  
2018-09-13 14:16:49,560 [ERROR  ] Failed to import resource: cannot import name Connection [ssh.py:<module>:14].  Known ones are: aws_ec2, docker_container, session, shell, singularity, ssh [base.py:factory:140] (ResourceError) 
```
that was because of outdated fabric module which had no Connection

kyleam:
Usually importing a resource fails because the user has passed an
unknown resource type, so we report the known resources in the
exception message.  But if the import of a known resource fails (e.g.,
because the resource imports an uninstalled module), it's confusing to
show a list of known resources.

Closes #297.

### DIFF
--- a/niceman/resource/base.py
+++ b/niceman/resource/base.py
@@ -134,10 +134,13 @@ class ResourceManager(object):
         except Exception as exc:
             # Typically it should be an ImportError, but let's catch and recast
             # anything just in case.
-            msg = exc_str(exc)
-            known = ResourceManager._discover_types()
-            if module_name not in known:
-                msg += ". Known ones are: {}".format(", ".join(known))
+            try:
+                msg = exc_str(exc)
+                known = ResourceManager._discover_types()
+                if module_name not in known:
+                    msg += ". Known ones are: {}".format(", ".join(known))
+            except Exception as exc2:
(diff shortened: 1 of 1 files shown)


niceman/resource/base.py
```diff
@@ -134,10 +134,12 @@ def factory(config):
         except Exception as exc:
             # Typically it should be an ImportError, but let's catch and recast
             # anything just in case.
+            msg = exc_str(exc)
+            known = ResourceManager._discover_types()
```
yarikoptic: I wouldn't be surprised if we manage to raise exception there as well, so may be try/except these 3 lines as well and add caught exception message (msg += " Failed to discover resource types: %s" % exc_str(exc2)) in except for it?

yarikoptic:
Thanks!

PR #311, opened 2018-10-18
BF: Fixed issues with creating a singularity resource

mjtravers:
Fixed 2 issues with the singularity resource:

1. When creating a new singularity resource, the name and id of the resource were getting the same value which is problematic when retrieving the resource config info.

2. Unsightly and false-positive ERROR messages were being displayed from the singularity get_instance_info() method.

### DIFF
--- a/niceman/interface/exec.py
+++ b/niceman/interface/exec.py
@@ -219,7 +219,7 @@ class Exec(Interface):
             sys.stderr.write(err)
         if error:
             lgr.error(
-                "Command %s failed to run in %s",
-                command, env_resource
+                "Command %s failed to run in %s: %s",
+                command, env_resource.name, error.msg
             )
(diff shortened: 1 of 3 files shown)


kyleam:
> @kyleam , what do you think about extending the id for singularity here?

It seems like a good idea to me, but I haven't looked into these failures yet.  Either way though ...

@mjtravers:
>     1. When creating a new singularity resource, the name and id of the resource were getting the same value which is problematic when retrieving the resource config info.

I should look into how this was failing and see if the resource manager should handle this better

mjtravers:
The test_singularity fails are updates needed to the test. Not sure about the Xwd_unchanged on yet

mjtravers:
While waiting for tests to run I added very small tweak to address #294

PR #320, opened 2018-10-29
BF: Update SSHSession to handle string command

kyleam:
Main changes: 

* Make it possible to use string commands with SSHSession because this is possible with other sessions and used by internal callers of execute_command().

* Provide better escaping for commands in SSHSession and SingularitySession.

* Provide better escaping for session methods like isdir().  This is unrelated to the initial purpose of this PR, but it was needed for adding the string/list command form tests to test_session_abstract_methods().

### DIFF
--- a/niceman/utils.py
+++ b/niceman/utils.py
@@ -12,6 +12,7 @@ import re
 import six
 
 import six.moves.builtins as __builtin__
+from six.moves import shlex_quote
 import time
 
 from os.path import curdir, basename, exists, realpath, islink, join as opj, isabs, normpath, expandvars, expanduser, abspath
@@ -1253,6 +1254,20 @@ def parse_semantic_version(version):
     else:
         raise ValueError(
             "{} does not appear to follow semantic versioning".format(version))
+
+
+def command_as_string(command):
+    """Convert `command` to the string representation.
+
+    Parameters
+    ----------
+    command : list or str
+        If it is a list, convert it to a string, quoting each element as
+        needed.  If it is a string, it is returned as is.
+    """
+    if isinstance(command, list):
+        command = " ".join(map(shlex_quote, command))
+    return command
 
 
 lgr.log(5, "Done importing niceman.utils")
--- a/niceman/resource/session.py
+++ b/niceman/resource/session.py
@@ -14,7 +14,6 @@ lgr = logging.getLogger('niceman.resource.session')
 
 import attr
 from functools import partial
-import json
 import os
 import os.path as op
 import re
--- a/niceman/resource/session.py
+++ b/niceman/resource/session.py
@@ -174,7 +174,7 @@ class Session(object):
 
         Parameters
         ----------
-        command : list
+        command : list or str
             Shell command string or list of command tokens to send to the
             environment to execute.
         env : dict, optional
@@ -206,7 +206,7 @@ class Session(object):
 
         Parameters
         ----------
-        command : list
+        command : list or str
             Shell command string or list of command tokens to send to the
             environment to execute.
         env : dict, optional
--- a/niceman/resource/session.py
+++ b/niceman/resource/session.py
@@ -17,6 +17,7 @@ from functools import partial
 import os
 import os.path as op
 import re
+from six.moves import shlex_quote
 
 from niceman.support.exceptions import SessionRuntimeError
 from niceman.cmd import Runner
@@ -601,7 +602,7 @@ class POSIXSession(Session):
 
     def exists_command(self, path):
         """Return the command to run for the exists method."""
-        command = ['test', '-e', path, '&&', 'echo', 'Found']
+        command = ['test', '-e', shlex_quote(path), '&&', 'echo', 'Found']
         return ['bash', '-c', ' '.join(command)]
 
     # def lexists(self, path):
@@ -660,7 +661,7 @@ class POSIXSession(Session):
 
     def isdir_command(self, path):
         """Return the command to run for the exists method."""
-        command = ['test', '-d', path, '&&', 'echo', 'Found']
+        command = ['test', '-d', shlex_quote(path), '&&', 'echo', 'Found']
         return ['bash', '-c', ' '.join(command)]
 
     def chmod(self, path, mode, recursive=False):
--- a/niceman/resource/singularity.py
+++ b/niceman/resource/singularity.py
@@ -10,13 +10,13 @@
 
 import attr
 import os
-import six
 from ..cmd import Runner
 from ..dochelpers import borrowdoc
 from ..support.exceptions import CommandError
 from .session import POSIXSession, Session
 from .base import Resource
 from ..utils import attrib
+from ..utils import command_as_string
 
 import logging
 lgr = logging.getLogger('niceman.resource.singularity')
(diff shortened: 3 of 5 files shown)


kyleam:
> Eh, looks like jobs are consistently failing on master (example)

One reason for the failures should be addressed with #321, but it looks like there's probably more going on.  Let's see what happens in the tests there.

kyleam:
> One reason for the failures should be addressed with #321, but it looks like there's probably more going on. Let's see what happens in the tests there.

Appears that was **the** reason.  Merged that to master and rebased this PR.

niceman/utils.py
```diff
@@ -1255,4 +1256,22 @@ def parse_semantic_version(version):
...
+
+
+def normalize_command(command):
+    """Convert `command` to the string representation.
+    """
+    if isinstance(command, list):
+        if len(command) == 1:
+            # This is either a quoted compound shell command or a simple
+            # one-item command. Pass it as is.
+            command = command[0]
```
yarikoptic: what if it is `["executable with space in its name"]`?

kyleam: In the context of `datalad run`, then the user has to the string form:

```
datalad run './stupid\ name.sh'
```

I don't see any reliable alternative.  We'd be left parsing an arbitrary string trying to guess the user's intention.  If you have a suggestion though, I think this is better suited for an issue on datalad.

yarikoptic: Ok, I will review the assumptions of the datalad run before further comments here...

yarikoptic: if I got it right, may be here in `niceman` we (for now at least) do not inherit them and always assume that the line is a list of arguments, and if a command as a string is needed, add an explicit cmdline flag for that (e.g. `--cmd-string`)?
May be actually we could add it also in datalad and have it not just bool, but a choice between bool and 'auto', and have 'auto' default which is what it is now deciding between the two modes?  Then whenever it is needed to state explicitly, user could?

kyleam: > Then whenever it is needed to state explicitly, user could?

In my view, that is more complicated and not any better than just saying that the user should just use the string form in this case.

kyleam: And to step back a bit: you're getting bogged down in a known edge case that *has* a solution (use the string form), while this PR fixes multiple issues that are more likely to be encountered and *don't have* a solution.

yarikoptic: I am "bogged down" to it because that is where with this change I came to realization of this problem exists.  "solution" is not really a solution since it implies that any script (python or shell) needing to run a single command would need always to do escaping of the command (might have spaces in the path etc) even if it provides it explicitly as a list in Python which typically means that no escaping generally is needed at that "user" end  of the story.

yarikoptic: but ok, let's deal with it later I guess if you feel that it is not "that important" and we have the issue throughout other sessions as well.  you are welcome to merge as is.

kyleam: > it implies that any script (python or shell) needing to run a single command would need always to do escaping of the command (might have spaces in the path etc)

In the context of `datalad run`, the focus is on CLI calls.  In that case, the user needs to quote or otherwise escape "path/with spaces/command.sh" in the same way they would have to in any command line call.  But ...

> even if it provides it explicitly as a list in Python which typically means that no escaping generally is needed at that "user" end of the story.

... thanks for the pushback here.  I'm wrong in trying to apply DataLad's CLI-focused `normalize_command` helper to python methods like `Session.execute_command`.   In the cases in the PR, there's no ambiguity there between list and string [0], so we can just map `shlex_quote` over the items unconditionally.  I'll push an update.

[0] The same thing is true for the rarer form of calling `Dataset.run` rather than `datalad run`.  If there is an easy way of telling that we're being called from the Python API rather than the CLI, we could change how this behavior for `Dataset.run` too.

kyleam:
Update pushed.

```
$ git range-diff master kyleam/bf-string-cmd__pr1 bf-string-cmd
 1:  17d07818 =  1:  dad8f1ae DOC: session: Consistently describe execute_command's 'command'
 2:  d9ebf211 =  2:  f1c9cb7d CLN: ssh: Drop an unused import
 3:  abcee1b9 =  3:  6ea67940 CLN: session: Drop an unused import
 4:  9a99946a =  4:  999dca4b BF: ssh: Handle a string command
 5:  e53d3039 !  5:  b3234580 ENH: utils: Add normalize_command
    @@ -1,8 +1,8 @@
     Author: Kyle Meyer <kyle@kyleam.com>
     
    -    ENH: utils: Add normalize_command
    +    ENH: utils: Add command_as_string
     
    -    A few session class will use this in _execute_command.
    +    A few session classes will use this in _execute_command.
     
     diff --git a/niceman/utils.py b/niceman/utils.py
     --- a/niceman/utils.py
    @@ -19,21 +19,17 @@
                  "{} does not appear to follow semantic versioning".format(version))
      
      
    -+# TODO: normalize_command is copied directly from datalad/interface/run.py.
    -+# It'd be better if the datalad function was moved to its datalad/utils.py, and
    -+# then we absorb normalize_command through the branch '3rd'.
    -+
    -+
    -+def normalize_command(command):
    ++def command_as_string(command):
     +    """Convert `command` to the string representation.
    ++
    ++    Parameters
    ++    ----------
    ++    command : list or str
    ++        If it is a list, convert it to a string, quoting each element as
    ++        needed.  If it is a string, it is returned as is.
     +    """
    -+    if isinstance(command, list):
    -+        if len(command) == 1:
    -+            # This is either a quoted compound shell command or a simple
    -+            # one-item command. Pass it as is.
    -+            command = command[0]
    -+        else:
    -+            command = " ".join(shlex_quote(c) for c in command)
    ++    if not isinstance(command, list):
    ++        command = " ".join(map(shlex_quote, command))
     +    return command
     +
     +
 6:  5c74dbf8 !  6:  a5e7216d BF: ssh: Provide better escaping of commands
    @@ -13,7 +13,7 @@
      
      from .base import Resource
      from ..utils import attrib
    -+from ..utils import normalize_command
    ++from ..utils import command_as_string
      from niceman.dochelpers import borrowdoc
      from niceman.resource.session import Session
      from ..support.exceptions import CommandError
    @@ -24,7 +24,7 @@
     -        if isinstance(command, list):
     -            command = ' '.join(command)
     -
    -+        command = normalize_command(command)
    ++        command = command_as_string(command)
              try:
                  result = self.connection.run(command, hide=True)
              except invoke.exceptions.UnexpectedExit as e:
 7:  34508a2f =  7:  15091268 CLN: ssh: Remove duplicate child methods
 8:  204ea620 =  8:  47125ca9 BF+ENH: session.{exists,isdir}_command: Escape path
 9:  194267a5 !  9:  f93d2150 BF: singularity: Don't convert string commands to a list
    @@ -26,7 +26,7 @@
      from .session import POSIXSession, Session
      from .base import Resource
      from ..utils import attrib
    -+from ..utils import normalize_command
    ++from ..utils import command_as_string
      
      import logging
      lgr = logging.getLogger('niceman.resource.singularity')
    @@ -40,7 +40,7 @@
              stdout, stderr = self._runner.run(
     -            ['singularity', 'exec', 'instance://' + self.name] + command,
     +            "singularity exec instance://{} {}".format(
    -+                self.name, normalize_command(command)),
    ++                self.name, command_as_string(command)),
                  expect_fail=True)
      
              return (stdout, stderr)
10:  afbf803d = 10:  40d2a53e BF: singularity: Escape paths for get and put
11:  dd41f98b = 11:  70309c03 TST: resource: Add some spaces to the test cases
12:  aa834951 = 12:  823d10ce TST: Test that each _execute_command variant can take a list or str
```

PR #321, opened 2018-10-29
BF: redhat: Don't instantiate RPMSource with unknown fields

kyleam:
test_redhat.py::test_tracer is failing on Travis with

    TypeError: __init__() got an unexpected keyword argument 'excluded'

Presumably this is due to a change in the Docker image used for this
test.

Instead of dying in such a situation, issue a warning that we've
encountered an unknown field in the 'yum --verbose repolist' output
and continue.  Perhaps we should add an "excluded" attributes to
RPMSource and/or update the test to use a stable Docker image, but we
should still make _find_all_sources resilient to unknown keys to avoid
such failures in the future.

### DIFF
--- a/niceman/distributions/redhat.py
+++ b/niceman/distributions/redhat.py
@@ -328,6 +328,7 @@ class RPMTracer(DistributionTracer):
         -------
         sources : list(RPMSource)
         """
+        attr_fields = {f.name for f in attr.fields(RPMSource)}
         sources = []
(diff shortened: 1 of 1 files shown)


kyleam:
@mjtravers, should `RPMSource` be updated with an `excluded` attribute?

kyleam:
Merging to get master out of its currently failing state.

PR #323, opened 2018-10-30
CLN+TST: Purge print statements in tests

### LINKED ISSUE #322: Reduce verbosity of tests on Travis
We're dumping too much to stdout/stderr in our tests.  Travis's online interface currently reports

    This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log.

which makes debugging failures less convenient.

Quickly glancing at [a log], some of the main offenders are

  * test_debian (biggest offender, it seems)
  * pip install
  * formats/tests/test_niceman
  * tests_retrace
  * test_docker_container.py

[a log]: https://api.travis-ci.org/v3/job/447955542/log.txt

kyleam:
All of these appear to be meant manual debugging and don't affect the
actual tests.  Remove them so that they don't pollute the test output.
This becomes more important as of 3ef0f5c0 (Added -s switch for pytest
in travis, 2018-09-13) because the test stdout/stderr is dumped to the
Travis logs, making the logs exceed Travis's online display limit.

Closes #322.

### DIFF
--- a/niceman/distributions/tests/test_debian.py
+++ b/niceman/distributions/tests/test_debian.py
@@ -11,7 +11,6 @@ import os
 from os.path import islink
 from os.path import join, isfile
 
-from pprint import pprint
 import logging
 
 import attr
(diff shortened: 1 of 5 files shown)


kyleam:
There are still some places where I think we should be quieter in the tests, but this gets us under Travis's display limits and fixes the core issue in #322.

Issue #316, opened 2018-10-26
Make --trace great again:  __call__() got an unexpected keyword argument 'resource'


yarikoptic:
could be a user "error", and just need a catch/dedicated exception (since may be I am not using `exec` correctly)?
```shell
$> niceman --dbg exec --trace /bin/ls
Enter a resource name or ID: localshell
 
Configuration file written in /home/yoh/.cache/niceman/traces/20181026114557-d1/config.yml
Edit that file then run the packer -- use 'reprozip pack -h' for help
Traceback (most recent call last):
  File "/home/yoh/proj/repronim/niceman/venvs/dev/bin/niceman", line 11, in <module>
    load_entry_point('niceman', 'console_scripts', 'niceman')()
  File "/home/yoh/proj/repronim/niceman/niceman/cmdline/main.py", line 251, in main
    ret = cmdlineargs.func(cmdlineargs)
  File "/home/yoh/proj/repronim/niceman/niceman/interface/base.py", line 288, in call_from_parser
    return cls.__call__(**kwargs)
  File "/home/yoh/proj/repronim/niceman/niceman/interface/exec.py", line 208, in __call__
    resource=session
TypeError: __call__() got an unexpected keyword argument 'resource'
()
> /home/yoh/proj/repronim/niceman/niceman/interface/exec.py(208)__call__()
-> resource=session
(Pdb) 

$> niceman  exec -r localshell --trace /bin/ls  
Configuration file written in /home/yoh/.cache/niceman/traces/20181026114708-4c/config.yml
Edit that file then run the packer -- use 'reprozip pack -h' for help
2018-10-26 11:47:08,695 [ERROR  ] __call__() got an unexpected keyword argument 'resource' [exec.py:__call__:208] (TypeError) 
```

this is on v0.0.6-215-ga266b5c

Issue #315, opened 2018-10-26
Make --trace great again:  wrong error message


yarikoptic:
```shell
$> niceman exec --trace /bin/ls
Enter a resource name or ID: localshell
 
2018-10-26 11:37:23,988 [ERROR  ] Using --trace requires ReproZip, a Linux-specific dependency [exec.py:__call__:193] (RuntimeError) 
(dev) 1 14274 ->1.....................................:Fri 26 Oct 2018 11:37:24 AM EDT:.
(git)hopa:~/proj/repronim/niceman[master]git
$> niceman --dbg exec --trace /bin/ls
Enter a resource name or ID: localshell

Traceback (most recent call last):
  File "/home/yoh/proj/repronim/niceman/venvs/dev/bin/niceman", line 11, in <module>
    load_entry_point('niceman', 'console_scripts', 'niceman')()
  File "/home/yoh/proj/repronim/niceman/niceman/cmdline/main.py", line 251, in main
    ret = cmdlineargs.func(cmdlineargs)
  File "/home/yoh/proj/repronim/niceman/niceman/interface/base.py", line 288, in call_from_parser
    return cls.__call__(**kwargs)
  File "/home/yoh/proj/repronim/niceman/niceman/interface/exec.py", line 193, in __call__
    raise RuntimeError("Using --trace requires ReproZip, "
RuntimeError: Using --trace requires ReproZip, a Linux-specific dependency
()
> /home/yoh/proj/repronim/niceman/niceman/interface/exec.py(193)__call__()
-> raise RuntimeError("Using --trace requires ReproZip, "
(Pdb) l
188  	                    local_trace_dir)
189  	
190  	            try:
191  	                from reprozip.tracer.trace import write_configuration
192  	            except ImportError:
193  ->	                raise RuntimeError("Using --trace requires ReproZip, "
194  	                                   "a Linux-specific dependency")
195  	            from rpaths import Path
196  	
197  	            # we rely on hardcoded paths in reprozip
198  	            write_configuration(
(Pdb) from reprozip.tracer.trace import write_configuration
*** ImportError: No module named distro
```

- [x] not sure what about `distro` here. Reprozip is "installed" 
```
(Pdb) import reprozip
(Pdb) print reprozip
<module 'reprozip' from '/home/yoh/proj/repronim/reprozip/reprozip/reprozip/__init__.pyc'>
```
from source git repo
```shell
(git)hopa:~/proj/repronim/reprozip[master]reprozip
$> git describe
[...]

PR #324, opened 2018-10-31
 `exec --trace` fixes and add tests

kyleam:
Main changes:

  * Fix `retrace` regression for `exec --trace` (#316).
  * Provide better encapsulation of `exec` modes (default, `--internal`, `--trace`)
  * Add some tests for `exec --trace`.
  * Use external_versions to check if reprozip is installed (#315).


kyleam:
I need to adjust test_exec_trace_docker for Travis.

niceman/interface/exec.py
```diff
@@ -32,6 +33,211 @@
 lgr = getLogger('niceman.api.exec')
 
 
+class CommandAdapter(object):
+    """Interface to kinds of commands `exec` can execute.
+
+    Each subclass should define a `execute` method and optionally defines
```
yarikoptic:
```suggestion
    Each subclass should define an `execute` method and optionally define
```

niceman/interface/exec.py
```diff
@@ -32,6 +33,211 @@
...
+        pass
+
+    def execute(self):
+        """Run the full command defined by `command` and `cmd_args`.
+
+        Returns
+        -------
+        A tuple, (stdout, stderr).
+        """
+        raise NotImplementedError
```
yarikoptic: we do use sparingly the `@abc.abstractmethod` etc sparingly through out the code, might be a good candidate to use them here as well

niceman/interface/exec.py
```diff
@@ -32,6 +33,211 @@
...
+class TracedCommand(CommandAdapter):
+    """A tracer-wrapped command.
+
+    In the pre-command step, a tracer (extracted from Reprozip) is downloaded
+    to the local machine if not already present.  This tracer is transferred to
+    the resource on which the command will be executed.
+
+    After the command is executed under the tracer, the post-command step
+    downloads the trace artifacts locally and calls `niceman retrace` on the
+    Reprozip spec.
```
yarikoptic: May be elaborate just a bit more, that with such setup reprozip is required only on the local host and not on the remote machine (if session is remote)?

niceman/interface/exec.py
```diff
@@ -32,6 +33,211 @@
...
+    """
+
+    def __init__(self, resource, command, cmd_args,
+                 remote_dir=None, local_dir=None):
+        super(TracedCommand, self).__init__(resource, command, cmd_args)
+
+        if not external_versions["reprozip"]:
+            raise RuntimeError("Using --trace requires ReproZip, "
```
yarikoptic: I think we will have more of similar errors, so may be worth adopting a dedicated exception e.g. from datalad

```python
class MissingExternalDependency(RuntimeError):
    """External dependency is missing error"""

    def __init__(self, name, ver=None, msg=""):
        super(MissingExternalDependency, self).__init__()
        self.name = name
        self.ver = ver
        self.msg = msg

    def __str__(self):
        to_str = str(self.name)
        if self.ver:
            to_str += " of version >= %s" % self.ver
        to_str += " is missing."
        if self.msg:
            to_str += " %s" % self.msg
        return to_str
```
?

niceman/interface/exec.py
```diff
@@ -32,6 +33,211 @@
...
+                                       "niceman_trace.gz")
+
+        # Remote session variables
+        self.remote_dir = remote_dir
+        self.remote_trace_dir = None
+        self.remote_tracer = None
+
+    def _prepare_local(self):
+        if not op.exists(self.local_tracer_gz):
+            import hashlib
```
yarikoptic: some log msg about the upcoming download might be appreciated

niceman/interface/exec.py
kyleam: Yes, I know, because it's on my to-do list to fix up how they are used in niceman :]  In Python 3, they don't raise the expected type error when a creating a child with unimplemented methods.  This is because the `__metaclass__ = abc.ABCMeta` isn't checked in Python 3 (which expects a `metaclass` keyword argument instead).  There's also the cosmetic issue that many of the methods marked with `abstractmethod` raise a `NotImplementedError`, which is pointless.

But for this specific case, I don't think the complication of using an ABC is worth it (especially given the py2/py3 compatibility kludge) because these classes are really just for internal organization of `exec.py`.

niceman/interface/exec.py
kyleam: will do

niceman/interface/exec.py
kyleam: will do

niceman/interface/exec.py
kyleam: good point

kyleam:
OK, test_exec_trace_docker is now passing on travis, and all suggestions but the ABC one should be addressed.

PR #325, opened 2018-11-02
BF+RF: interface: Rename "exec" to "execute"

kyleam:
```
"exec" is a poorly chosen name because it conflicts with the `exec`
statement in Python 2.  Things like

  python2 -c "import niceman.api.exec"
  python2 -c "from niceman.api import exec as nm"
  python2 -c "import niceman.interface.exec as nm_exec"

all fail with syntax errors.
```

I ran into this with #324.  This needs to be resolved in order to handle the Python 2 test errors there.

### DIFF
--- a/niceman/interface/__init__.py
+++ b/niceman/interface/__init__.py
@@ -27,7 +27,7 @@ _group_dataset = (
         ('niceman.interface.start', 'Start'),
         ('niceman.interface.stop', 'Stop'),
         ('niceman.interface.login', 'Login'),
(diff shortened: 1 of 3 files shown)


yarikoptic:
jesus... even
```shell
$> python2 -c "import niceman.api as nm; print(nm.exec)"
```
is no good.

kyleam:
> jesus... even

Yep, that was meant to be my second example in the commit message (woops).

Issue #307, opened 2018-10-02
ResourceManager should try to match short resource IDs


kyleam:
`niceman ls` reports truncated resource IDs.  As a user, I'd expect to be able to use these short IDs unless they are ambiguous.

yarikoptic:
I was even under impression that we already did that!

yarikoptic:
indeed it (or may be docker itself did it, not sure yet) used to:
```shell
(git)hopa:~/proj/repronim/niceman[tags/v0.0.6^0]git
$> niceman  login -id c6f              
Enter a resource type [docker-container]: 

root@c6f0d0afcdc4:/# exit
```

yarikoptic:
I guess could be as simple as
```suggestion
$> git diff
diff --git a/niceman/resource/base.py b/niceman/resource/base.py
index 0c21106..2bc24cd 100644
--- a/niceman/resource/base.py
+++ b/niceman/resource/base.py
@@ -182,11 +182,11 @@ class ResourceManager(object):
     def _find_resources(self, resref, resref_type):
         def from_name(x):
             return [(name, config) for name, config in self.inventory.items()
-                    if x == name]
+                    if name.startswith(x)]
 
         def from_id(x):
             return [(name, config) for name, config in self.inventory.items()
-                    if x == config.get("id")]
+                    if config.get("id", '').startswith(x)]
 
         results_name = None
         results_id = None
```

PR #330, opened 2018-11-06
BF: allow to match resources also by beginning of their name/id

yarikoptic:
- logic now permits for multiple records with the same not only ID but also name.
  I am not sure when we should allow multiple records with the same ID though
  but that assumption may be should be verified elsewhere
- we match first resources using "startswith" but then prune the result
  if an exact match found.  This way we would select "res" even if also "res-1"
  etc available.   Not yet sure if that is an 'ok' or desired behavior -- are we asking for trouble? ;-)

TODOs:

- [ ] verify that current tests don't fail
- [ ] add dedicated test(s)
- [ ] blessing of the team on possible code "unpleasadness"

Closes: #307

### DIFF
--- a/niceman/resource/base.py
+++ b/niceman/resource/base.py
@@ -182,11 +182,11 @@ class ResourceManager(object):
     def _find_resources(self, resref, resref_type):
         def from_name(x):
             return [(name, config) for name, config in self.inventory.items()
-                    if x == name]
+                    if name.startswith(x)]
 
         def from_id(x):
-            return [(name, config) for name, config in self.inventory.items()
-                    if x == config.get("id")]
+            return [(config.get("id", ''), config) for _, config in self.inventory.items()
+                    if config.get("id", '').startswith(x)]
 
         results_name = None
         results_id = None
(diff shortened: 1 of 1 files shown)


niceman/resource/base.py
```diff
@@ -182,11 +182,11 @@ def _discover_types():
     def _find_resources(self, resref, resref_type):
         def from_name(x):
             return [(name, config) for name, config in self.inventory.items()
-                    if x == name]
+                    if name.startswith(x)]
```
kyleam: Hmm, I'm not sure that we should bother supporting partial name matches.  A name should already serve as a short label, whereas partial matching of an ID is useful because they can be quite long and an initial substring likely to be unique.  I'm basically viewing this as git's hexsha versus a refname.

In addition to matching my expectations, doing so should reduce the complexity of this PR.  With `from_id`, it'd be the current pass to see if there are full matches, and, if none, a second `startswith` pass, I think.

yarikoptic: Agree about names! It did feel weird and too magic (although my ssh aliases all short) to go by  names abbreviations. Oh well, at least I had fun. I also at the end felt that just having to passes might be cleaner. Will see...

kyleam:
Closing in favor of #361.

yarikoptic:
eh, it was all just words @kyleam ... I am really closing it now ;)

Issue #314, opened 2018-10-22
`run` WiP review notes


yarikoptic:
Since there is no PR (yet), I've decided to collect some notes here on reviewing @kyleam 's WiP https://github.com/ReproNim/niceman/compare/master...kyleam:scratch/nf-run

# job spec

## resource_name

seems to require `resource_name` as the leading level, but why?  shouldn't job spec should be about job, and not about resource?  we would specify the resource as the argument to `run`, so the job could be submitted to different resources without modifying job spec itself.  But see below a related item in "ability to tune from cmdline/call?"

## PBS-agnostic job requirements?

relates to above and may be was envisioned already to be done that way later.  IMHO we should aim to avoid PBS specific parameters in the job spec (i.e.`pbs_l_args`) and define a list of the ones we would map into corresponding PBS arguments (have `requirements` with `.walltime`, `.runtime`, `.memory`) .  How does gc3pie specifies those?  E.g. [gc3pie](https://gc3pie.readthedocs.io/en/master/programmers/api/gc3libs.html?highlight=requested_walltime) provides those for jobs specification (`requested_walltime`, etc)

### ability to tune from cmdline/call?

Generally: ideally any "job spec" related change could be done in command line.  With that in mind, `command` itself could be defined in `job.yaml`, so below examples are as if we did not consider it as something special in the cmdline API for now.  So prototypical invocations could be

   - `niceman run -r discovery --js common4allmyjobs.yaml -j command "myscript.py -param1"`
      where I added `-j` to be able to override/add any job spec argument.  Levels within yaml dict structure would be separated by `.`.  I do  not yet foresee any need to reference arrays/lists (I could be wrong). 
So in this command I am pretty much mimicing what is done now, just adding `-j command` to explicitly point that it is the field within jobspec to be modified
   - `niceman run -r discovery  --js myanalysis.yaml -j requirements.memory 1gb -j requirements.walltime 120h -j comment "This is a real thing which needs a lot of time / memory" --input fullsizeddataset`
      where `myanalysis.yaml` is the analysis "specification" which already has a `command` listed (with the `{inputs}` placeholder in the command) and even possibly has already some requirements specified, but they might not be appropriate for the "fullsizeddataset", which is what I am overriding from the command line.
   - above commands should be "executable" with `-r hydra` which would be another resource which uses another batch system (condor)
[...]

PR #331, opened 2018-11-08
NF: run command

kyleam:
Here's the current (rough) state of the `run` command.  In my view, there are at least the following blockers:

- [x] Decide on what the core orchestrators are.

    Currently there are two barely operational orchestrators, both based on a local-remote pair of DataLad datasets.  One of them just stores the results with `datalad add`, and the other uses a datalad-htcondor approach to package up the outputs and reinject them into the local repository as a run commit.

    These probably need a larger rework than just polishing.  They only support shared file systems, so at the very least we should probably rename them with the idea that we may eventually support non-shared file systems.

    Even if the current ones stay as is, we need a non-DataLad variant.

    *Edit*: These have been reworked a good amount.  There still isn't any support for non-shared file systems, but the orchestrators are probably at a state where it makes sense to move on, revisiting pain points that arise.

- [x] Improve documentation, especially higher-level overview of the current design.

   *Edit*: This has been improved somewhat.  I may try to do one more pass.

- [x] Add at least some form of tests.

- [x] Add a way to fetch/retrieve run results without `--follow`.

- [x] Add a way to list available job parameters.

- ~Decide what information should be attached/configured with a resource, and how.~

  Things like the root run directory, the submitter, ...  Adding duplicate attribs to each resource seems problematic.
   
    *Edit*: This hasn't been resolved, but it's part of a larger problem of associating information with resources without needing to add an `attrib`, and I think it can wait for a follow-up PR.


kyleam:
From @yarikoptic's #314:
> * then going further, may be "resource" should be a part of the job spec, but otherwise a subject for the aforementioned overrides in the same way?  This way we might be able to support specification of "execute on multiple resources" (where specification would be a list of items, `-j resource hydra discovery`) after we figure out also how to introduce parametrization of the job (run the same command template with different substitutions, e.g. sweeping through subjects).

kyleam:
From my POV the current state is an OK stopping point for this PR.

kyleam:
Pushed update to resolve latest conflicts with master.

PR #332, opened 2018-11-12
Adjust ABCs to be Py3-compatible

kyleam:
The key patch is 50deea04 (BF: Declare abc.ABCMeta in py3-compatible way).

### DIFF
--- a/niceman/distributions/base.py
+++ b/niceman/distributions/base.py
@@ -15,6 +15,7 @@ import collections
 import yaml
 
 from importlib import import_module
+from six import add_metaclass
 from six import viewvalues
 
 from niceman.utils import attrib
@@ -68,11 +69,10 @@ class Package(SpecObject):
         return tuple(getattr(self, a) for a in self._cmp_fields)
 
 
+@add_metaclass(abc.ABCMeta)
 @attr.s
 class Distribution(SpecObject):
     """Base class for distributions"""
-
-    __metaclass__ = abc.ABCMeta
 
     # Actually might want/need to go away since somewhat duplicates the class
     # name and looks awkward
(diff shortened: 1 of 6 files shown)


niceman/distributions/base.py
```diff
@@ -194,7 +194,7 @@ def _init(self):
 
     @abc.abstractmethod
     def identify_distributions(self, files):
-        raise NotImplementedError()
+        return
```
yarikoptic: may be indeed pointless, but how plain `return` is better (if it is not better - why bother/make it worse)?
at least ` raise NotImplementedError` serves as some kind of an additional guard if `abc` and metaclass stuff doesn't work.

kyleam: > may be indeed pointless, 

It is.

> but how plain `return` is better (if it is not better - why bother/make it worse)?

`return` or `pass` is required to avoid a syntax error unless there's a docstring.  Depending on taste (consistency versus necessity), they could be dropped if there is a docstring.  But they are just simple stand-ins.  Doing anything more involved is likely to confuse a reader because (1) it's unconventional and (2) it suggests that the code is reachable because otherwise why bother.

> at least ` raise NotImplementedError` serves as some kind of an additional guard if `abc` and metaclass stuff doesn't work.

That instantiation with abstract methods should raise a `TypeError` is by design (PEP 3119).

yarikoptic: > That instantiation with abstract methods should raise a TypeError is by design (PEP 3119).

that was my point that "should" might not work, as e.g. you have discovered that `__metaclass__ = abc.ABCMeta` now might not be the right way to go.  With `pass` or `return` you would just miss that such protection stopped working for one reason or another.  Explicitly raised exception wouldn't.   But I am not going to battle it further since hopefully one way or another it wouldn't make difference in the future.

PR #335, opened 2018-11-15
TST: Don't call pytest fixtures directly

### LINKED ISSUE #333: Travis tests failing with pytest 4.0.0
The BOTO_CONFIG=/tmp/nowhere test run is failing on master (e.g., see [job 455625370]).  I haven't looked at it in detail, but the failures seem to be related to direct fixture calls being unsupported in the latest pytest (v4.0.0).  For example

```

    def test_distribution_sub():
>       (p1, p1v10, p1v11, p1ai, p1aa, p1v11ai, p2) = setup_packages()
E       _pytest.warning_types.RemovedInPytest4Warning: Fixture "setup_packages" called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.

niceman/distributions/tests/test_redhat.py:159: RemovedInPytest4Warning
```

[job 455625370]: https://travis-ci.org/ReproNim/niceman/jobs/455625370

kyleam:
This has been deprecated since pytest 3.7 [0] and leads to an
error (by default [1]) starting with pytest 4.

Fixes #333.

[0]: https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly
[1]: https://docs.pytest.org/en/latest/changelog.html#removals

### DIFF
--- a/niceman/distributions/tests/test_debian.py
+++ b/niceman/distributions/tests/test_debian.py
@@ -227,8 +227,8 @@ def test_package_satisfies(setup_packages):
     assert p1v11ai.satisfies(p1v11)
 
 @pytest.fixture
-def setup_distributions():
-    (p1, p1v10, p1v11, p1ai, p1aa, p1v11ai, p2) = setup_packages()
+def setup_distributions(setup_packages):
+    (p1, p1v10, p1v11, p1ai, p1aa, p1v11ai, p2) = setup_packages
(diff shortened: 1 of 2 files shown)


yarikoptic:
Eh,  even less of python left in pytest ;-) only magic is left

PR #336, opened 2018-11-16
MNT: Avoid use of deprecated inspect.getargspec()

kyleam:
inspect.getargspec() was deprecated in Python 3, briefly removed
entirely, and then restored in 37dc2b2883 (Issue #25486: Resurrect
inspect.getargspec in 3.6. Backout a565aad5d6e1., 2016-01-11).

Much if not all of the code we're updating came from DataLad via 3rd.
DataLad has since dealt with this issue.  Add our own compatibility
function because merging in the relevant DataLad bits would require a
lot of work at this point.


yarikoptic:
thank you!

Issue #292, opened 2018-09-11
provide connection info in debug mode for ssh


yarikoptic:
ATM, suffering from #286 wanted to just ssh directly (which is I guess is possible) but login -l debug does not give any hint on what I should ssh into

```
$> niceman login -l 1 test-ec2-west-medium2     
2018-09-11 16:40:56,583 [DEBUG  ] Unknown host key! 
2018-09-11 16:40:56,588 [DEBUG  ] Trying ssh-agent key da535f52cdb011f86e794db310b0b5c6 
2018-09-11 16:40:56,789 [DEBUG  ] ssh-agent key da535f52cdb011f86e794db310b0b5c6 failed! 
2018-09-11 16:40:56,991 [DEBUG  ] Opening TTY connection via SSH. 
Welcome to Ubuntu 16.04.5 LTS (GNU/Linux 4.4.0-1065-aws x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  Get cloud support with Ubuntu Advantage Cloud Guest:
    http://www.ubuntu.com/business/services/cloud

13 packages can be updated.
5 updates are security updates.

New release '18.04.1 LTS' available.
Run 'do-release-upgrade' to upgrade to it.


Last login: Tue Sep 11 20:40:46 2018 from 194.59.251.100
```
would be nice if it provided ip:port and point to the key file  used

yarikoptic:
@mjtravers : issue remains
```shell
(git)hopa:~/proj/repronim/niceman[master]git
$> niceman login -l 1 discovery              
2018-10-26 11:59:03,584 [DEBUG  ] Opening TTY connection via SSH. 
[yhalchen@discovery ~]$ exit
Exited terminal session.
2018-10-26 11:59:10,750 [Level 5] Exiting 
```

yarikoptic:
Fixed by #337

PR #337, opened 2018-11-24
ENH: Added debug logging to display SSH connection parameters

mjtravers:
Addresses issue #292

### DIFF
--- a/niceman/resource/ssh.py
+++ b/niceman/resource/ssh.py
@@ -67,16 +67,9 @@ class SSH(Resource):
             }
         )
 
-        if self.port:
-            port = self.port
-        else:
-            port = 22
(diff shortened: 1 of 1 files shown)


niceman/resource/ssh.py
```diff
@@ -66,6 +66,18 @@ def connect(self, password=None):
...
+            port = self.port
+        else:
+            port = 22
+        lgr.debug("Connecting to remote at {}@{}:{}".format(self.user,
+            self.host, port))
+        if self.key_filename:
+            lgr.debug("Using SSH keyfile {}".format(self.key_filename))
+        elif password:
+            lgr.debug("Using provided password for SSH authentication")
+
```
yarikoptic: May be better to wrap it into that single entry above?

Issue #317, opened 2018-10-26
Fix(?) stopping of the ec2 resource


yarikoptic:
Originally reported in https://github.com/ReproNim/niceman/issues/282#issuecomment-419484977

```shell
$> niceman stop test-ec2-west
2018-09-06 14:29:24,440 [ERROR  ] 'NoneType' object has no attribute 'stop' [aws_ec2.py:stop:201] (AttributeError) 
```

yarikoptic:
@mjtravers for what I know -- it might still be running!
```shell
$> niceman ls          
RESOURCE NAME        TYPE                 ID                  STATUS    
-------------        ----                 --                  ------    
debian-testing-conda docker-container     e41aa8f26a77ec2dacd running   
discovery            ssh                  88a9e369-7780-489e- N/A       
localshell           shell                thebest             None      
smaug                ssh                  8b253827-9133-4a3b- N/A       
smaug2               ssh                  16412f12-8d99-48e7- N/A       
test                 docker-container     c6f0d0afcdc48a93c10 running   
test-ec2-west        aws-ec2              i-0368a48188f646b34 running   
test-ec2-west-medium aws-ec2              i-08ec5e3cc0af73c60 running   
(dev) 1 17218.....................................:Mon 05 Nov 2018 11:14:58 PM EST:.
(git)hopa:~/proj/repronim/niceman[master]git
$> niceman stop test-ec2-west
2018-11-05 23:15:19,151 [ERROR  ] 'NoneType' object has no attribute 'stop' [aws_ec2.py:stop:206] (AttributeError) 
```

PR #338, opened 2018-11-25
BF: Updated Resource code for 'start' and 'stop' methods to handle re…

mjtravers:
…source connection properly and update resource inventory status appropriately

1. Fixed ordering of resource.connect() and resource.stop/start() method calls.
2. Now updates resource status in inventory.yml when a resource is started/stopped
3. Handles resources that do not have stopped states. (i.e. singularity, shell, and ssh)

Covers issue #317

### DIFF
--- a/niceman/interface/start.py
+++ b/niceman/interface/start.py
@@ -42,7 +42,7 @@ class Start(Interface):
 
     @staticmethod
     def __call__(resref, resref_type="auto"):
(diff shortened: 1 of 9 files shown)


kyleam:
> aside from the newly failing tests (assuming those are actually related)

Looks like those might've been connection issues.  I've restarted them, and they've passed.

PR #339, opened 2018-11-27
support for command-line interface exit statuses

chaselgrove:


niceman/cmdline/main.py
```diff
@@ -267,7 +267,9 @@ def main(args=None):
             # print('%s (%s)' % (exc_str(exc), exc.__class__.__name__))
             lgr.error('%s (%s)' % (exc_str(exc), exc.__class__.__name__))
             sys.exit(1)
-    if hasattr(cmdlineargs, 'result_renderer'):
-        cmdlineargs.result_renderer(ret)
+    if ret:
+        if hasattr(cmdlineargs, 'result_renderer'):
+            cmdlineargs.result_renderer(ret)
+        return ret.status
```
yarikoptic: not to make it into a "bikeshed" but it could be quickly improved into something more generic/flexible than relying on specific return type.
0. do not condition on `if ret` -- `None` as a result is quite valid and renderer might/should be happy to "render" it then.
1. make `result_renderer` return exit code to exit with
2. if it returns non-0, exit with that code

chaselgrove:
I would rather have decoupled the display from the determination of the status of the operation, but try this instead.

yarikoptic:
If this minimal change is what you need, good by me ;)

PR #341, opened 2018-12-10
 TST: execute: Fix skipif condition

kyleam:
```
There are two issues with this condition: (1) it was intended to say
"not in" rather than "in" and (2) an ExternalVersions instance's
__contains__(item) always returns False if __getitem__(item) hasn't
been called previously.  The combination of no. 1 and no. 2 means the
test would never be skipped.

No. 2 has been fixed as of DataLad's e71a57aa3 (BF: external_versions:
Make __contains__ work with non-primed items, 2018-12-03) and has been
incorporated via 3rd.

Update the skipif condition accordingly.
```

### DIFF
--- a/niceman/interface/tests/test_execute.py
+++ b/niceman/interface/tests/test_execute.py
@@ -123,7 +123,7 @@ def test_trace_docker(docker_container, trace_info):
 @pytest.mark.skipif(os.environ.get('NICEMAN_TESTS_NONETWORK'),
                     reason="No network")
 @pytest.mark.skipif("cmd:apt-cache" in external_versions,
-                    reason="Not apt-cache")
+                    reason="No apt-cache")
 def test_trace_local(trace_info):
     with patch("niceman.resource.ResourceManager._get_inventory") as get_inv:
         config = {"status": "running",
--- a/niceman/interface/tests/test_execute.py
+++ b/niceman/interface/tests/test_execute.py
@@ -122,7 +122,7 @@ def test_trace_docker(docker_container, trace_info):
 # TypeError under Python 2.
 @pytest.mark.skipif(os.environ.get('NICEMAN_TESTS_NONETWORK'),
                     reason="No network")
-@pytest.mark.skipif("cmd:apt-cache" in external_versions,
+@pytest.mark.skipif("cmd:apt-cache" not in external_versions,
                     reason="No apt-cache")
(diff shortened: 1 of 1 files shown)


niceman/interface/tests/test_execute.py
```diff
@@ -122,8 +122,8 @@ def test_trace_docker(docker_container, trace_info):
 # TypeError under Python 2.
 @pytest.mark.skipif(os.environ.get('NICEMAN_TESTS_NONETWORK'),
                     reason="No network")
-@pytest.mark.skipif("cmd:apt-cache" in external_versions,
-                    reason="Not apt-cache")
+@pytest.mark.skipif("cmd:apt-cache" not in external_versions,
+                    reason="No apt-cache")
```
yarikoptic: good catch! 
although functionally identical ATM (see below) I wonder if more correct form would be `not external_versions["cmd:apt-cache"]`?

```shell
$ python -c 'from niceman.support.external_versions import external_versions as ev; print("cmd:apt-cache" in ev)'                                 

2018-12-10 10:11:24,024 [ERROR  ] Failed to start ['apt-cache', '-v']'': [Errno 2] No such file or directory [subprocess.py:_execute_child:1025] 
False
(niceman-dev) [yhalchen@discovery7 niceman]$ 
$ python -c 'from niceman.support.external_versions import external_versions as ev; print(ev["cmd:apt-cache"])'                                      
2018-12-10 10:11:43,191 [ERROR  ] Failed to start ['apt-cache', '-v']'': [Errno 2] No such file or directory [subprocess.py:_execute_child:1025] 
None
```

kyleam: Sorry, I'm not following why the `__getitem__` variant would be more correct.

yarikoptic: I thought that it would make sense to eventually return `None` if nothing is installed, while `__contains__` to return True if we already "know" the version (which could be None).  But now I remembered that you fixed in DataLad recently and brought here the:
```shell
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -225,7 +225,7 @@ class ExternalVersions(object):
         return self._versions.keys()
 
     def __contains__(self, item):
-        return item in self._versions
+        return bool(self[item])
```
so I guess we will stick to this semantic, sorry for the noise

PR #345, opened 2019-01-02
BF: do not blow `ls` when status is None

yarikoptic:
also state "available" for localshell

### DIFF
--- a/niceman/interface/ls.py
+++ b/niceman/interface/ls.py
@@ -55,7 +55,7 @@ class Ls(Interface):
     @staticmethod
     def __call__(verbose=False, refresh=False):
         id_length = 19  # todo: make it possible to output them long
-        template = '{:<20} {:<20} {:<%(id_length)s} {:<10}' % locals()
+        template = '{:<20} {:<20} {:<%(id_length)s} {!s:<10}' % locals()
         ui.message(template.format('RESOURCE NAME', 'TYPE', 'ID', 'STATUS'))
         ui.message(template.format('-------------', '----', '--', '------'))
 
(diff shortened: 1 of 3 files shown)


yarikoptic:
> AFAICT this won't have any effect. Isn't something more needed on top? Perhaps

why?  I think it did have desired effect for me - that it listed it as "available" whenever before was just None (default `default`)

kyleam:
> why?

It seems like, if "available" is the default, it should be what is listed when a shell is created with no other options.  But that's not what I see

```
% niceman ls
RESOURCE NAME        TYPE                 ID                  STATUS    
-------------        ----                 --                  ------    
% niceman create -t shell lsh
2019-01-02 15:58:03,998 [INFO   ] Created the environment lsh 
% niceman ls
RESOURCE NAME        TYPE                 ID                  STATUS    
-------------        ----                 --                  ------    
lsh                  shell                18e2c36c-0ed1-11e9- N/A 
``` 

With the patch that I suggested:

```
% niceman ls                 
RESOURCE NAME        TYPE                 ID                  STATUS    
-------------        ----                 --                  ------    
% niceman create -t shell lsh
2019-01-02 16:00:10,990 [INFO   ] Created the environment lsh 
% niceman ls                 
RESOURCE NAME        TYPE                 ID                  STATUS    
-------------        ----                 --                  ------    
lsh                  shell                64944182-0ed1-11e9- available 
```

> I think it did have desired effect for me - that it listed it as "available" whenever before was just None (default `default`)

As I mentioned above, I still don't know how to get in the `None` state you report.  But if "available" is the default, I think it should be shown instead of "N/A" when I create a shell with `niceman create -t shell lsh`.

Issue #340, opened 2018-11-29
Make possible to establish ssh sessions via password authentication


yarikoptic:
- [x] if Fabric could ask for credentials (currently we just blow with `PasswordRequiredException`) or we would need to catch/ask/provide
- [ ] establish the controlpath for repeated relogins without authentication

yarikoptic:
here is the helper we have in DataLad
https://github.com/datalad/datalad/blob/master/datalad/support/sshconnector.py#L67

yarikoptic:
- [ ] try Mux-transport-for-client branch of radssh which is now submitted as a PR https://github.com/paramiko/paramiko/pull/1341/files  so we could pass `controlpath`

mjtravers:
The first go around with using the Mux-transport-for-client paramiko branch showed that the new code only connects to existing ControlMaster proxy connections and does not create them.

It was also found that the code depends on a ControlMaster running at least OpenSSH v7.4. Here is the exception raised when trying to connect to my locally setup control master using the ssh command line tool which has a version of OpenSSH_7.2p2:
```
Unable to connect to ControlPath '/home/matt/.ssh/ssh-root@localhost:32768' - Mux proxy
request rejected (Requires OpenSSH 7.4 or greater) - got unexpected reply: 0x80000003 
paramiko.Message('\x00\x00\x00\x02\x00\x00\x00\x13unsupported request')
```

mjtravers:
Was able to connect without a password via the ControlMaster process after setting up a openssh-client version >= v7.4 using the multiplexer branch of paramiko

yarikoptic:
so, that PR relies on this new version in 7.4:
https://www.openssh.com/txt/release-7.4
```
 * ssh(1): Add a proxy multiplexing mode to ssh(1) inspired by the
   version in PuTTY by Simon Tatham. This allows a multiplexing
   client to communicate with the master process using a subset of
   the SSH packet and channels protocol over a Unix-domain socket,
   with the main process acting as a proxy that translates channel
   IDs, etc.  This allows multiplexing mode to run on systems that
   lack file- descriptor passing (used by current multiplexing
   code) and potentially, in conjunction with Unix-domain socket
   forwarding, with the client and multiplexing master process on
   different machines. Multiplexing proxy mode may be invoked using
   "ssh -O proxy ..."
```
from reading which it suggests that it was added primarily for "systems that lack file- descriptor passing (used by current multiplexing code) "... I wonder which systems those are? ;)
[...]

PR #357, opened 2019-01-05
BF: Added password prompt for SSH login

mjtravers:
First part of #340 

Will keep it separate from the second part so the password prompt can get into the impending point release.

### DIFF
--- a/niceman/resource/ssh.py
+++ b/niceman/resource/ssh.py
@@ -84,9 +84,6 @@ class SSH(Resource):
         try:
             self._connection.open()
         except AuthenticationException:
-            # Give the user a chance to enter the password again.
-            if password:
-                print('Password failed.')
             password = getpass.getpass()
             self._connection = Connection(
                 self.host,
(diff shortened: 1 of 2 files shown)


niceman/resource/ssh.py
```diff
@@ -79,7 +81,20 @@ def connect(self, password=None):
                   self._connection.port,  # Fabric defaults to 22.
                   auth)
 
-        self._connection.open()
+        try:
+            self._connection.open()
+        except AuthenticationException:
+            # Give the user a chance to enter the password again.
```
kyleam: I was a bit confused by this comment, particularly "again".  AFAICT the user only has one chance to enter a password.  The only time there would already be a password is if the `password` parameter was specified, but that is only relevant for testing, right?  

So I think "again" should be dropped, and it should be made clearer that the `if password` bit is currently for testing purposes, either by tweaking the message  or adding a comment.

mjtravers: Agreed. Neither the comment nor the if statement are necessary. Will remove them both.

PR #361, opened 2019-01-14
NF: resource: Support partial matches on ID

### LINKED ISSUE #307: ResourceManager should try to match short resource IDs

kyleam:
Resource IDs are often long and unwieldy to type.  Make it possible to
match them by giving a unique prefix (à la git hexsha matching).
Don't bother with supporting partial matching for names because a name
should already serve as a short label (like a branch name, to continue
the git analogy).

Closes #307.
Supersedes #330.

### DIFF
--- a/niceman/resource/base.py
+++ b/niceman/resource/base.py
@@ -187,12 +187,9 @@ class ResourceManager(object):
 
         results_name = None
         results_id = None
-        if resref_type == "auto":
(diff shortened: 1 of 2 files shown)


yarikoptic:
coolio, thanks!

Issue #359, opened 2019-01-08
NICEMAN -> ReproMan


yarikoptic:
Time has come again for a rename. And we should do it asap.  But I wondered if to simplify we should just merge all PRs which are in good state (not failing tests) first, even though they are WiP, and then proceed from there?

TODOs:
- [x] codebase renames (#362) 
- [ ] Add to README information about prior name
- [x] rename repo
- [ ] rename pypi/debian packages
- [ ] add instructions for config renames etc

kyleam:
I'll take care of rebasing #330 and #331 after the rename.  I think #357 and #358 are good to be merged.  

So that leaves deciding about #273.

PR #362, opened 2019-01-17
RF: NICEMAN -> ReproMan

yarikoptic:
Ref: #359


yarikoptic:
So let's make it happen

PR #364, opened 2019-01-31
ENH: inventory updates during resource create, added paramiko debug t…

### LINKED ISSUE #346: enable paramiko/fabric logging (into our handlers) if our debug level < DEBUG
In the light of #344 it is hard to figure out what is going on in fabric, so we need an easy way to enable its debug output.  I guess the easiest is just to direct their logger into our handlers

### LINKED ISSUE #348: Creating a new resource should initiate inventory record (on disk) asap
ATM it seems to be waiting e.g. for aws-ec2 to instantiate/startup, but that might fail and then we would have those orphans around in the cloud.  So the record should be created/saved asap and then only status etc get updated

Related #298 and #342

mjtravers:
…o logging

Fixes #346 
Fixes #348 

This PR:

1. Adds paramiko debugging to logging when < DEBUG
2. Allows resource.create() to save the inventory data as needed during the process
3. Fixes a bug found in aws-ec2 resource where key_filename was put in a list

### DIFF
--- a/reproman/resource/aws_ec2.py
+++ b/reproman/resource/aws_ec2.py
@@ -111,7 +111,7 @@ class AwsEc2(Resource):
         """
         Create an EC2 instance.
 
-        Returns
+        Yields
         -------
         dict : config and state parameters to capture in the inventory file
         """
--- a/reproman/resource/aws_ec2.py
+++ b/reproman/resource/aws_ec2.py
@@ -162,6 +162,15 @@ class AwsEc2(Resource):
         # Save the EC2 Instance object.
         self._ec2_instance = self._ec2_resource.Instance(instances[0].id)
         self.id = self._ec2_instance.instance_id
+        self.status = self._ec2_instance.state['Name']
+
+        # Send initial info back to be saved in inventory file.
+        yield {
+            'id': self.id,
+            'status': self.status,
+            'key_name': self.key_name,
+            'key_filename': self.key_filename
+        }
 
         lgr.info("Waiting for EC2 instance %s to start running...", self.id)
         self._ec2_instance.wait_until_running(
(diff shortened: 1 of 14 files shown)


kyleam:
Haven't had a chance to review the rest of the PR but...

> Fixes a bug found in aws-ec2 resource where key_filename was put in a list

Do you actually run into an error due to key_filename being in a list?

Putting it in a list was intentionally done to work around an issue with fabric.  See the commit message of 3807f1287 (BF: aws_ec2: Pass key_filename as a list, 2019-01-04).

mjtravers:
Yes, I was getting an error when the key_filename is set to None:

```
$ reproman login matt-test
2019-01-31 16:37:03,032 [ERROR  ] 'NoneType' object has no attribute 'endswith' [client.py:_key_from_filepath:564] (AttributeError) 
```

Okay, I see the issue is how paramiko client.py was handling the [None] value for key_filename:

```
       if key_filename is None:
            key_filenames = []
        elif isinstance(key_filename, string_types):
            key_filenames = [key_filename]
        else:
            key_filenames = key_filename
```

I put a line in the SSH.connect() method to convert the key_filename to a list as a fix until your PR gets merged into Fabric. I like it there so additional calls (if there ever are any) to SSH() don't have to remember to list the key_filename quirk.

kyleam:
> Yes, I was getting an error when the key_filename is set to None: [...]

I see.  Thanks for the details and the fix.



> [...] until your PR gets merged into Fabric. I like it there so additional calls (if there ever are any) to SSH() don't have to remember to list the key_filename quirk.

Agreed.  It's meant as a temporary workaround.  I should have left a comment indicating this.  Thanks for doing so in your fix.

mjtravers:
Awesome. Thanks for the testing fix.

Yeah, I forgot about the documentation and need to add a test. I will fix/add those and commit.

mjtravers:
No problem, let me update the docstring with more detail

kyleam:
>  let me update the docstring with more detail

great, thanks

Issue #367, opened 2019-02-07
Multiple (duplicate?) distributions in trace files


chaselgrove:
In the traces at http://www.onerussian.com/tmp/niceman-traces-1.tgz, there are multiple venv distributions (with the same path and venv_version) and multiple debian distributions.

kyleam:
> duplicate?

Based on diffing the individual distributions in those files, there seems to be some differences (e.g., whether there are files associated with a package), though I've also found some that are identical too.

The (or at least _a_) core issue here is that in 2cf46df0 we punted on merging the multiple iterations:

```
ENH: iterate over tracers until nothing new is discovered

Note: this is not yet complete since result would duplicate distributions
and possibly packages within them, since every pass is independent from
previous ones.

On a joint meeting of two us we decided that we should leave the
resolution to the "unification" step which would pull those together
since "unification" was envisisioned, and it would come handy anyways
for other usecases (e.g. joining multiple specs)
```

AFAIK this hasn't been revisited at all.

Here's what seems to be going on.  $ENV/bin/python was given as a path to trace.  This isn't recognized as a virtualenv file, so it is added to the unknown files [*].  On the next iteration, this passes through DebTracer as an unknown file and so on until we get back to the virtualenv tracer.  It detects that the $ENV/bin/python file is under the $ENV virtualenv, so it returns that distribution (this time _without_ paths that were claimed by virtualenv in the previous iteration).  I think any Debian or Git repo tracing that was triggered by the initial virtualenv tracing will be triggered again, leading to the multiple Debian and Git distributions.

So, it's an interaction between us not merging distributions after multiple retracing iterations and the virtualenv tracer gettting confusing by $ENV/bin/python.  Based on a quick test, this patch would fix the virutalenv confusion:

```diff
diff --git a/reproman/distributions/venv.py b/reproman/distributions/venv.py
index f66b65e54..242c83ad0 100644
--- a/reproman/distributions/venv.py
+++ b/reproman/distributions/venv.py
@@ -160,7 +160,9 @@ def identify_distributions(self, files):
             # system wide installation of python
             for path in unknown_files.copy():
                 if is_subpath(path, venv_path) and op.islink(path):
-                    unknown_files.add(op.realpath(path))
+                    rpath = op.realpath(path)
+                    if not is_subpath(rpath, venv_path):
+                        unknown_files.add(rpath)
                     unknown_files.remove(path)
 
             packages = []
```
---

[...]

PR #371, opened 2019-02-13
BF: virtualenv: Don't return internal symlinks as unknown paths

kyleam:
We resolve virtualenv symlinks and report them as unknown files
because links in the virtual environment may point to system
packages ($ENV/lib/python3.5/types.py -> /usr/lib/python3.5/types.py),
and we want to detect those resolved paths with another
tracer (usually DebTracer) on the next `retrace` iteration.

However, a symlink may point to another file under the same virtual
environment, as is the case with "$ENV/bin/python -> $ENV/bin/pythonX"
links.  Returning the resolved path of an internal link as an unknown
file to `retrace` will result in a duplicate [*] virtualenv
distribution because the path will reach the virtualenv tracer on the
next iteration.  (This can in turn lead to duplicate Git distributions
if any of the virtualenv packages are editable.)

Re: #367

[*]: Or at least near-duplicate: the second distribution will lack any
     files that were assigned to virtualenv packages during the first
     pass.

### DIFF
--- a/reproman/distributions/tests/test_venv.py
+++ b/reproman/distributions/tests/test_venv.py
@@ -17,6 +17,7 @@ import logging
 
 from reproman.cmd import Runner
 from reproman.utils import chpwd
+from reproman.utils import on_linux
 from reproman.tests.utils import create_pymodule
 from reproman.tests.utils import skip_if_no_network, assert_is_subset_recur
 from reproman.tests.utils import swallow_logs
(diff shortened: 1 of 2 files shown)


kyleam:
Looks like the original [realpath] call in venv.py wasn't covered.  I'll adjust the tests.

[realpath]: https://codecov.io/gh/ReproNim/reproman/src/b72353753c4369745993d101dece02f6172006b6/reproman/distributions/venv.py#L163

PR #374, opened 2019-02-14
Drop Python 2 support

### LINKED ISSUE #369: RF:  dropping python2 support
The verdict was made to drop python2 support in reproman since anyways it is mostly a command line tool anyways, and we interact with the remote sessions from local environment so do not care about what version of python is on remote end. So, possibly to take advantage of python3-only features we are dropping python2 **now** (well, as soon as PR comes in ripping all compatibility shimming).  I think code in PRs shouldn't conflict much if at all.
The question remaining - what minimal version of python3 we would like to support? I guess it would depend on what new features we would like to make use of, and for that we better familiarize us with what is available across python3 versions.   Please chime in posting your favorite (I can only google up some) summaries for "what's new in python 3.x"

kyleam:
Here's an initial wave of changes to strip out py2-specific bits.  ~There are definitely some things left to do (e.g., adjust CONTRIBUTING.md), but I'm hoping the tests pass.~

Closes #369.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,9 @@ matrix:
     env:
     - REPROMAN_LOGLEVEL=2
     - REPROMAN_LOGTARGET=/dev/null
-  - python: 3.5
+  - python: 3.6
+    # Note: This this no network run appears to hang or otherwise
+    # behave oddly on 3.5.
     env:
     - REPROMAN_TESTS_NONETWORK=1
     # must operate nicely with those env variables set
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,30 +12,25 @@ cache:
 
 matrix:
   include:
-  - python: 2.7
-  # no loop dev support on travis yet :-/ https://github.com/travis-ci/travis-ci/issues/2700
-  #- python: 2.7
+  - python: 3.5
     env:
     - REPROMAN_TESTS_SSH=1
-  - python: 2.7
+  - python: 3.5
     # By default no logs will be output. This one is to test with log output at INFO level
     env:
     - REPROMAN_LOGLEVEL=INFO
-  - python: 2.7
+  - python: 3.5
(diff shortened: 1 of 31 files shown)


kyleam:
In the initial push, the [py3.5/"no network" run][0] failed.  Looking at the two errors, it seems like they are two virtualenv tests that **should have been skipped** based on the no network condition.  I can't reproduce that on my system that has py3.5, though.  And oddly these get skipped as intended [when we use py3.6][1].

So we're all green with the latest push.  I'm not feeling too motivated to dig into those 3.5 "not skipped" oddities, unless someone has a good idea of what might be going on.

[0]: https://travis-ci.org/ReproNim/reproman/builds/493388571
[1]: https://travis-ci.org/ReproNim/reproman/builds/493468815

kyleam:
> rerase - so six was providing some shim ?

Here's the `reraise` code from `six`:

```python
if PY3:
    exec_ = getattr(moves.builtins, "exec")

    def reraise(tp, value, tb=None):
        if value is None:
            value = tp()
        if value.__traceback__ is not tb:
            raise value.with_traceback(tb)
        raise value

else:
    def exec_(_code_, _globs_=None, _locs_=None):
        """Execute code in a namespace."""
        if _globs_ is None:
            frame = sys._getframe(1)
            _globs_ = frame.f_globals
            if _locs_ is None:
                _locs_ = frame.f_locals
            del frame
        elif _locs_ is None:
            _locs_ = _globs_
        exec("""exec _code_ in _globs_, _locs_""")

    exec_("""def reraise(tp, value, tb=None):
    raise tp, value, tb
""")
```

kyleam:
Looking at that made me realize that I could've just used the bound `with_traceback`.  Updated.

kyleam:
@yarikoptic Thanks for taking a look.  I'll merge once the tests complete.

PR #376, opened 2019-02-19
Add new module for defining test skip conditions

kyleam:
The goal of this PR is to make it easier to add custom skip functions.  (I started looking at this when working on adding a custom "skip if condor not installed" function.)  I've added a new module tests/skip.py that defines two entry points, `skipif` and `mark`.  These provide access to all the conditions.  For example, for the no network case, it'd look something like

```python
@mark.skipif_no_network
def test_some_network():
    pass
```

or

```python
def test_some_thing():
    [...]
    if [some on-the-fly reason to skip]:
        skipif.no_network()
```

Those `skipif.*` and `mark.skipif_*` functions above are defined based on a single function, which is  member of the list of functions in `skip.CONDITION_FNS`:

```python
def no_network():
    return ("no network settings",
            os.environ.get('REPROMAN_TESTS_NONETWORK'))
```

There are some clean-up and preparatory commits aside from the commits that define the interface above.  The most notable ones add SVN and Singularity to `external_versions`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,7 @@ script:
   - python setup.py build
   - if [ ! -z "$REPROMAN_TESTS_NONETWORK" ]; then sudo route add -net 0.0.0.0 netmask 0.0.0.0 dev lo; fi
   # - $NOSE_WRAPPER `which nosetests` -s -v --with-doctest --doctest-tests --with-cov --cover-package reproman --logging-level=INFO
-  - coverage run `which py.test` -s --integration reproman
+  - coverage run `which py.test` -s -rs --integration reproman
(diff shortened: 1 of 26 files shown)


kyleam:
I've incorporated this into the nf-run branch because I wanted to set up the condor testing.  (I'll rewind that merge if this PR gets rejected.)

While doing this, I noticed another nice advantage of making our skip decorators pytest marks.  We can use it with `pytest.param` within `parametrize`:

https://github.com/ReproNim/reproman/blob/1a7b57d5d6ac23b23f8e10f16867cb0ca623c2bc/reproman/support/jobs/tests/test_orchestrators.py#L84-L92

When condor isn't available, that skips the tests that involve condor, but runs the non-condor combinations.

kyleam:
From call: @yarikoptic suggests to use `on_windows` for condition name.  Will do.

PR #378, opened 2019-02-25
retrace: Map pyc files back to their source

kyleam:
When pyc files come in, convert them to the non-cached name because this is what can be associated with packages (both virtualenv and system).  The approach as of this PR's 4ca6ff47f isn't sufficient because it covers only the virtualenv package case.  If the pyc file is part of the system's python distribution, it will pass it back out as an unknown file, and DebTracer won't detect it.

In order to deal with this, we need to do two things to unknown files that are subpaths of the virtualenv: (1) apply the pyc -> py mapping over them and (2) widen the symlink resolution to handle cases where the leading path has a directory symlink (e.g.,  `$VENV/lib/python3.5/encodings -> /usr/lib/python3.5/encodings`).

### DIFF
--- a/reproman/utils.py
+++ b/reproman/utils.py
@@ -14,6 +14,7 @@ import builtins
 from shlex import quote as shlex_quote
 import time
 
+import os.path as op
 from os.path import curdir, basename, exists, realpath, islink, join as opj, isabs, normpath, expandvars, expanduser, abspath
 from urllib.parse import quote as urlquote, unquote as urlunquote, urlsplit
 
@@ -1381,6 +1382,35 @@ def merge_dicts(ds):
     for d in ds:
         merged.update(d)
     return merged
+
+
+def pycache_source(path):
+    """Map a pycache path to the original path.
+
+    Parameters
+    ----------
+    path : str
+        A Python cache file.
+
+    Returns
+    -------
+    Path of cached Python file (str) or None if `path` doesn't look like a
+    cache file.
+    """
+    if not path.endswith(".pyc"):
+        lgr.debug("Path does not look like a pyc file: %s", path)
+        return
+
+    if "__pycache__" not in path:  # py2
+        pyfile = path[:-1]
+    else:
+        # It should be a py3-style path, e.g., "__pycache__/f.cpython-35.pyc".
+        leading, base = op.split(path)
+        name, *_ = base.rsplit(".", 2)
+        pyfile = op.join(leading[:-len("__pycache__")], name + ".py")
+    lgr.debug("Converted pycache file %s to source file %s",
+              path, pyfile)
+    return pyfile
 
 
 lgr.log(5, "Done importing reproman.utils")
--- a/reproman/utils.py
+++ b/reproman/utils.py
@@ -1397,16 +1397,17 @@ def pycache_source(path):
     Path of cached Python file (str) or None if `path` doesn't look like a
     cache file.
     """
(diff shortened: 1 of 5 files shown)


kyleam:
> In order to deal with this, we need to do two things to unknown files that are subpaths of the virtualenv: (1) apply the pyc -> py mapping over them and (2) widen the symlink resolution to handle cases where the leading path has a directory symlink (e.g., `$VENV/lib/python3.5/encodings -> /usr/lib/python3.5/encodings`).

Actually doing that wound't be sufficient because the DebTracer wouldn't know how to handle the pyc files that get passed back to it.  So instead of doing no. 1 in all tracers, I've done it once in retrace.py.

reproman/utils.py
```diff
@@ -1383,4 +1384,33 @@ def merge_dicts(ds):
...
+    ----------
+    path : str
+        A Python cache file.
+
+    Returns
+    -------
+    Path of cached Python file (str) or None if `path` doesn't look like a
+    cache file.
+    """
+    if not path.endswith(".pyc"):
```
yarikoptic: so we will not bother with `.pyo`  (from `python -O`) since they are unlikely to encounter?

If you decide to add support for them, note that with py3 they seems to not acquire `.pyo` extension like with python2 but rather get additional suffix to the filename.  Here is a result of little local exploration of importing a module with and without `-O`
```shell
$> find hello 
hello
hello/__init__.pyc
hello/__init__.pyo
hello/__pycache__
hello/__pycache__/__init__.cpython-37.opt-1.pyc
hello/__pycache__/__init__.cpython-37.pyc
hello/__init__.py
```
I ran `locate` on my laptop for .opt-1.pyc and found none, but for .pyo did find quite a few hits in virtual environments around the projects where I am not sure if I ever ran them explicitly with `python -O`, e.g. even in our `simple_workflow`
```
/home/yoh/proj/repronim/reprozip/reprounzip-docker/reprounzip/__init__.pyo
/home/yoh/proj/repronim/reprozip/reprounzip-docker/reprounzip/unpackers/__init__.pyo
/home/yoh/proj/repronim/simple_workflow/simple_workflow/miniconda/envs/bh_demo/lib/python2.7/BaseHTTPServer.pyo
/home/yoh/proj/repronim/simple_workflow/simple_workflow/miniconda/envs/bh_demo/lib/python2.7/Bastion.pyo
/home/yoh/proj/repronim/simple_workflow/simple_workflow/miniconda/envs/bh_demo/lib/python2.7/CGIHTTPServer.pyo
/home/yoh/proj/repronim/simple_workflow/simple_workflow/miniconda/envs/bh_demo/lib/python2.7/ConfigParser.pyo
/home/yoh/proj/repronim/simple_workflow/simple_workflow/miniconda/envs/bh_demo/lib/python2.7/Cookie.pyo
/home/yoh/proj/repronim/simple_workflow/simple_workflow/miniconda/envs/bh_demo/lib/python2.7/DocXMLRPCServer.pyo
```
so may be after all, while at it, worthwhile adding support for those?

kyleam: > so we will not bother with `.pyo` (from `python -O`) since they are unlikely to encounter?

No, I just didn't think about it.  Thanks for bringing it up.

kyleam:
I'll wait until tomorrow for additional feedback and then merge.

yarikoptic:
Nope, you will not! I will :-)

PR #383, opened 2019-02-26
CLN: Drop Emacs configuration line from file headers

kyleam:
These lines are visual noise for those that don't use Emacs and can be
an annoyance for those that do.  They configure py-indent-offset, but
this variable is from a third-party python mode.  Emacs users that
don't use this mode get prompted about possibly setting an unsafe
value and need to adjust their configuration to avoid being prompted
about this in the future.

_If_ we want to provide developers using Emacs with configuration for
this project, the better place to do it would be a top-level
.dir-locals.el.  But our settings at this point are typical for a
python project and match the default settings for Emacs's builtin
python.el.

### DIFF
--- a/reproman/__init__.py
+++ b/reproman/__init__.py
@@ -1,4 +1,3 @@
-# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
 # ex: set sts=4 ts=4 sw=4 noet:
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 #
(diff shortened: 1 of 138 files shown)


yarikoptic:
one test failed but I thought it shouldn't relate... by mistake restarted all the tests.
overall -- I don't mind stripping them away here and/or datalad if you find that anyhow useful.  If I manage to indent with tabs when opening in emacs at some point though - I will curse you! ;)

kyleam:
[This change obviously isn't important, but the existence of those lines have annoyed me from the beginning :) ]  

> If I manage to indent with tabs when opening in emacs at some point though - I will curse you! ;)

Both the builtin python.el and the thrid-part python-mode set `indent-tabs-mode` to nil by default, so you'd have to willingly configure those tabs :]

kyleam:
> one test failed but I thought it shouldn't relate.

Ha, I didn't see which one failed before you restarted, but I think it did related.

`test_ssh_class` copies itself:

https://github.com/ReproNim/reproman/blob/d987a9f5963ab9802fb08c14173afce763e5e04d/reproman/resource/tests/test_ssh.py#L89-L90

And then looks for a specific line:

https://github.com/ReproNim/reproman/blob/d987a9f5963ab9802fb08c14173afce763e5e04d/reproman/resource/tests/test_ssh.py#L99

Off by one after removing the header.

PR #394, opened 2019-03-21
Retrace a docker container running in the resource

### LINKED ISSUE #388: retrace docker container
currently we can retrace only an image, but not the container
We should be able to retrace container, and point to the docker image which was used to start it

That could provide a solution to be able to trace execution of docker images.

mjtravers:
Fixes #388 

This PR will upload a docker shim to the resource environment to detect the container image repo digest, which will then get passed to the retracer.

- Not sure I put the docker.shim file in the proper directory
- Also added session handling for 'env' and 'cwd' when executing a command


mjtravers:
@kyleam @yarikoptic 
I am getting this datalad error in the travis test output:
https://travis-ci.org/ReproNim/reproman/jobs/520501164

I am guessing we are pulling too recent a version of datalad as I found this in datalad/interface/run.py:
`# MIH: is_dirty() is gone, but status() can do all of the above!`

Although I am still looking around the code to see where/how/when we install datalad.

Thoughts?

kyleam:
> I am guessing we are pulling too recent a version of datalad

Right, `is_dirty` was dropped.  I'll have to update the code.  In the meantime, you could explicitly install v0.11.4 in tools/ci/install_datalad.

kyleam:
>  I'll have to update the code. In the meantime, you could explicitly install v0.11.4 in tools/ci/install_datalad.

Fixed in master, which I've merged into your branch.

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/perl
+#emacs: -*- mode: shell-script; c-basic-offset: 4; tab-width: 4; indent-tabs-mode: t -*-
+#ex: set sts=4 ts=4 sw=4 noet:
+#
+#
+# This script is a shim for the docker executable. The shim intercepts
+# calls to the docker executable, runs a image id discovery process, and
```
kyleam: s/a/an/

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+# then proceeds with the original call to the original docker executable.
+#
+require 5.002;
+
```
kyleam: Should we use `strict` and `warnings`?  (It wouldn't point to anything major in the current code; just a few redefined variables.)

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+# Print debugging if docker debug switch is set
+my $DEBUG = 0;
+foreach(@ARGV) {
+    $DEBUG = 1 if ($_ eq '--debug' or $_ eq '-D');
+}
```
kyleam: Hmm, because we're going to go through the arguments later to look for "run", I think we should just do so now and

  * look for --debug/-D only in the options before "run"
  * warn if there are any other options before "run" because AFAICS we are currently ignoring them
  * die if "run" is not found (currently we continue with start_index as an undefined variable)

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+	my $command = shift;
+	my $default_message = shift;
+
+	my $output = qx/$command 2>&1/;
+	if ($? != 0) {
+		if ($output) {
+			die $output
+		}
+		else {
+  		die $default_message;
```
kyleam: nitpick: Shouldn't that else's `die` be indented?  But the above is short/simple enough that the if/else block could be converted to a single line.

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+debug("EXTRA_TRACE_FILE: ${EXTRA_TRACE_FILE}");
+my $WAIT_SCRIPT = "wait.sh";
+
+# Locate original Docker executable (Most likely in /usr/bin)
+my $stdout = exec_handler("which -a docker", "Unable to execute system 'which' command");
+my @paths = split(/\s+/, $stdout);
+my $original_docker_exec = '';
+for my $path (@paths) {
+  $original_docker_exec = $path if $path ne $0;
+}
```
kyleam: I don't quite get what's going on here.  My guess is that we use `-a` because the first item should be our own shim, but, if there's more than one non-shim docker in the path, won't the above set the original one to the last one, which wouldn't be the one would take precedence in the unwrapped docker call?

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+print $fh <<'EOF';
+#!/bin/sh
+sleep 30
+EOF
+close($fh);
+chmod(0755, "${TRACERS_DIR}/${WAIT_SCRIPT}");
+
+# Pre-run the docker container in detached mode to get its id
+my $cmd = "${original_docker_exec} run -d --rm ";
+$cmd .= "-v ${TRACERS_DIR}/${WAIT_SCRIPT}:/opt/${WAIT_SCRIPT} ";
```
kyleam: Hmm I thought `-v` was supposed to be limited to directories?

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+$cmd .= "--entrypoint /opt/${WAIT_SCRIPT} ";
+my ($start_index) = grep { $ARGV[$_] eq 'run' } (0 .. @ARGV-1);
```
kyleam: As mentioned above, we should die appropriately when "run" isn't given.  Also I think something like

```perl
List::Util qw(first);
first { $ARGV[$_] eq 'run' } 0..$#ARGV;
```

would be more direct way to get the index.

reproman/interface/docker.shim
```diff
@@ -0,0 +1,141 @@
...
+$cmd .= join(" ", @ARGV[($start_index+1)..$#ARGV]);
```
kyleam: Do we need any quoting for the shell here?

reproman/interface/execute.py
```diff
@@ -132,6 +133,9 @@ def __init__(self, resource, command, cmd_args,
         self.local_trace_dir = op.join(local_cache_dir, 'traces', self.exec_id)
         self.local_tracer_gz = op.join(self.local_tracer_dir,
                                        "reproman_trace.gz")
+        self.local_docker_shim = op.join(op.dirname(op.realpath(__file__)),
+            "docker.shim")
```
kyleam: nitpick: Not indenting to match the opening parenthesis on the previous line will make most pep8 linters light up, I think.  (This applies to a few spots in this series.)

reproman/interface/execute.py
```diff
@@ -190,25 +194,43 @@ def _prepare_remote(self):
...
             "--logfile", op.join(self.remote_trace_dir, "tracer.log"),
             "--dbfile", op.join(self.remote_trace_dir, "trace.sqlite3"),
             "--"
         ]
         return self.session.execute_command(
-            cmd_prefix + [self.command] + self.cmd_args)
+            cmd_prefix + [self.command] + self.cmd_args,
+            env=dict(
+                PATH="{}:$PATH".format(remote_tracer_dir),
+                REPROMAN_TRACERS_DIR=remote_tracer_dir,
```
kyleam: s/TRACERS/TRACER/ for consistency?

reproman/interface/execute.py
```diff
@@ -228,12 +250,22 @@ def post_command(self):
             sort_packages=False,
             find_inputs_outputs=True)
 
+        # include extra trace output if we have any
```
kyleam: This comment doesn't really add anything to the code.  IMO what would be useful is a comment (somewhere, probably not here) about the purpose of the extra trace file.

reproman/interface/execute.py
```diff
@@ -228,12 +250,22 @@ def post_command(self):
...
+        local_extra_trace_file = op.join(self.local_trace_dir,
+            self.extra_trace_file)
+        if op.exists(local_extra_trace_file):
+            with open(local_extra_trace_file, 'r') as fp:
+                extra_files = yaml.load(fp)
```
kyleam: s/load/safe_&/

reproman/interface/tests/test_execute.py
```diff
@@ -143,3 +148,41 @@ def test_trace_local(trace_info):
...
+@mark.skipif_no_docker_engine
+def test_docker_shim():
+    shim = op.join(op.dirname(op.realpath(__file__)),
+            "../docker.shim")
+
+    image_digest = 'debian@sha256:a94839ed73dff831b2382b087b1008877d796946cc716afd4fc72150e082d1b8'
+    command = shim + " --debug run --rm {} cat /etc/debian_version".format(
+        image_digest)
+
+    tempdir = tempfile.mkdtemp()
```
kyleam: Any reason to not use pytest's tmpdir?

reproman/interface/tests/test_execute.py
```diff
@@ -143,3 +148,41 @@ def test_trace_local(trace_info):
...
+        "REPROMAN_EXTRA_TRACE_FILE": op.join(tempdir, "trace.extra.out")
+    }
+
+    # Test with theoretically successful docker run
+    with swallow_logs(new_level=logging.DEBUG) as log:
+        Runner().run(command, env=env)
+        assert_in('buster/sid', log.lines)
+        assert_in('[DEBUG DOCKER SHIM] Found docker executable: /usr/bin/docker', log.lines)
+        assert_in('[DEBUG DOCKER SHIM] Found digest ID: {}'.format(image_digest), log.lines)
+        assert op.exists(op.join(tempdir, env['REPROMAN_EXTRA_TRACE_FILE']))
```
kyleam: These assert_in (and assert_in_in) calls could just be `assert X in log.out`, which would could then take advantage of pytest's assertion introspection.

reproman/resource/docker_container.py
```diff
@@ -225,14 +225,7 @@ class DockerSession(POSIXSession):
...
-
-        if cwd:
-            # TODO: implement
-            # raise NotImplementedError("handle cwd for docker")
-            lgr.warning("cwd is not handled in docker yet")
-            pass
+        command = self._prefix_command(command, env=env, cwd=cwd)
         # if command_env:
             # TODO: might not work - not tested it
             # command = ['export %s=%s' % k for k in command_env.items()] + command
```
kyleam: Since you've added support, can we drop this comment about a potential way to support it?  And I think get_updated_env is handled elsewhere, so we can drop the commented `command_env` assignment above.

reproman/resource/session.py
```diff
@@ -523,6 +521,40 @@ def _parse_envvars_output(self, out):
...
+        and cwd are None
+
+        Parameters
+        ----------
+        command : string or list
+            shell command to be run
+        env : dict
+            ENV vars to prefix to command
+        cwd : string
+            working directory to cd into befoe the command is executed
```
kyleam: s/befoe/before/

reproman/resource/session.py
```diff
@@ -523,6 +521,40 @@ def _parse_envvars_output(self, out):
...
+        with_shell : boolean
+            if True, a "/bin/sh -c" call will wrap the cwd, env and command
+
+        Returns
+        -------
+        string
```
kyleam: But not if `env` and `cwd` are false, right?  I suspect it'd be cleanest to always give back a string.  Plus I think the caller would expect `with_shell` to have an effect even if `env` and `cwd` are false.

reproman/resource/session.py
```diff
@@ -523,6 +521,40 @@ def _parse_envvars_output(self, out):
...
+            command with env and cwd settings prefixed
+        """
+        prefix = ""
+        if env:
+            prefix = " ".join(['export %s=%s &&' % k for k in env.items()])
```
kyleam: * For consistency, should use `.format`
 * Don't we only need one `export` (i.e. doesn't `export` support multiple k=v pairs)?
 * Do we need to `shlex_quote` the values?

reproman/resource/session.py
```diff
@@ -707,11 +739,23 @@ def get_local_session(env={'LC_ALL': 'C'}, pty=False, shared=None):
...
+    # If an update env var references itself, merge the original env value
+    # into the update value.
+    # (This is to somewhat generalize the "PATH=/foo:$PATH" special case.)
+    for k in update:
+        if k not in env or not isinstance(update[k], str) or \
+                not isinstance(env[k], str):
+            continue
+        token = "${}".format(k)
+        if token in update[k]:
+            update[k] = update[k].replace(token, env[k])
```
kyleam: Hmm, I think this approach is pretty brittle in terms of quoting and partial matches.  For example,

```
>       assert get_updated_env({'a': '/foo', 'b': 2}, {'a': '/bar:$apple'}) \
            == {'a': '/bar:$apple', 'b': 2}
E       AssertionError: assert {'a': '/bar:/foopple', 'b': 2} == {'a': '/bar:$apple', 'b': 2}
E         Omitting 1 identical items, use -vv to show
E         Differing items:
E         {'a': '/bar:/foopple'} != {'a': '/bar:$apple'}
```

reproman/resource/ssh.py
```diff
@@ -180,16 +180,11 @@ def _execute_command(self, command, env=None, cwd=None, handle_permission_denied
...
-                      + ' ' + command
-
-        if cwd:
-            raise NotImplementedError("implement cwd support")
+        command = self._prefix_command(command, env=env, cwd=cwd,
+            with_shell=False)
 
         try:
-            result = self.connection.run(command, hide=True)
+            result = self.connection.run(command_as_string(command), hide=True)
```
kyleam: If _prefix_command were to always return a string, the above command_as_string call could be dropped.  (I think we also need to worry about double-quoting the string.)

reproman/interface/docker.shim
mjtravers: For the second bullet, no options are ignored for the running of the command after we are done gathering the info we need:

`$cmd = "${original_docker_exec} " . join(" ", @ARGV);`

We do ignore the options when we are figuring out the container id because they are not consequential for what we are doing for that run.

kyleam: Indeed.  Thanks for the correction.

reproman/interface/docker.shim
mjtravers: -v can bind mount either files or directories. The "-v" kinda implies "volumes" which makes it confusing.

https://docs.docker.com/storage/bind-mounts/

kyleam: Good to know.  

I scanned the `man docker-run` description for `-v` again and still didn't spot mention of non-directories.  Assuming I'm not missing it, that also adds to the confusion.

reproman/interface/docker.shim
mjtravers: Yes, this is a problem. It should get the second item in the list

reproman/resource/session.py
mjtravers: The _prefix_command method simply returns the original string if env and cwd are false.

Right on the with_shell, it should add the /bin/sh invokation even without env or cwd

kyleam: > The _prefix_command method simply returns the original string if env and cwd are false.

But my point is that the command can come in as a list, I think, so you'd be returning a list.

mjtravers: Yeah, I figured that out after I commented. There are a couple of problems. I have sorted it out now but have triggered some test fails that need to be tracked down and fixed

reproman/interface/execute.py
```diff
@@ -190,25 +197,44 @@ def _prepare_remote(self):
         # TODO: might want to add also a "marker" so within the trace
         #       we could avoid retracing session establishing bits themselves
 
+        # Upload Docker shim
+        self.remote_docker_shim = op.join(remote_tracer_dir, "docker")
+        if not self.session.exists(self.remote_docker_shim):
+            self.session.put(self.local_docker_shim, self.remote_docker_shim)
```
kyleam: This has the potential to become stale if the shim is updated.  Given this is a small script, I think it'd be better to just put the file there (at least with local and ssh, `put` seems happy to overwrite the file).  To avoid the unnecessary copy we could even check whether the command is "docker".

mjtravers: This has been fixed as suggested:

be05d09362196764e7bec940ceb2248f659afdda

kyleam: Thanks, @mjtravers.  From my POV this is good to go.

yarikoptic:
Thank you @mjtravers and @kyleam for the review.  I guess if any other additional concerns arise, we will address them later ;)  Merging!

Issue #400, opened 2019-04-01
pyout: "Interference" between pyout output and interactive prompt


yarikoptic:
Observed while testing #399: If upon `--refresh` access to the resource requires interactive session (e.g. enter a password), then

1. It might be not 100% clear to which resource that password prompt relates
```shell
$> reproman ls --refresh
NAME                 TYPE             ID                                          STATUS          
debian-testing-conda docker-container 64e16d0fd025                                CONNECTION ERROR
discovery            ssh              88a9e369-7780-489e-a255-5a34d1b1717b        querying…       
localshell           shell            thebest                                     available       
smaug                ssh              8b253827-9133-4a3b-9405-f7b39ef88eb7        N/A             
smaug-test           ssh              16412f12-8d99-48e7-a40e-940dc4454201        N/A             
test                 docker-container c6f0d0afcdc48a93c1000…f7e9d3d368d197027c6a0 CONNECTION ERROR
Password: 
```
In my case it was a single one (`discovery`) I guess and I could identify by seeing `querying…`.  But what would happen if it is multiple?  I wonder if there is any way we would be able somehow to disambiguate?

2. after I "fail" to authenticate (just pressed `Enter`) - the entry for discovery was duplicated (since pyout does not know I guess that was more output to the terminal which it was not in control of), and followup `localshell` line is gone:
```
$> reproman ls --refresh
NAME                 TYPE             ID                                          STATUS          
debian-testing-conda docker-container 64e16d0fd025                                CONNECTION ERROR
discovery            ssh              88a9e369-7780-489e-a255-5a34d1b1717b        querying…       
discovery            ssh              88a9e369-7780-489e-a255-5a34d1b1717b        CONNECTION ERROR
smaug                ssh              8b253827-9133-4a3b-9405-f7b39ef88eb7        N/A             
smaug-test           ssh              16412f12-8d99-48e7-a40e-940dc4454201        N/A             
test                 docker-container c6f0d0afcdc48a93c1000…f7e9d3d368d197027c6a0 CONNECTION ERROR
Password: 
```
Is there may be some way for pyout to sense that there was additional output? e.g. to have knowledge of total # of lines in the terminal and current position, so it could see if it matches its internal "knowledge"?  

If there is no way, I guess the workaround could be at least some cmdline option to `ls` to disable concurrent querying and proceed one at a time?
Or may be there is a way to "mock" PTY so it goes through pyout somehow?

kyleam:
[...]

PR #402, opened 2019-04-02
ENH: ssh: Include resource name in password prompt

kyleam:
When the inventory contains a password-protected resource, calling
`niceman ls --refresh` will display a "Password: " prompt _before_
displaying the line for the resource.  Add the resource name to the
prompt so that users know what they are entering a password for.

For operations that work on a single resource (e.g., login), this
added information isn't useful, but it also doesn't hurt.

Re: #400


yarikoptic:
Sweet! thanks

PR #408, opened 2019-04-25
run: Parallel subjobs

kyleam:
Rework the orchestrators and submitters to support concurrent subjobs.

In terms of the `run` interface, this introduces a `--batch-spec` flag and a `--batch-parameter` flag for simple cases.  These control the number of subjobs.

An example invocation would be:

```
reproman run --follow -r ss --sub condor --orc datalad-pair-run --bp 'thing=thing-*' \
  --input '{p[thing]}' sh -c 'cat {p[thing]} {p[thing]} >doubled-{p[thing]}'
```

Assuming `[thing-a, thing-b, ...]` are globbed, that's equivalent to a --batch-spec with

```
thing:
- thing-a
- thing-b
[...]
```

### DIFF
--- a/reproman/interface/tests/test_run.py
+++ b/reproman/interface/tests/test_run.py
@@ -106,6 +106,20 @@ def test_combine_batch_params(params, expected):
     assert actual == expected
 
 
+def test_combine_batch_params_glob(tmpdir):
+    tmpdir = str(tmpdir)
+    create_tree(tmpdir, {"aaa": "a",
+                         "subdir": {"b": "b", "c": "c"}})
+    with chpwd(tmpdir):
+        res = sorted(_combine_batch_params(["foo=a*,subdir/*,other"]),
+                     key=lambda d: d["foo"])
+        assert list(res) == [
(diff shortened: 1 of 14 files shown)


reproman/interface/tests/test_run.py
```diff
@@ -74,6 +78,87 @@ def test_combine_specs(specs, expected):
...
+      """\
+- a: '1'
+  b: '3'
+- a: '2'
+  b: '3'""")],
+    ids=["empty", "one", "two, one varying"])
+def test_resolve_batch_params_eq(tmpdir, params, spec):
+    fname = op.join(str(tmpdir), "spec.yml")
+    with open(fname, "w") as fh:
+        fh.write(spec)
```
yarikoptic: Minor FYi: there is create_tree which is used by with_tree decorator

kyleam: Yes, I even use `create_tree` in this PR (ae58628b) :)

I do like `{create,with}_tree`, but I also don't think there's much value in replacing every simple `write` call with it.

PR #413, opened 2019-05-08
 BF: orchestrator: Fix target commit checkout for shell resource

kyleam:
Adjust the local shell prepare_remote() procedure to check out the target *after* an existing remote dataset is updated.


kyleam:
Hmm, I'll need to look into that failure.  I believe this passed for me locally.

yarikoptic:
With 10x more lines provided for tests than for the actual change it must be correct! Thanks ;-)

PR #414, opened 2019-05-08
RF: external_versions: Drop cmd:datalad

kyleam:
There's no reason to add a custom "cmd:" variant when we can inspect
at the version from the python module.

### DIFF
--- a/reproman/support/external_versions.py
+++ b/reproman/support/external_versions.py
@@ -98,12 +98,6 @@ def _get_condor_version():
     return _runner.run(['condor_version'])[0].split()[1]
 
 
-def _get_datalad_version():
-    """Return version of available datalad"""
-    # Example output: datalad 0.11.3.dev17
-    return _runner.run(['datalad', '--version'])[0].split()[1]
-
-
 class ExternalVersions(object):
     """Helper to figure out/use versions of the externals (modules, cmdline tools, etc).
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
There could come a surprise when datalad module is different from the datalad used in command line. Having seen people just adjusting PYTHONPATH to point somewhere instead of using eg virtualenv is not even uncommon. Do might be better to keep them separate. But what problem are you trying to solve?

kyleam:
@yarikoptic:

> There could come a surprise when datalad module is different from the datalad used in command line. 

True, and I view that as all the more reason to drop the cmd:datalad entry.  (The initial text of the commit message even mentioned this, though I pruned it before submitting.)  We use the Python API for any operations on the local machine, so skip.no_datalad should look there and it is unnecessary and potentially confusing to carry cmd:datalad at all.

> Do might be better to keep them separate. But what problem are you trying to solve?

We use only the Python API locally.  This PR isn't joining them: it's just dropping code that I mistakenly added when I really meant `external_versions["datalad"]`.  If down the road we need to inspect the command-line version of DataLad (e.g., perhaps you want to help users troubleshoot or inspect installation issues), it's trivial to resurrect this.  Until then it's unnecessary code that I added by mistake.

That being said, I'm fine keeping the cmd:datalad entry around, and updating this PR to only adjust skip.no_datalad.

kyleam:
I've updated the series to first drop the incorrect use of cmd:datalad and then follow up with a commit that removes cmd:datalad from external_versions.  If you'd like to keep cmd:datalad  around, we can just drop the second commit.

yarikoptic:
since there is a purpose (to avoid confusion in our reproman codeworld) - I am ok with it.

PR #418, opened 2019-05-16
NF(WiP): SpecObject .find and .__iadd__

yarikoptic:
Sits on top of #273.  Only the last commit (6e9178f3652ec1d7ced44dbbed71c16ec050e1f7 to be rewritten during WiP) is pertinent to this PR ATM

Should make it possible to "join" multiple specifications, and also
avoid duplicates while retracing.
    
Seems to run without crashing on a sample but I do not think it works correctly yet


kyleam:
On the call, @chaselgrove asked if anyone had a snippet for triggering the duplicate distributions in order to test this PR.  I wasn't able to find a snippet posted anywhere (e.g., to generate linked traces in gh-367).  If my understanding of the issue is correct, the necessary condition is for a tracer to add a *new* file to the list of unknown files that will be picked up by a tracer on the *next* iteration.  Because of this, the order of the tracers in retrace.py matters.  The current order in retrace.py is `DebTracer`, `RPMTracer`, `CondaTracer`, `VenvTracer`, `VCSTracer`, `DockerTracer`, `SingularityTracer`.

On a Debian system, here's a snippet that triggers two Debian distribution entries:

```sh
cd $(mktemp -d --tmpdir reproman-retrace-XXXXXXX)
virtualenv --python=python3 venv
. ./venv/bin/activate
reproman retrace /bin/sh $PWD/venv/lib/python3.7/abc.py
```

So for the first pass, the `DebTracer` adds a debian distribution entry because it picked up `dash` from `/bin/sh`, then `VenvTracer` adds a venv distribution based on encountering abc.py.  In the virtual environment, abc.py is a link to `/usr/lib/python3.7/abc.py`, so the VenvTracer passes that file back as an unknown file.  On the next iteration `DebTracer` picks it up, and adds a *second* debian distribution.

---

On a non-Debian system I think this is harder to trigger.  The easiest thing I came up with so far is to put `VCSTracer` before `VenvTracer`:

```diff
diff --git a/reproman/interface/retrace.py b/reproman/interface/retrace.py
index 62f1be60d..597b327aa 100644
--- a/reproman/interface/retrace.py
+++ b/reproman/interface/retrace.py
@@ -244,6 +244,6 @@ def get_tracer_classes():
     from reproman.distributions.vcs import VCSTracer
     from reproman.distributions.docker import DockerTracer
     from reproman.distributions.singularity import SingularityTracer
-    Tracers = [DebTracer, RPMTracer, CondaTracer, VenvTracer, VCSTracer,
+    Tracers = [DebTracer, RPMTracer, CondaTracer, VCSTracer, VenvTracer,
         DockerTracer, SingularityTracer]
     return Tracers
```

Then something pass `reproman retrace` a file from a git repo and a file from a virtual environment that has an editable package for that git repo:

```sh
cd $(mktemp -d --tmpdir reproman-retrace-XXXXXXX)

git clone https://github.com/docker/docker-py

virtualenv --python=python3 venv
. ./venv/bin/activate
pip install -e docker-py

reproman retrace $PWD/venv/lib/python3.7/site-packages/six.py $PWD/docker-py/setup.py
```

I see the distribution for the virtual environment sandwiched between two git distributions.

yarikoptic:
@chaselgrove will you have some time in the near future to address @kyleam 's comment?

PR #420, opened 2019-05-22
TST: Don't assume Ed is installed

kyleam:
test_dpkg_manager_identify_packages was failing for me locally because I didn't have Ed installed on my system.

### DIFF
--- a/reproman/distributions/tests/test_debian.py
+++ b/reproman/distributions/tests/test_debian.py
@@ -15,6 +15,7 @@ import logging
 from reproman.distributions.debian import DebTracer
 from reproman.distributions.debian import DEBPackage
 from reproman.distributions.debian import DebianDistribution
+from reproman.support.external_versions import external_versions
 
 import pytest
 
@@ -25,6 +26,8 @@ from reproman.tests.skip import mark
 
 
 @mark.skipif_no_apt_cache
+@pytest.mark.skipif("cmd:ed" not in external_versions,
+                    reason="Ed is not installed")
 def test_dpkg_manager_identify_packages():
     files = ["/bin/ed"]
     tracer = DebTracer()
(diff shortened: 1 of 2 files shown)


kyleam:
The `REPROMAN_TESTS_NONETWORK=1 http_proxy= https_proxy=` run seems to be consistently hanging and getting terminated due to a lack of output (restarted twice).

reproman/distributions/tests/test_debian.py
```diff
@@ -28,6 +26,8 @@
 
 
 @mark.skipif_no_apt_cache
+@pytest.mark.skipif("cmd:ed" not in external_versions,
```
yarikoptic: I can indeed see minimized environments having no `ed`. I am just curious -- was it some real system without `ed` installed?

kyleam: I like to think my laptop is a real system :]

From the commit message of e2ec169a5:

> This change assumed that Ed is installed, which isn't the case for my local system.

yarikoptic:
> The `REPROMAN_TESTS_NONETWORK=1 http_proxy= https_proxy=` run seems to be consistently hanging and getting terminated due to a lack of output (restarted twice).

I guess something changed in travis configuration... have no clue why it would hang on `route add `.
I think we should remove it, especially since we did that in datalad awhile ago in 64a4dd7aa31b7ce8d890e8cf6830c1aee1905d6b with me consciously describing it as

    Remove messing with the route table -- not sure how/if it works currently
    
    see eg. https://github.com/datalad/datalad/pull/841 where reconfirmed that it just
    screws it up

kyleam:
> I think we should remove it, especially since we did that in datalad awhile ago in 64a4dd7aa31b7ce8d890e8cf6830c1aee1905d6b with me consciously describing it as

All right, thanks for the pointer.  I'll open a PR to do that.

reproman/distributions/tests/test_debian.py
kyleam: Oh, and you don't even have to look at the commit messages :]  From the PR description:

> test_dpkg_manager_identify_packages was failing for me locally because I didn't have Ed installed on my system.

kyleam:
Closing since gh-423 is proposing to find a more stable target.

PR #425, opened 2019-05-23
RF: use /bin/bash instead of /bin/ed or /sbin/iptables, define centrally

### LINKED ISSUE #423: replace tracing /sbin/iptables with smth else (/bin/bash?)
Since we are not tracking symlinks/alternatives (#102 ) ATM, tests started to fail for me locally due to

```
$> ls -l {/etc/alternatives,{/usr,}/sbin}/iptables 
lrwxrwxrwx 1 root root 25 Feb 23 10:58 /etc/alternatives/iptables -> /usr/sbin/iptables-legacy*
lrwxrwxrwx 1 root root 18 Dec 26 11:19 /sbin/iptables -> /usr/sbin/iptables*
lrwxrwxrwx 1 root root 26 Dec 26 11:19 /usr/sbin/iptables -> /etc/alternatives/iptables*
```
and then `dpkg -S` knowing only about the actual one

```shell
$> dpkg -S {/etc/alternatives,{/usr,}/sbin}/iptables /usr/sbin/iptables-legacy 
dpkg-query: no path found matching pattern /etc/alternatives/iptables
dpkg-query: no path found matching pattern /usr/sbin/iptables
dpkg-query: no path found matching pattern /sbin/iptables
iptables: /usr/sbin/iptables-legacy
```

yarikoptic:
/bin/ed was found to be absent on real-life systems (apparently not that many
tools depend on good old ed).

iptables is being deprecated by nftables, and exists as a symlink/alternative
on Debian systems. Unfortunately we do not even track alternatives ATM (ref #102)

Closes #423

### DIFF
--- a/reproman/distributions/tests/test_debian.py
+++ b/reproman/distributions/tests/test_debian.py
@@ -25,19 +25,23 @@ from unittest import mock
 
 from reproman.utils import swallow_logs
 from reproman.tests.skip import mark
+from reproman.tests.utils import (
+    COMMON_SYSTEM_PATH,
+    COMMON_SYSTEM_PACKAGE,
+)
 
 
 @mark.skipif_no_apt_cache
 def test_dpkg_manager_identify_packages():
-    files = ["/bin/ed"]
+    files = [COMMON_SYSTEM_PATH]
     tracer = DebTracer()
     (packages, unknown_files) = \
         tracer.identify_packages_from_files(files)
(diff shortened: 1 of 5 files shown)


kyleam:
Thank you for your support of Ed-less systems.

yarikoptic:
> Thank you for your support of Ed-less systems.

anything to bring us into the bright Ed-less future!

Issue #406, opened 2019-04-11
clarify new (3.x) singularity ability to get images with @digest specification


yarikoptic:
See https://github.com/ReproNim/reproman/pull/405#issuecomment-482228428 for initial finding/discussion

kyleam:
Let's see what comes of https://github.com/sylabs/singularity/issues/3287

satra:
@kyleam - a few days back i had to authenticate to docker hub to even pull busybox

kyleam:
> @kyleam - a few days back i had to authenticate to docker hub to even pull busybox

Hmm, interesting.  I don't think that's what's going on in the snippets from the linked singularity issue, though, because the call succeeds despite the warning if I drop the `@ID` bit:

```
$ singularity pull shub://vsoch/hello-world      
WARNING: Authentication token file not found : Only pulls of public images will succeed
 62.32 MiB / 62.32 MiB [=========================================] 100.00% 8.31 MiB/s 7s
```

PR #426, opened 2019-05-23
Singularity 3.x compatibility fixes

yarikoptic:
Note: it doesn't anyhow addresses fetching from shub:// urls with `@digest` (#406) which ideally we should get closure on IMHO from upstream.

### DIFF
--- a/reproman/resource/singularity.py
+++ b/reproman/resource/singularity.py
@@ -65,6 +65,16 @@ class Singularity(Resource):
             self.id = None
             self.status = None
 
+    def _run_instance_command(self, cmd, args=[], **kwargs):
+        """Singularity 3.x changed from "instance.cmd" to just "instance cmd"
+        This is a helper to centralize execution
+        """
+        cmd = ['instance.%s' % cmd] \
+            if external_versions['cmd:singularity'] < '3' \
+            else ['instance', cmd]
+        return self._runner.run(['singularity'] + cmd + args, **kwargs)
+
+
     def create(self):
         """
         Create a container instance.
@@ -81,8 +91,8 @@ class Singularity(Resource):
             # Start the container instance.
             # NOTE: Logging stdout and stderr hangs the run call, so we
             # disable the logging in the run call below.
-            self._runner.run(['singularity', 'instance.start', self.image,
-                self.name], log_stdout=False, log_stderr=False)
+            self._run_instance_command('start', [self.image, self.name],
+                log_stdout=False, log_stderr=False)
             info = self.get_instance_info()
 
         # Update status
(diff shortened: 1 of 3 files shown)


reproman/resource/singularity.py
```diff
@@ -81,8 +91,8 @@ def create(self):
             # Start the container instance.
             # NOTE: Logging stdout and stderr hangs the run call, so we
             # disable the logging in the run call below.
-            self._runner.run(['singularity', 'instance.start', self.image,
-                self.name], log_stdout=False, log_stderr=False)
+            self._run_instance_command('start', [self.image, self.name],
+                log_stdout=False, log_stderr=False)
```
kyleam: I don't know what linter pycharm uses, but I would have guessed that it would yell at you about "continuation line under-indented for visual indent" (E128).

reproman/resource/singularity.py
```diff
@@ -65,6 +65,16 @@ def connect(self):
...
+    def _run_instance_command(self, cmd, args=[], **kwargs):
+        """Singularity 3.x changed from "instance.cmd" to just "instance cmd"
+        This is a helper to centralize execution
+        """
+        cmd = ['instance.%s' % cmd] \
+            if external_versions['cmd:singularity'] < '3' \
+            else ['instance', cmd]
+        return self._runner.run(['singularity'] + cmd + args, **kwargs)
+
+
```
kyleam: Additional pep8 nit: Too many blank lines.

reproman/resource/singularity.py
yarikoptic: ok, will address this and another one nearby... FWIW -- pycharm does scream ;-)

reproman/resource/singularity.py
```diff
@@ -65,6 +65,16 @@ def connect(self):
             self.id = None
             self.status = None
 
+    def _run_instance_command(self, cmd, args=[], **kwargs):
+        """Singularity 3.x changed from "instance.cmd" to just "instance cmd"
+        This is a helper to centralize execution
+        """
+        cmd = ['instance.%s' % cmd] \
+            if external_versions['cmd:singularity'] < '3' \
+            else ['instance', cmd]
```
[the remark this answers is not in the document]
yarikoptic: yeah, I am somewhat on the fence every such time ... greedy me nearly always wins to avoid duplication.  But indeed it might be better to switch in general to more conventional one unless `X if ... else ` is very short/obvious

reproman/resource/singularity.py
yarikoptic: yeap

yarikoptic:
now that file is fully PEP8 compliant! (besides a left single TODO ;))

kyleam:
> now that file is fully PEP8 compliant! (besides a left single TODO ;))

Thanks.  A bit much for my tastes (and the args default value was fine in this case), but I'm not going to waste cycles or time on linting.  All good from my POV once tests pass.

PR #432, opened 2019-06-04
ENH: VenvEnvironment: Add system_site_packages attribute

### LINKED ISSUE #428: venv tracer: reflect in the spec that it was configured to use system-site-packages
ATM, if we trace venv created with `--system-site-packages`, we do return information about those system site packages but do nohow mention them in the `venv` record itself.  Then when we get to reinitialize such an environment we would be out of luck since we would need to know either to use that option or not (ATM we do not specify it).

I think we can deduce that if a known package is installed outside of the venv, am I right @kyleam ?

kyleam:
VenvPackage has a local attribute that is set to true if the package
is a local package (as classified by 'pip list --local'), so if
someone wanted to determine whether the environment was created with
--system-site-packages, they could inspect VenvEnvironment for any
packages without local=True.  But this is tedious to do by visual
inspection.  Plus we plan to add support for pruning packages without
files, in which case inspecting the environment object after the fact
would no longer be a reliable way of knowing that
--system-site-packages was used.

Add a VenvEnvironment attribute that explicitly marks whether the
environment has non-local packages.

Closes #428.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,11 @@ matrix:
     env:
     - REPROMAN_TESTS_SSH=1
     - REPROMAN_TESTS_DEPS=core
+    - REPROMAN_TESTS_ASSUME_SSP=1
+    # Set this to test tracing virtual environments created with
+    # --system-site-packages.
+    virtualenv:
+      system_site_packages: true
   - python: 3.5
     # By default no logs will be output. This one is to test with log output at INFO level
     env:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -347,6 +347,13 @@ Refer reproman/config.py for information on how to add these environment variabl
 - *REPROMAN_TESTS_SSH*: 
   Skips SSH tests if this flag is **not** set
 - *REPROMAN_LOGTRACEBACK*: 
+- *REPROMAN_TESTS_ASSUME_SSP*:
+  Set this to indicate that tests can assume that Python system site
+  packages are exposed in the testing environment (i.e., an
+  environment created with `virtualenv --system-site-packages ...`
+  would include system packages) and that there is at least one system
+  package present.  This is currently only relevant for the virtualenv
+  distribution tests.
   Runs TraceBack function with collide set to True, if this flag is set to 'collide'.
   This replaces any common prefix between current traceback log and previous invocation with "..."
 - *REPROMAN_TESTS_NOTEARDOWN*: 
--- a/reproman/distributions/tests/test_venv.py
+++ b/reproman/distributions/tests/test_venv.py
@@ -35,7 +35,12 @@ PY_VERSION = "python{v.major}.{v.minor}".format(v=sys.version_info)
 def venv_test_dir():
     skipif.no_network()
     dirs = AppDirs('reproman')
-    test_dir = os.path.join(dirs.user_cache_dir, 'venv_test')
+
+    ssp = os.getenv("REPROMAN_TESTS_ASSUME_SSP")
+    # Encode the SSP value in the directory name so that the caller doesn't
+    # have to worry about deleting the cached venvs before setting the flag..
+    test_dir = os.path.join(dirs.user_cache_dir,
+                            'venv_test{}'.format("_ssp" if ssp else ""))
     if os.path.exists(test_dir):
         return test_dir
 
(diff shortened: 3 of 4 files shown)


kyleam:
Hmph the added test is failing on Travis but passing locally for me.  That probably means that this PR's approach for setting the system_site_packages attribute is not always sufficient, but what's going on isn't obvious to me.  I'll need to revisit this.

kyleam:
OK, the issue with testing on Travis is resolved.  I don't like the idea of modifying a whole Travis run for the sake of one test, but I couldn't come up with another way to make this work on Travis.

<details>
<summary>range-diff</summary>

```
-:  --------- > 1:  c3ee9349c TST: travis: Expose system packages for virtualenv tests
1:  5680e44b8 ! 2:  684f55f83 ENH: VenvEnvironment: Add system_site_packages attribute
    @@ -4,35 +4,47 @@
     
         VenvPackage has a local attribute that is set to true if the package
         is a local package (as classified by 'pip list --local'), so if
    -    someone wanted to determine whether the environment was created with
    -    --system-site-packages, they could inspect VenvEnvironment for any
    -    packages without local=True.  But this is tedious to do by visual
    -    inspection.  Plus we plan to add support for pruning packages without
    -    files, in which case inspecting the environment object after the fact
    -    would no longer be a reliable way of knowing that
    -    --system-site-packages was used.
    +    someone wanted to determine whether --system-site-packages is needed
    +    to create the current environment, they could inspect VenvEnvironment
    +    for any packages without local=True.  But this is tedious to do by
    +    visual inspection, so add a system_site_packages attribute to
    +    VenvEnvironment that explicitly marks whether the environment has
    +    non-local packages.
     
    -    Add a VenvEnvironment attribute that explicitly marks whether the
    -    environment has non-local packages.
    +    Note that a system_site_packages value of False isn't a guarantee that
    +    --system-site-packages was _not_ used: it may have been used but no
    +    system site packages were installed at the time of tracing.
     
         Closes #428.
     
      diff --git a/reproman/distributions/tests/test_venv.py b/reproman/distributions/tests/test_venv.py
      --- a/reproman/distributions/tests/test_venv.py
      +++ b/reproman/distributions/tests/test_venv.py
    +@@
    + def venv_test_dir():
    +     skipif.no_network()
    +     dirs = AppDirs('reproman')
    +-    test_dir = os.path.join(dirs.user_cache_dir, 'venv_test')
    ++
    ++    ssp = os.getenv("REPROMAN_TESTS_ASSUME_SSP")
    ++    # Encode the SSP value in the directory name so that the caller doesn't
    ++    # have to worry about deleting the cached venvs before setting the flag..
    ++    test_dir = os.path.join(dirs.user_cache_dir,
    ++                            'venv_test{}'.format("_ssp" if ssp else ""))
    +     if os.path.exists(test_dir):
    +         return test_dir
    + 
     @@
          runner.run([pip1, "install", "attrs"])
          # Make sure we're compatible with older pips.
          runner.run([pip1, "install", "pip==9.0.3"])
     +
    -+    # Also set up an environment with --system-site-packages. We won't do
    -+    # detailed inspection of this because its structure depends on a system we
    -+    # don't control, but we still want to make sure that VenvEnvironment's
    -+    # system_site_packages attribute is set correctly.
    -+    pip2 = op.join("venv-nonlocal", "bin", "pip")
    -+    runner.run(["virtualenv", "--python", PY_VERSION, "--system-site-packages",
    -+                "venv-nonlocal"])
    -+    runner.run([pip2, "install", "attrs"])
    ++    if ssp:
    ++        # The testing environment supports --system_site_packages.
    ++        pip2 = op.join("venv-nonlocal", "bin", "pip")
    ++        runner.run(["virtualenv", "--python", PY_VERSION,
    ++                    "--system-site-packages", "venv-nonlocal"])
    ++        runner.run([pip2, "install", "attrs"])
     +
          return test_dir
      
    @@ -52,6 +64,9 @@
              assert_is_subset_recur(expect, attr.asdict(distributions), [dict, list])
      
      
    ++@pytest.mark.skipif(not os.getenv("REPROMAN_TESTS_ASSUME_SSP"),
    ++                    reason=("Will not assume system site packages "
    ++                            "unless REPROMAN_TESTS_ASSUME_SSP is set"))
     +@pytest.mark.integration
     +def test_venv_system_site_packages(venv_test_dir):
     +    with chpwd(venv_test_dir):
    @@ -62,6 +77,9 @@
     +            tracer.identify_distributions([op.join("venv-nonlocal", libpath)]))
     +        assert len(dists) == 1
     +        vdist = dists[0][0]
    ++        # We won't do detailed inspection of this because its structure depends
    ++        # on a system we don't control, but we still want to make sure that
    ++        # VenvEnvironment's system_site_packages attribute is set correctly.
     +        expect = {"environments": [{"packages": [{"files": [libpath],
     +                                                  "name": "attrs"}],
     +                                    "system_site_packages": True}]}
    @@ -95,9 +113,7 @@
          python_version = attrib()
     +    system_site_packages = attrib(
     +        default=False,
    -+        doc="""Was this environment create with --system-site-packages? This is
    -+        set to true if any non-local packages are present in the
    -+        environment.""")
    ++        doc="Are any non-local packages present in the environment?")
          packages = TypedList(VenvPackage)
      
      
```

</details>

yarikoptic:
Thank you @kyleam .  I have not tried to come up with some alternative sensing for either `attempt to create an environment with 'virtualenv --system-site-packages' would have no effect.` programmatically, so let's go for this "instructed" way for now.

PR #433, opened 2019-06-07
Singularity fixes and testing improvements

kyleam:
This PR includes various fixes and touch-ups to `{resource,distribution}/singularity.py` and their tests.  The more notable ones:

  * Fix IndexError when creating a Singularity resource and there are existing instances.

  * The Singularity resources image attribute is now required (rather than defaulting to None, which wasn't handled).

  * Testing should now work with Singularity v3, with one expected failure due to gh-406.

  * test_singularity_resource_class tries harder to clean up its test instances on failure and to not collide with existing instance names.

### DIFF
--- a/reproman/distributions/tests/test_singularity.py
+++ b/reproman/distributions/tests/test_singularity.py
@@ -7,8 +7,7 @@
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
-import os
-import tempfile
+import os.path as op
 
 from ...cmd import Runner
 from ...distributions.singularity import SingularityTracer
@@ -17,13 +16,13 @@ from ...tests.skip import mark
 
 @mark.skipif_no_network
 @mark.skipif_no_singularity
-def test_singularity_trace():
-
+def test_singularity_trace(tmpdir):
+    tmpdir = str(tmpdir)
     # Download and set up singularity image file
     runner = Runner()
     runner.run(['singularity', 'pull',
         'shub://vsoch/hello-world@42e1f04ed80217895f8c960bdde6bef4d34fab59'])
-    image_file = tempfile.mktemp()
+    image_file = op.join(tmpdir, "img")
     os.rename('vsoch-hello-world-master-latest.simg', image_file)
 
     # Test tracer class
@@ -42,9 +41,6 @@ def test_singularity_trace():
     assert dist.images[0].singularity_version == '2.4-feature-squashbuild-secbuild.g780c84d'
     assert dist.images[0].base_image == 'ubuntu:14.04'
     assert 'non-existent-image' in remaining_files
-
-    # Clean up test files
-    os.remove(image_file)
 
     # Run tracer against a Singularity Hub url
     files = ['shub:/GodloveD/lolcow@a59d8de3121579fe9c95ab8af0297c2e3aefd827']
--- a/reproman/distributions/tests/test_singularity.py
+++ b/reproman/distributions/tests/test_singularity.py
@@ -11,8 +11,6 @@ import os
 import tempfile
 
 from ...cmd import Runner
-# from ...distributions.singularity import SingularityDistribution
-# from ...distributions.singularity import SingularityImage
 from ...distributions.singularity import SingularityTracer
 from ...tests.skip import mark
 
--- a/reproman/distributions/tests/test_singularity.py
+++ b/reproman/distributions/tests/test_singularity.py
@@ -29,7 +29,7 @@ def test_singularity_trace():
     # Test tracer class
     tracer = SingularityTracer()
 
-    files = [image_file, 'non-existant-image']
+    files = [image_file, 'non-existent-image']
     dist, remaining_files = next(tracer.identify_distributions(files))
 
     assert dist.images[0].md5 == 'ed9755a0871f04db3e14971bec56a33f'
@@ -41,7 +41,7 @@ def test_singularity_trace():
     assert dist.images[0].build_size == '333MB'
     assert dist.images[0].singularity_version == '2.4-feature-squashbuild-secbuild.g780c84d'
     assert dist.images[0].base_image == 'ubuntu:14.04'
-    assert 'non-existant-image' in remaining_files
+    assert 'non-existent-image' in remaining_files
 
     # Clean up test files
     os.remove(image_file)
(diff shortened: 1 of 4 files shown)


reproman/resource/tests/test_singularity.py
```diff
@@ -10,77 +10,92 @@
...
+            assert resource.status == 'running'
+
+            # Test trying to create an already running instance.
+            resource_duplicate = Singularity(name=name, image=image)
+            resource_duplicate.connect()
+            assert resource_duplicate.id.startswith(name + "-")
+            assert resource_duplicate.status == 'running'
+            list(resource_duplicate.create())
+            assert any("Resource '{}' already exists".format(name) in l
+                       for l in log.lines)
```
mjtravers: Why this way rather than using tests.utils.assert_in() ?

reproman/distributions/tests/test_singularity.py
```diff
@@ -7,56 +7,43 @@
...
     # Download and set up singularity image file
-    runner = Runner()
-    runner.run(['singularity', 'pull',
-        'shub://vsoch/hello-world@42e1f04ed80217895f8c960bdde6bef4d34fab59'])
-    image_file = tempfile.mktemp()
-    os.rename('vsoch-hello-world-master-latest.simg', image_file)
+    runner = Runner(cwd=tmpdir)
+    expected_md5sum = "96563a1ee2a3d9dbd082efb8d263fc09"
+    location = "singularityhub/busybox@{}" + expected_md5sum
+    runner.run(['singularity', 'pull', '--name', 'img', 'shub://' + location])
```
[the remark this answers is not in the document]
kyleam: > It looks like the pulling by hash fail for versions > 3 is limited to shub and not docker hub: [...]

Interesting.

> We could pull an image to use in testing from docker hub rather than shub:

I'd vote to stick to an shub url for a few reasons:

  * Pulling from docker:// doesn't contain much useful metadata for us to inspect (e.g., there doesn't seem to be a size or base image, and we can't expect a certain build date because the build happens at pull).

  * The docker:// pull takes more time because the build happens at pull.

  * I think there's value in marking something that should work as broken.  And there's an unmerged upstream fix (3652), so it probably won't be broken indefinitely.

reproman/resource/tests/test_singularity.py
kyleam: > Why this way rather than using tests.utils.assert_in() ?

Short answer: assert_in doesn't offer any advantage over `assert ... in ...`, but you're right that the form I chose isn't good.  Will update.

Longer: With d3b16d4b5 (TST: Add pytest replacements for nose.tools helpers, 2018-06-25), these helpers were kept around for compatibility and to avoid code churn.  As the code comment from that commit says, I think we should generally avoid these in new code because they don't take advantage of pytest's assertion rewriting.  We could tell pytest to rewrite assertions in the test.utils module with pytest.register_assert_rewrite, but I didn't bother since most of these helpers are replaced adequately by plain assertions + pytest's rewriting.

That being said, you're right that what I did here is **not** an improvement because the message by pytest doesn't provide any details about the failure.  If I had done the faithful conversion of

```
assert "Resource '{}' already exists.".format(name) in log.lines
```

the error reporting would have been as good as assert_in.  I chose to do the `any` bit and match within strings of log.lines because I thought matching on a period was unnecessarily brittle.

So I could (1) use the snippet above with the period, (2) drop the period and use assert_re_in, or (3) drop the period and use `assert ... in log.out`.  I'll go with 3.

kyleam:
Thanks for the review @mjtravers.  Pushed an update.


<details>
<summary>range-diff</summary>

```
 1:  863407d0b !  1:  76372ff90 TST: resource.singularity: Use a more unique instance name
    @@ -56,8 +56,7 @@
              assert resource_duplicate.status == 'running'
              list(resource_duplicate.create())
     -        assert_in("Resource 'foo' already exists.", log.lines)
    -+        assert any("Resource '{}' already exists".format(name) in l
    -+                   for l in log.lines)
    ++        assert "Resource '{}' already exists".format(name) in log.out

              # But using a different name with the same image would work.
     -        resource_nondup = Singularity(name="foo_nondup", image=image)
 2:  ae0c5fc80 !  2:  e371d9fcc TST: resource.singularity: Try to clean up instance on failure
    @@ -25,8 +25,7 @@
     -        assert resource_duplicate.id.startswith(name + "-")
     -        assert resource_duplicate.status == 'running'
     -        list(resource_duplicate.create())
    --        assert any("Resource '{}' already exists".format(name) in l
    --                   for l in log.lines)
    +-        assert "Resource '{}' already exists".format(name) in log.out
     -
     -        # But using a different name with the same image would work.
     -        resource_nondup = Singularity(name=name + "_nondup", image=image)
    @@ -63,8 +62,7 @@
     +            assert resource_duplicate.id.startswith(name + "-")
     +            assert resource_duplicate.status == 'running'
     +            list(resource_duplicate.create())
    -+            assert any("Resource '{}' already exists".format(name) in l
    -+                       for l in log.lines)
    ++            assert "Resource '{}' already exists".format(name) in log.out
     +
     +            # But using a different name with the same image would work.
     +            resource_nondup = Singularity(name=name + "_nondup", image=image)
 3:  c7b51fdf0 =  3:  d27d75746 CLN+TST: distribution.singularity: Remove commented imports
 4:  6b40bcdf2 =  4:  dd692f8c1 CLN+TST: distribution.singularity: Fix typos
 5:  1a0d1e644 =  5:  b6cf507c6 TST: distribution.singularity: Prefer pytest's tmpdir to mkdtemp
 6:  f0c443285 =  6:  4e9a284b0 BF+TST: distribution.singularity: Handle Singularity v3 extension
 7:  cfff703c9 =  7:  489dbfad9 TST: distribution.singularity: Mark expected Singularity v3 failure
 8:  d44921bc9 =  8:  ac89b3b73 TST: distribution.singularity: Test just one source
 9:  0f16b3e7b =  9:  3fe4e3ebb CLN+TST: distribution.singularity: Move repeated access to a variable
10:  184ff0040 = 10:  8c47caf48 ENH+TST: distribution.singularity: Pull by md5sum rather than commit
11:  a235dedb4 = 11:  13805ed92 ENH+TST: distribution.singularity: Test a smaller image
```

</details>

mjtravers:
@kyleam 
Update looks good. I will merge this PR unless there is a reason to hold off. Merge?

kyleam:
>  Update looks good. I will merge this PR unless there is a reason to hold off. Merge?

Should be good to go (or at least is hopefully a net improvement :]).  Thanks.

PR #434, opened 2019-06-13
TST: travis: Use py3.5 for REPROMAN_TESTS_ASSUME_SSP build

kyleam:
As explained in c3ee9349c (TST: travis: Expose system packages for
virtualenv tests, 2019-06-05), the REPROMAN_TESTS_ASSUME_SSP build
needs to be matched up with the default Python version on Travis.

The default distribution for Travis has recently changed from Trusty
to Xenial, so the py3.4/REPROMAN_TESTS_ASSUME_SSP is failing [0].  Use
py3.6 for the REPROMAN_TESTS_ASSUME_SSP, which should come built
in.  (And switch the existing py3.6 run to py3.4 so that we still have
a run for 3.4, which we claim to support.)

[0]: https://travis-ci.org/ReproNim/reproman/jobs/545343359#L194

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,9 @@ matrix:
     # --system-site-packages.
     virtualenv:
       system_site_packages: true
+    # We're assuming a system python version.  Pin the dist so that
+    # this run doesn't break when the default distribution changes.
+    dist: xenial
   - python: 3.5
     # By default no logs will be output. This one is to test with log output at INFO level
     env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
     - REPROMAN_TESTS_SSH=1
     - INSTALL_DATALAD=1
     - INSTALL_CONDOR=1
-  - python: 3.4
+  - python: 3.5
     env:
     - REPROMAN_TESTS_SSH=1
     - REPROMAN_TESTS_DEPS=core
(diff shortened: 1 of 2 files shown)


kyleam:
> Use py3.6 for the REPROMAN_TESTS_ASSUME_SSP, which should come built in.

I was [wrong about that][0].  (I concluded that from [here][1].)  Will try 3.7, I guess.

[0]: https://travis-ci.org/ReproNim/reproman/jobs/545354132#L159
[1]: https://docs.travis-ci.com/user/reference/xenial/#python-support

kyleam:
> Will try 3.7, I guess.

Huh, that is also reported as not installed: https://travis-ci.org/ReproNim/reproman/jobs/545358873#L159

kyleam:
We could of course explicitly specify `dist: trusty` for that run, but it'd be nice to get it working with Xenial.

kyleam:
OK, so it seems what I was missing is that seeing "3.5 is not installed; attempting download" does not mean that the system Python 3.5 won't if `virtualenv: system_site_packages: true`.  Compare [this] to [that].

[this]: https://travis-ci.org/ReproNim/reproman/jobs/545406002#L159
[that]: https://travis-ci.org/ReproNim/reproman/jobs/545406003#L159

yarikoptic:
anything to make testing more robust! ;) thanks!

PR #435, opened 2019-06-14
BF: Correct reported resource types

kyleam:
This PR corrects spots where we output an unusable, underscore-delimited resource type instead of a hyphenated one (e.g., docker_container instead of docker-container).  It also improves the error message when a underscore-delimited form is used.

### DIFF
--- a/reproman/resource/base.py
+++ b/reproman/resource/base.py
@@ -42,7 +42,7 @@ def discover_types():
         f_ = op.basename(f)
         if f_ in ('base.py',) or f_.startswith('_'):
             continue
-        l.append(f_[:-3])
+        l.append(f_[:-3].replace('_', '-'))
     return sorted(l)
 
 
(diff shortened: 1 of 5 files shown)


reproman/resource/base.py
```diff
@@ -29,6 +29,63 @@
...
+    l = []
+    for f in glob(op.join(op.dirname(__file__), '*.py')):
+        f_ = op.basename(f)
+        if f_ in ('base.py',) or f_.startswith('_'):
+            continue
+        l.append(f_[:-3].replace('_', '-'))
+    return sorted(l)
+
+
+def get_resource_class(name):
```
kyleam: There's a good amount of pure code movement here, so something like `git {log,show} --color-moved=dimmed_zebra --color-moved-ws=allow-indentation-change ...` might be helpful for reviewing.

kyleam:
The failing test is unrelated and addressed by #434.

yarikoptic:
Sounds good to my jetlagged mind. Thanks!

PR #436, opened 2019-06-21
ENH: resource: Tolerate unknown resource parameters in inventory

### LINKED ISSUE #171: While instantiating resources from the inventory -- guard against parameters read from the file
Should check which are expected from the attr's, and then if there are some additional spurious ones, should filter them out, issue a warning, and then proceed with instantiation using only the "correct" ones.

kyleam:
If we encounter an invalid parameter when instantiating resources from the inventory, issue a warning and filter out the parameter rather than raising a `ResourceError`.

Closes #171.

### DIFF
--- a/reproman/conftest.py
+++ b/reproman/conftest.py
@@ -11,6 +11,7 @@
 # Some commonly used fixtures
 
 from reproman.formats.tests.fixtures import demo1_spec, reprozip_spec2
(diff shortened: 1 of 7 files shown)


yarikoptic:
LGTM. Pushed one tiny insignificant improvement

PR #437, opened 2019-07-12
TST: Mark failures due to offline Singularity Hub as expected

kyleam:
Singularity Hub has been down for a while.  Mark the tests that fail
because of this as expected failures so that we don't have to look
into each Travis build to make sure that the only failures are the
Singularity Hub-related tests.


yarikoptic:
thank you @kyleam!

PR #449, opened 2019-08-08
TST: travis: Set up SSH for localhost

kyleam:
Doing so will allow us to test `run`, in particular the set prepartion
and state management of DataLad datasets over SSH, without needing to
go through a Docker container.

The tools/ci/ scripts are copied from DataLad (6307824f2), with slight
modifications.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,10 @@ before_install:
   # The ultimate one-liner setup for NeuroDebian repository
   - bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh)
   - travis_retry sudo apt-get update -qq
+  - if [ ! -z "${REPROMAN_TESTS_SSH:-}" ]; then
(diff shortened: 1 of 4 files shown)


kyleam:
Hmm all those runs failed with

```
sudo apt-get update -qq
E: Failed to fetch http://apt.postgresql.org/pub/repos/apt/dists/xenial-pgdg/main/binary-amd64/Packages.gz  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
travis_time:end:0607ab62:start=1565280734704750502,finish=1565280756110488188,duration=21405737686The command "bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh)" failed and exited with 100 during .
```

I'll wait a bit and then see if it sorts itself out with a restart.

https://travis-ci.org/ReproNim/reproman/builds/569428133

yarikoptic:
happened to me as well, and restart eventually helped. Seems to be green now , will merge

PR #452, opened 2019-08-22
run: Use new ssh setup in tests

kyleam:
I'm marking this as a draft because this should probably be split up into at least two PRs, but I haven't gotten around to doing that yet and I'd like to see if it runs on Travis OK because it depends on some Travis-specific setup added in gh-449.

The two main intermixed topics here are

  * adjusting more tests to use SSH resource rather than a local shell resource.  This is important because much of prepare_remote() goes through a different code path if we use the local shell.
  * improving the dirty checks, particularly when there are subdatasets.

### DIFF
--- a/reproman/support/jobs/orchestrators.py
+++ b/reproman/support/jobs/orchestrators.py
@@ -673,12 +673,14 @@ class PrepareRemoteDataladMixin(object):
                           for ln in failed_ref.strip().splitlines()]
         return failed
 
-    def _assert_clean_repo(self):
+    def _assert_clean_repo(self, cwd=None):
         cmd = ["git", "status", "--porcelain",
                "--ignore-submodules=none", "--untracked-files=normal"]
-        if self._execute_in_wdir(cmd):
+        out, _ = self.session.execute_command(
+            cmd, cwd=cwd or self.working_directory)
+        if out:
             raise OrchestratorError("Remote repository {} is dirty"
-                                    .format(self.working_directory))
+                                    .format(cwd or self.working_directory))
 
     def _checkout_target(self):
         self._assert_clean_repo()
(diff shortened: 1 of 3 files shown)


kyleam:
Oh no, our mystery from DataLad's issue 3596 has crossed repos:

```
>           raise GitCommandError(self.args, status, errstr)
E           git.exc.GitCommandError: Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(128)
E             cmdline: /usr/lib/git-annex.linux/git fetch --progress --prune --recurse-submodules=no -v testssh
E             stderr: 'fatal: protocol error: bad line length character: usag'
```

https://travis-ci.org/ReproNim/reproman/jobs/575138936#L2743

reproman/support/jobs/orchestrators.py
```diff
@@ -664,13 +668,19 @@ def failed_subjobs(self):
                     # All right, something looks off.
                     raise exc
             else:
-                failed = list(map(int, failed_ref.strip().split())) or failed
+                # Line format: mode type object filename
+                failed = [int(ln.split()[3])
```
yarikoptic: What happens with filename with spaces? Should split get limited?

kyleam: This subtree is controlled by us, and we write files with integer names to it.

> What happens with filename with spaces?

If a file like "3 2" ended up in there, we'd mistakenly consider job 3 a failed job.  The larger problem though would be how the file name "3 2" got there.

> Should split get limited?

I don't think it matters.  If someone is tampering with `<metadirectory>/failed/`, I'd say all bets are off.  But sure, we might as well.  It doesn't cost anything to ensure we get the full file name and in some cases has a better response to the unexpected invalid file (crash rather take the leading integer silently).

I'll instead update it to split on tab because the format is `mode<SP>type<SP>object<TAB>filename`.  In the scenario above, `int("3 2")` would then fail with a ValueError.  We don't need to worry about "3<TAB>2" because Git will convert the tab to "\t".

kyleam:
I said:
> this should probably be split up into at least two PRs

I've put the first part at gh-453.  The protocol error is resolved but there's additional issues to work out with the localhost ssh on Travis.  Once that's resolved, I'll open another PR with the other topic from this PR.

PR #457, opened 2019-09-12
RF: Singularity hub is back although different.

yarikoptic:
Let's see if anything still works


yarikoptic:
heh, all the signs point again to the throttling by singularity hub, so I guess it cannot really be used in tests any longer

Issue #470, opened 2019-10-19
run: add the additional info level message about "adding the results to git" 


yarikoptic:
ATM we are seeing only "2019-10-18 20:41:13,005 [INFO   ] Fetching results for 20191018-173634-245d" which is actually not exactly the current status.  The results were fetched but now I see datalad activity in adding the results to local dataset.  It would be nice to have more specific updates on the progress since we need to know if it is ok to shut down the remote resource since every minute count$ ;)

Issue #469, opened 2019-10-18
run: option to specify resource actions upon at the initiation and upon completion of the run in --follow mode


yarikoptic:
AWS EC2 time is not cheap, we should make it possible to "fire and forget" a computation without monitoring the progress to just shut it down to not incur charges when it is already idling. 

Ideally `reproman run -r mylarge-ec2` would start `mylarge-ec2` first, do all the run needs, fetch the results and stop the resource upon completion.  So may be it could be as simple as a predefined list of actions to do on the resource, e.g. `--resource-actions={start,stop-after,delete-after}`?

kyleam:
gh-479 addressed this aside from the "email-after" suggestion.

PR #479, opened 2019-10-24
run: Extend --follow with optional stopping/deleting actions

kyleam:
This is a work-in-progress series that addresses gh-470 and partially addresses gh-469.

### DIFF
--- a/reproman/interface/run.py
+++ b/reproman/interface/run.py
@@ -22,6 +22,7 @@ from shlex import quote as shlex_quote
 from reproman.interface.base import Interface
 from reproman.interface.common_opts import resref_opt
 from reproman.interface.common_opts import resref_type_opt
+from reproman.support.constraints import EnsureChoice
 from reproman.support.jobs.local_registry import LocalRegistry
 from reproman.support.jobs.orchestrators import Orchestrator
 from reproman.support.jobs.orchestrators import ORCHESTRATORS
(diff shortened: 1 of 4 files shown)


yarikoptic:
looks good to me.  I will merge

Noted that it seems that `fetch` for `FetchDatalad*` is reported not covered by tests. Are those just not enabled or indeed there is no test to cover possible setups with datalad orchestrators?

kyleam:
> Noted that it seems that `fetch` for `FetchDatalad*` is reported not covered by tests. Are those just not enabled or indeed there is no test to cover possible setups with datalad orchestrators?

I think it is the not-too-uncommon case of codecov reporting things as uncovered that are actually covered.  Later today I'll double check the coverage of the `fetch()` methods with a local run.

kyleam:
Locally I ran

    coverage run -m pytest -xs -rs --integration \
      reproman/interface/tests/test_run.py \
      reproman/support/jobs/tests/test_orchestrators.p

The following lines are uncovered in `FetchDataladPairMixin.fetch`:

https://github.com/ReproNim/reproman/blob/942bbab693431015bbb82e5770b97c087c57108c/reproman/support/jobs/orchestrators.py#L945

https://github.com/ReproNim/reproman/blob/942bbab693431015bbb82e5770b97c087c57108c/reproman/support/jobs/orchestrators.py#L960-L961


The following line is uncovered in `FetchDataladRunMixin.fetch`:

https://github.com/ReproNim/reproman/blob/942bbab693431015bbb82e5770b97c087c57108c/reproman/support/jobs/orchestrators.py#L1012

Scanning through the Travis log, the datalad-based tests do seem to be getting executed (i.e. they haven't started to be skipped for some reason), so I feel fairly confident that the codecov report isn't accurate.

PR #481, opened 2019-11-05
MNT: Unpin docker-py

### LINKED ISSUE #387: Stale docker-py dependency
It looks like we're relying on a pretty old and unsupported version of docker-py.

<https://github.com/ReproNim/reproman/blob/c6dc43f00d7a4826717025500b6a883a19a42a87/setup.py#L71>

The docker module hasn't been released under the name docker-py [since 2016](https://pypi.org/project/docker-py). Since then, there hav been [many releases](https://pypi.org/project/docker) under the name "docker" (no "-py"). I haven't looked into this deeply, but simply changing 'docker-py>=0.3.2' to 'docker' won't work without further adjustments. (There first error I ran into was "module 'docker' has no attribute 'Client' [docker\_container.py:connect:100]".) Also, it seems like we'd have to make adjustments in order to [work with dockerpty](https://github.com/d11wtq/dockerpty/issues/79).

kyleam:
```
The docker module hasn't been released under the name "docker-py"
since 2016 [0] and has seen many releases under the name "docker" [1].
So we've effectively been pinning the release to version 1.10.6 even
though the current version is 4.1.0.

To update to latest version, it looks like we can get by with
substituting APIClient for Client.  Set the new floor to version
2.3.0, the release that introduced APIClient.

[0]: https://pypi.org/project/docker-py
[1]: https://pypi.org/project/docker
```

Closes #387.

---

Light local testing suggests this might work, but let's see what the test run says.  Also, before merging, we should make sure `dockerpty` is hooked up correctly (see gh-387).

### DIFF
--- a/reproman/distributions/tests/test_docker.py
+++ b/reproman/distributions/tests/test_docker.py
@@ -22,7 +22,7 @@ from ...tests.skip import mark
 @mark.skipif_no_network
 @mark.skipif_no_docker_engine
 def test_docker_trace_tag():
-    client = docker.Client()
+    client = docker.APIClient()
     client.pull('alpine:3.6')
 
     tracer = DockerTracer()
@@ -40,7 +40,7 @@ def test_docker_trace_tag():
 @mark.skipif_no_network
 @mark.skipif_no_docker_engine
 def test_docker_trace_id():
-    client = docker.Client()
+    client = docker.APIClient()
     repo_id = 'sha256:f625bd3ff910ad2c68a405ccc5e294d2714fc8cfe7b5d80a8331c72ad5cc7630'
     name = 'alpine@' + repo_id
     client.pull(name)
@@ -62,7 +62,7 @@ def test_docker_trace_id():
 @mark.skipif_no_network
 @mark.skipif_no_docker_engine
 def test_docker_trace_local_image():
-    client = docker.Client()
+    client = docker.APIClient()
     client.pull('alpine:3.6')
     tracer = DockerTracer()
     # Test tracing a local image not saved in a repository
@@ -85,7 +85,7 @@ def test_docker_trace_local_image():
 @mark.skipif_no_docker_engine
 def test_docker_distribution():
 
-    client = docker.Client()
+    client = docker.APIClient()
     session = get_local_session()
 
     # Verify alpine:3.5 image is not stored locally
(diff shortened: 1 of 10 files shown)


kyleam:
I said:

> To update to latest version, it looks like we can get by with substituting APIClient for Client.

Not quite.  We also need to update for `get_archive`s new return value.

<details>
<summary>range-diff</summary>

```
1:  5bfa0116c ! 1:  8ab2d0e17 MNT: Unpin docker-py
    @@ Commit message
         So we've effectively been pinning the release to version 1.10.6 even
         though the current version is 4.1.0.
     
    -    To update to latest version, it looks like we can get by with
    -    substituting APIClient for Client.  Set the new floor to version
    -    2.3.0, the release that introduced APIClient.
    +    To update to latest version, we need to substitute APIClient for
    +    Client (renamed in v2.3.0).  We also need to teach DockerSession.get()
    +    about the new return value of APIClient.get_archive(): as of v3.0.0,
    +    the content is returned as a generator of chunks rather than a byte
    +    string.
    +
    +    Specify v3.0.0 as the new docker-py floor in setup.py
     
         Closes #387.
     
    @@ reproman/resource/docker_container.py: def connect(self):
      
              containers = []
              for container in self._client.containers(all=True):
    +@@ reproman/resource/docker_container.py: def get(self, src_path, dest_path=None, uid=-1, gid=-1):
    +         dest_path = self._prepare_dest_path(src_path, dest_path)
    +         dest_dir = os.path.dirname(dest_path)
    +         stream, stat = self.client.get_archive(self.container, src_path)
    +-        tarball = tarfile.open(fileobj=io.BytesIO(stream.read()))
    ++        # get_archive() returns a generator with the content (in 2 MB chunks by
    ++        # default). Consider exposing those chunks as a stream rather than
    ++        # joining them.
    ++        tarball = tarfile.open(fileobj=io.BytesIO(b"".join(stream)))
    +         tarball.extractall(path=dest_dir)
    +         os.rename(os.path.join(dest_dir, src_basename), dest_path)
    + 
     
      ## reproman/resource/tests/test_docker_container.py ##
     @@
    @@ setup.py: def findsome(subdir, extensions):
          ],
          'docker': [
     -        'docker-py>=0.3.2',
    -+        'docker>=2.3.0',
    ++        'docker>=3.0.0',
              'dockerpty',
          ],
          'aws': [
```

</details>

kyleam:
>  Also, before merging, we should make sure `dockerpty` is hooked up correctly (see [gh-387](https://github.com/ReproNim/reproman/issues/387)).

Tested `reproman login` locally.  Worked fine.

kyleam:
Thanks for taking a look.

PR #485, opened 2019-11-14
DOC: Add sections on managing resources and execution

kyleam:
This series extends the documentation.  Commits 1-4 are cleanup/setup commits.  Commit 5 adds a page on managing resources, and commit 6 adds a page on execution that's mostly dedicated to `reproman run`.  I'm not sure I'm happy with the state of the `run` parts, but I think it passes the threshold of "more useful than nothing at all".

To generate the html output for review, you can run `make -C docs html`.  Perhaps there's enough here that it's now worth hosting somewhere (probably either readthedocs or github pages).

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -5,7 +5,11 @@
 # Treat warnings as errors.... hierarchy.rst is just a dump of notes ATM
 # SPHINXOPTS    = -W
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+# sphinx-build might be coming from system-wide install, and would
+# use /usr/bin/python3 instead of /usr/bin/env python3, so it would not
+# find our module installed under virtualenv
+SPHINXCMD     = $(shell which sphinx-build)
+SPHINXBUILD   = python3 $(SPHINXCMD)
 PAPER         =
 BUILDDIR      = build
 
--- a/docs/source/execute.rst
+++ b/docs/source/execute.rst
@@ -50,7 +50,7 @@ where the working tree looks like this::
       |-- f0.csv -> ../.git/annex/objects/[...]
       `-- f1.csv -> ../.git/annex/objects/[...]
 
-The clean.py script takes two positional arguments (e.g., ``./clean.py
+The ``clean.py`` script takes two positional arguments (e.g., ``./clean.py
 data/f0.csv cleaned/f0.csv``), where the first is a data file to process
 and the second is a path to write the output (creating directories if
 necessary).
@@ -74,7 +74,7 @@ calling ``reproman run --list=orchestrators``.
 
 The main orchestrator choices are ``datalad-pair``,
 ``datalad-pair-run``, and ``datalad-local-run``. If the remote has
-DataLad available, you should go with one of the ``pair`` orchestrators.
+DataLad available, you should go with one of the ``datalad-pair*`` orchestrators.
 These will sync your local dataset with a dataset on the remote machine
 (using `datalad publish`_), creating one if it doesn't already exist
 (using `datalad create-sibling`_).
@@ -100,7 +100,7 @@ specify that the ``datalad-pair-run`` orchestrator should be used::
     ./clean.py data/f0.csv cleaned/f0.csv
 
 Notice that in addition to the orchestrator, we specify the input file
-that needs to be available on the remote. If this file were tracked by
+that needs to be available on the remote. If this file was tracked by
 Git rather than by git-annex, we could get by without declaring it as an
 input because the same revision of the dataset is checked out on the
 remote.
@@ -134,7 +134,7 @@ Our last example invocation could be extended to use Condor like so::
 Note that which batch systems are currently supported is mostly a matter
 of which systems ReproMan developers currently have at their disposal.
 If you would like to add support for your system (or have experience
-with more general approach like DRMAA), we'd welcome help in this area.
+with more general approach like DRMAA_), we'd welcome help in this area.
 
 
 Detached jobs
@@ -181,7 +181,7 @@ The ``--batch-spec`` option is the more cumbersome but more flexible
 counterpart to ``--batch-parameter``. Its value should point to a YAML
 file that defines a series of records, each one with all of the
 parameters for a single subjob command. The equivalent of
-``--batch-parameters=f0,f1`` would be a YAML file with the following
+``--batch-parameter name=f0,f1`` would be a YAML file with the following
 content::
 
    - name: f0
@@ -275,3 +275,5 @@ spec.yaml
 .. _datalad run: http://docs.datalad.org/en/latest/generated/man/datalad-run.html
 .. _datalad update: https://datalad.readthedocs.io/en/latest/generated/man/datalad-update.html
 .. _datalad-htcondor: https://github.com/datalad/datalad-htcondor
+
+.. _DRMAA: https://en.wikipedia.org/wiki/DRMAA
--- a/docs/source/execute.rst
+++ b/docs/source/execute.rst
@@ -1,0 +1,277 @@
+.. _execute:
+
+Execute
+*******
+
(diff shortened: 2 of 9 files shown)


yarikoptic:
Now we have https://reproman.readthedocs.io/en/latest/

I guess we could just merge and proceed from there if you don't mind, so we could review rendered version etc

kyleam:
> I guess we could just merge and proceed from there if you don't mind, so we could review rendered version etc

I don't know what all "etc" encompasses, but reviewing the rendered version locally is easy enough to do.  Let's not avoid a normal review process for that reason.

yarikoptic:
"etc." == ", users could find rendered version (some documentation is better than no documentation, see how it would look in production (different sphinx/extensions could render differently), etc."  ;-)
But ok, I will review from local rendering

yarikoptic:
Great beginning! ;)  I have pushed a few minor tune ups, and will merge as soon as tests pass

kyleam:
>  I have pushed a few minor tune ups, and will merge as soon as tests pass

Please wait for me to have a chance to look at your changes.

docs/source/execute.rst
```diff
@@ -181,7 +181,7 @@ The ``--batch-spec`` option is the more cumbersome but more flexible
 counterpart to ``--batch-parameter``. Its value should point to a YAML
 file that defines a series of records, each one with all of the
 parameters for a single subjob command. The equivalent of
-``--batch-parameters=f0,f1`` would be a YAML file with the following
+``--batch-parameter name=f0,f1`` would be a YAML file with the following
```
kyleam: Good catch.  Thanks.

docs/source/execute.rst
```diff
@@ -74,7 +74,7 @@ calling ``reproman run --list=orchestrators``.
 
 The main orchestrator choices are ``datalad-pair``,
 ``datalad-pair-run``, and ``datalad-local-run``. If the remote has
-DataLad available, you should go with one of the ``pair`` orchestrators.
+DataLad available, you should go with one of the ``datalad-pair*`` orchestrators.
```
kyleam: OK with me, but given the context of the previous sentence, I think my version was clear and easier on the eyes.

docs/source/execute.rst
```diff
@@ -50,7 +50,7 @@ where the working tree looks like this::
       |-- f0.csv -> ../.git/annex/objects/[...]
       `-- f1.csv -> ../.git/annex/objects/[...]
 
-The clean.py script takes two positional arguments (e.g., ``./clean.py
+The ``clean.py`` script takes two positional arguments (e.g., ``./clean.py
```
kyleam: I don't have a strong preference on whether we should use code tags for file names, but we should use a consistent style across the document.  There is another clean.py instance in this same file.  I'll update it.

docs/source/execute.rst
```diff
@@ -100,7 +100,7 @@ specify that the ``datalad-pair-run`` orchestrator should be used::
     ./clean.py data/f0.csv cleaned/f0.csv
 
 Notice that in addition to the orchestrator, we specify the input file
-that needs to be available on the remote. If this file were tracked by
+that needs to be available on the remote. If this file was tracked by
```
kyleam: Although not fleshed out, this is describing a case where the file is tracked by git-annex, not Git.  In that case, "were" is correct.  I'll rephrase this to try to avoid confusion.

docs/Makefile
```diff
@@ -5,7 +5,11 @@
 # Treat warnings as errors.... hierarchy.rst is just a dump of notes ATM
 # SPHINXOPTS    = -W
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+# sphinx-build might be coming from system-wide install, and would
+# use /usr/bin/python3 instead of /usr/bin/env python3, so it would not
+# find our module installed under virtualenv
+SPHINXCMD     = $(shell which sphinx-build)
+SPHINXBUILD   = python3 $(SPHINXCMD)
```
kyleam: Hmm, I'm not sure this is a good idea.  By hard coding python3, you're assuming the system has a python3 executable.  Instead can't a user rely on the `[devel-docs]` target so that sphinx-build is present in the virtualenv regardless of their system?

docs/source/execute.rst
yarikoptic: I went for more specific to make it more explicit and future proof (if another non `datalad-` `pair`) appears, and also consistent with use of `*` glob around

docs/source/execute.rst
yarikoptic: sorry, still missing it -- we are talking about a file (single) not files, so why `were` not `was`?

docs/Makefile
yarikoptic: yeah, probably should have been just `python`.  But without this change, I cannot build docs in a virtualenv with sphinx coming from the system-wide installation, sphinx then fails to find reproman module.

yarikoptic:
> Please wait for me to have a chance to look at your changes.

Thanks for the review!  Let me merge as is now and then proceed with tune ups in a follow up PR. I would like to refer to the doc in a reply to a [question on NeuroStars](https://neurostars.org/t/datalad-containers-data-organization-multiple-questions/5509).

docs/Makefile
kyleam: So why not make a virtualenv that includes sphinx rather than use the system-wide one?  That way, it works for you while not breaking it for people that don't share your system setup.

docs/source/execute.rst
kyleam: "If I were you, then who would be me?"

yarikoptic: yeah, I know about that one but I thought it is specifically about "I"("we") and "you" (that one is plural anyways). Google gives 143k for "if file was" vs 8k for "if file were", which doesn't make it correct of cause, but suggests so.

docs/Makefile
yarikoptic: I typically (and how our CONTRIBUTING.md says ATM) virtualenv's are initiated with `--system-site-packages`.  That makes their initiation faster, verifies correct operation with the system-wide packages (and that possibly we need to boost up some versioned depends if not).  We test on travis without relying on system wide installed versions.  Altogether that broadens the coverage.

> ... it works for you while not breaking it for people ...
Things shouldn't break on my or anyone else's system, if they break -- we should have them fixed.

docs/source/execute.rst
kyleam: The key thing about choosing "was" or "were" with a conditional and a subject that is singular is whether the condition is real or imaginary/hypothetical.

docs/Makefile
kyleam: > Things shouldn't break on my or anyone else's system, if they break -- we should have them fixed.

It breaks when they don't have a python3 executable.

yarikoptic: once again, I think it would be better to just change it to `python`.  Having said that, do you know a distribution where having any version of python3 installed would result in no `python3` executable?  on debian it seems to be not the case. installing `python3-attr` resulted in installing `python3-minimal` which provides python3.
Please let me know if I should submit a PR for it , or you will include it in the follow up (to address other comments)

docs/source/execute.rst
yarikoptic: I am ok either way if you think that "were" was correct. So far found no clear rule besides "subjunctive mood" but that typically seems to refer to the live subjects, and "if file were" sounded a bit odd to me.

kyleam: I've already reworded it on master: 4739fea9fd

docs/Makefile
kyleam: > Please let me know if I should submit a PR for it , or you will include it in the follow up (to address other comments)

I don't plan on submitting a follow-up PR.  On master I've already addressed the two minor points  that I intended to include with this PR before the merge: 4739fea9f, 1da8a366d

PR #488, opened 2019-11-19
ENH: docker: Pull :latest if no tag or digest is specified

### LINKED ISSUE #483: create: Default to :latest tag when pulling Docker image
`reproman create` callers that are familiar with the command-line interface to `docker pull` would reasonably expect that, if they omit a tag, the default `:latest` will be used.  Instead we download all tags:

```
$ docker images busybox
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE

$ reproman create bb -t docker-container -b image=busybox >notag-out 2>&1

$ wc -l notag-out                      
84 notag-out                                    

$ grep -c "Download complete" notag-out
7                                               

$ tail -n3 notag-out                   
2019-11-05 15:14:27,259 [INFO] Downloading [====================>                              ]  924.4kB/2.202MB 
ERROR: 
Interrupted by user while doing magic

$ docker images busybox
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
busybox             1-musl              ff773d70e0ec        5 days ago          1.44MB
busybox             1-glibc             7636bfb4b772        5 days ago          5.2MB
busybox             1-uclibc            020584afccce        5 days ago          1.22MB
busybox             1-ubuntu            d34ea343a882        3 years ago         4.35MB
busybox             1.21-ubuntu         d34ea343a882        3 years ago         4.35MB
busybox             1.21.0-ubuntu       d34ea343a882        3 years ago         4.35MB
busybox             1.23                a84c36ecc374        4 years ago         1.1MB
```

kyleam:
If a tag or digest isn't specified when calling `reproman create`, we
download all tagged images.  This behavior will likely be confusing to
any callers that are familiar with the command-line interface to
docker pull, which defaults to :latest if a tag or digest is omitted.
And the behavior doesn't fit our goal: starting a new container from a
single image.

Teach the DockerContainer resource to append ":latest" when it
receives a bare image name.

Closes #483.

### DIFF
--- a/reproman/resource/docker_container.py
+++ b/reproman/resource/docker_container.py
@@ -30,6 +30,16 @@ import logging
 lgr = logging.getLogger('reproman.resource.docker_container')
 
 
+def _image_latest_default(image):
+    # Given the restricted character set for names, the presence of ":" or "@"
+    # should be a reliable indication of a tag or digest, respectively. See
+    # - https://docs.docker.com/engine/reference/commandline/tag/#extended-description
+    # - vendor/github.com/docker/distribution/reference/regexp.go
+    if ":" not in image and "@" not in image:
+        image += ":latest"
+    return image
+
+
 @attr.s
 class DockerContainer(Resource):
     """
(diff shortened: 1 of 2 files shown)


yarikoptic:
upon a quick check I didn't find any indication that default tag could be anything than ":latest", so should be good to go.

PR #490, opened 2019-12-02
run an HTCondor cluster on AWS

### LINKED ISSUE #392: Prepare cluster AMI
To simplify submission of jobs to the cloud, we ideally should provide some lean AMI(s) for a "cluster the cloud", which we could dynamically bring up/turn down.  Eventually it should even scale dynamically ;)
AMIs should ideally be equipped with docker and singularity, and either condor or torque.
Might be worth looking inside C-PAC (https://fcp-indi.github.io/docs/user/index.html) how they orchestrate the cloud and/either they are using any batch orchestrator. Otherwise https://elasticluster.readthedocs.io/en/latest/ is probably the best bet, although ATM says that " Automated setup of: HPC clusters using SLURM or GridEngine;", so might be worth contributing if needed.

mjtravers:
Fixes #392 

Basic functionality. Still a work in progress

### DIFF
--- a/reproman/resource/aws_condor.py
+++ b/reproman/resource/aws_condor.py
@@ -27,12 +27,13 @@ class AwsCondor(Resource):
     # Configurable options
     size = attrib(
         default=2,
-        doc="Number of EC2 instances in the HTCondor cluster"
+        doc="Number of EC2 instances in the HTCondor cluster.",
+        converter=int
     )
     access_key_id = attrib(
         doc="AWS access key for remote access to your Amazon subscription.")
     secret_access_key = attrib(
-        doc="AWS secret access key for remote access to your Amazon subscription")
+        doc="AWS secret access key for remote access to your Amazon subscription.")
     instance_type = attrib(default="t2.medium",
         doc="The type of Amazon EC2 instance to run. (e.g. t2.medium)")  # EC2 instance type
     security_group = attrib(default="default",
@@ -44,7 +45,7 @@ class AwsCondor(Resource):
     key_filename = attrib(
         doc="Path to SSH private key file matched with AWS key name parameter.")  # SSH private key filename on local machine.
     image = attrib(default="ami-080483fe6f1382eab",
-        doc="AWS image ID from which to create the running instance")  # NITRC-CE v0.50.0
+        doc="AWS image ID from which to create the running instance.")  # NITRC-CE v0.50.0
 
     id = attrib()
     nodes = attrib()  # Node 0 is the master node
@@ -65,7 +66,7 @@ class AwsCondor(Resource):
         if not self.nodes:
             self.nodes = []
             node_configs = []
-            for i in range(int(self.size)):
+            for i in range(self.size):
                 node_configs.append({
                     "type": "aws-ec2",
                     "name": self.name + "_{}".format(i),
@@ -83,7 +84,7 @@ class AwsCondor(Resource):
             self.nodes = []
 
         # Create a connection for each node in the cluster.
-        for i in range(int(self.size)):
+        for i in range(self.size):
             node = resource_manager.factory(node_configs[i])
             node.connect()
             self.nodes.append(node)
@@ -108,17 +109,13 @@ class AwsCondor(Resource):
             "key_name": self.key_name,
             "key_filename": self.key_filename,
             "image": self.image,
-            "nodes": []
-        }
-        for i, node in enumerate(self.nodes):
-            inventory["nodes"].append({
+            "nodes": [{
                 "type": "aws-ec2",
                 "name": "{}_{}".format(self.name, i)
-            })
-        yield inventory
+            } for i in range(len(self.nodes))]
+        }
 
-        if not self.size:
-            raise ValueError("Missing --size switch when creating cluster")
+        yield inventory
 
         def create_ec2(node):
             node_inventory = {}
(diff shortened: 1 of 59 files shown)


reproman/resource/aws_condor.py
```diff
@@ -0,0 +1,180 @@
...
+    def start(self):
+        """
+        Start all the EC2 instances in the cluster.
+        """
+        for node in self.nodes:
+            node.start()
+
+    def stop(self):
+        """
+        Stop all the EC2 instances in the AWS subscription.
```
kyleam: s/in the AWS subscription/in the cluster/ for consistency with the two docstrings above?

reproman/resource/session.py
```diff
@@ -348,6 +349,25 @@ def put(self, src_path, dest_path, uid=-1, gid=-1):
         """
         raise NotImplementedError
 
+    def put_text(self, text, target, executable=False):
```
kyleam: `diff --color-moved=...` tells me this is pure code movement from orchestrators.py.  Looks good/makes sense to me.

reproman/support/jobs/job_templates/cluster/condor_config.local.j2
```diff
@@ -0,0 +1,56 @@
+# THIS FILE IS CONTROLLED BY ELASTICLUSTER
```
kyleam: setup.py should be adjusted so that this file is included in `package_data`.

---

I see you've gone with ansible's convention of using the `.j2` extension for jinja templates.  Not something that should be done in this PR, but perhaps I should change all the other template files to do the same.

mjtravers: @kyleam 
Actually, why don't I move this .j2 extension to .template to be consistent with the other templates. We are not actually using Ansible. I think I did that because, at the time I created the first version of the template, I was trying to integrate elasticluster which was using Ansible. Elasticluster uses the .j2 extension.

kyleam: @mjtravers Sounds good to me.

chaselgrove:
This looks fine to me and works with datalad-pair-run and plain.  The Travis test time out, but tests pass for me.

What else does this need?  Anything else to test?

kyleam:
> The Travis test time out

I'll take a look tomorrow to see if I can figure out what's stalling.

kyleam:
> > The Travis test time out
> 
> I'll take a look tomorrow to see if I can figure out what's stalling.

[Looks][0] like it's `test_orchestrators.py::test_orc_datalad_concurrent[sub:condor-orc:pair-run]`.  Without the changes in this PR, the Travis job doesn't stall.  The likely culprit from this PR is

```diff
diff --git a/reproman/support/jobs/job_templates/submission/condor.template b/reproman/support/jobs/job_templates/submission/condor.template
index bdaf040ad..c125fe507 100644
--- a/reproman/support/jobs/job_templates/submission/condor.template
+++ b/reproman/support/jobs/job_templates/submission/condor.template
@@ -9,6 +9,8 @@ environment  = ""
 Output  = {{ _meta_directory }}/stdout.$(Process)
 Error   = {{ _meta_directory }}/stderr.$(Process)
 Log     = {{ _meta_directory }}/log.$(Process)
+should_transfer_files   = Yes
+when_to_transfer_output = ON_EXIT
 
 {#
   TODO: Need to check spec form compatibility between different batch
```

The changes in this PR do not make any of the tests stall for me locally, so perhaps we're looking at git-annex-related ssh stalling that we've been dealing with on DataLad's end.  That was specific to Xenial, so I've triggered a [job][1] with Bionic to see if the stall still happens there.

In my local runs, the above change leads to a failure in `test_orc_datalad_run[sub:condor-orc:pair]`.  Interestingly, that seems to have passed in the stalled job on Travis.  I need to look into it more.

[0]: https://travis-ci.org/github/ReproNim/reproman/builds/715246144#L3178
[1]: https://travis-ci.org/github/ReproNim/reproman/builds/715294491

kyleam:
I just pushed a commit (ec48e8461) that removes those condor settings.  As I mentioned in that commit, those settings are from trying to get things working without NFS, and I don't think they make sense as of 3d93c7e0a.

chaselgrove:
Works on macOS.

kyleam:
> Works on macOS.

Great, thanks for checking.

The Travis job no longer stalls, and [the job with condor enabled][1] passes.  [Another job fails][2], but it's in the `make -C ...` phase after the tests.  I haven't seen that in recent Travis runs, I don't see it locally, and it seems unlikely to be related to the PR, so I'd say we should wait to worry about it until it pops up again.

[1]: https://travis-ci.org/github/ReproNim/reproman/jobs/717353082
[2]: https://travis-ci.org/github/ReproNim/reproman/jobs/717353083#L1978

chaselgrove:
@kyleam I think we overlapped earlier; I started the macOS test on 929119e, before your message.  Do your changes need another test?

kyleam:
> @kyleam I think we overlapped earlier; I started the macOS test on [929119e](https://github.com/ReproNim/reproman/commit/929119e9ac6333d36f6048b608c3b6a5155a654e), before your message.

Ah, thought that was fast :)

> Do your changes need another test?

If you don't mind, it'd be good to confirm.  Thanks.

chaselgrove:
@kyleam ec48e84 works on macOS.

PR #491, opened 2019-12-02
RF: use python not python3 for running sphinx-build

yarikoptic:
To resolve a hypothetical issue of no python3 executable being
available, while still allowing for virtualenv and system-wide
installation of sphinx.

I will merge if does not cause any breakage in CI

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -9,7 +9,7 @@ SPHINXOPTS    =
 # use /usr/bin/python3 instead of /usr/bin/env python3, so it would not
 # find our module installed under virtualenv
 SPHINXCMD     = $(shell which sphinx-build)
-SPHINXBUILD   = python3 $(SPHINXCMD)
+SPHINXBUILD   = python $(SPHINXCMD)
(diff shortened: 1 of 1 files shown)


kyleam:
In https://github.com/ReproNim/reproman/pull/485#discussion_r352753782 you said

> once again, I think it would be better to just change it to `python`.  Having said that, do you know a distribution where having any version of python3 installed would result in no `python3` executable?

I don't.  I mistakenly thought arch had only a `python` (which is py3), but it's just that arch uses python 3 for their plain, going against [pep-0304][0].  arch still has a `python3` executable.  Given pep-0304, I'd say assuming the presence of a `python3` executable is correct, and my concerns from gh-485 that prompted this PR are misguided.  So I'd vote to close this without a merge.

[0]: https://legacy.python.org/dev/peps/pep-0394/

yarikoptic:
d'oh, saw your comment only upon refreshing the page and after the merge. One way or another I think it all should work, so let's let it RiP until we actually hit an issue

Issue #484, opened 2019-11-10
SLURM submitter 


yarikoptic:
SLURM is taking over HPC world (discovery will move within a year, other HPCs I palpated use slurm as well).  We are doomed to have a submitter for it.
There are packages in debian for it, but unfortunately it doesn't have similar to htcondor super-convenient Michael's debconf interface to configure a basic deployment. 

Reference: Condor submitter: https://github.com/ReproNim/reproman/blob/master/reproman/support/jobs/submitters.py#L155

PR #494, opened 2019-12-05
run: Add Slurm support

kyleam:
This is an initial stab at Slurm support (gh-484).  There are still things to flesh out (most of the known ones should have to-do comment placeholders), and I bet someone familiar with Slurm could suggest better ways to do things (e.g., is it possible to get the job status as a JSON record?).  But I was able to submit simple commands (with a single job and with multiple subjobs), so things seem to be wired up correctly at least at a basic level.

---

Given that I don't have access to an environment with Slurm, setting that up was the more involved part.  Here are the details:

<details>
<summary>Slurm setup</summary>

* Clone <https://github.com/giovtorres/slurm-docker-cluster>

* Apply the patch at the end of this post.  It configures the image for ssh.  A few of the changes in that diff might not be strictly necessary, but I didn't spend much time fiddling with it once I got something working.

* Build the image: `docker build -t slurm-docker-cluster:19.05.1 .`  This takes a bit of time because it builds Slurm from source.

* Run `docker-compose up -d` and then `./register_cluster.sh`.

* I didn't specify a host port in docker-compose.yml, so check the assigned port (`docker port slurmctld 22`).

* Add an entry to .ssh/config for the slurmctld container along the lines of

  ```
  ...
  Host slurm
  HostName <IP>
  User root
  ControlMaster no
  Port <N>
  ```

* `ssh-copy-id slurm` so that you don't have to worry about entering the password ("root").

Then we're to the standard reproman stuff:

* `reproman create sl -t ssh -b host=slurm`

* Create some datalad dataset and run something like `reproman run -r sl --follow --orc datalad-local-run --submitter=slurm --jp root_directory=/data --bp say=a,b sh -c "sleep 10; echo i say {p[say]} >{p[say]}"`

Notice the `root_directory`.  slurm-docker-cluster's README.md says to submit the jobs from that mount point, and I think that's necessary to make things work, but I haven't really looked into it.

</details>

<details>
<summary>sshd patch</summary>

```diff
diff --git a/Dockerfile b/Dockerfile
index d143635..197e92d 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,8 @@ RUN set -ex \
        psmisc \
        bash-completion \
        vim-enhanced \
+       openssh-clients \
+       openssh-server \
     && yum clean all \
     && rm -rf /var/cache/yum

@@ -83,10 +85,21 @@ RUN set -x \
     && chown -R slurm:slurm /var/*/slurm* \
     && /sbin/create-munge-key

+RUN echo 'root:root' |chpasswd
+
+RUN sed -ri 's/^#?PermitRootLogin\s+.*/PermitRootLogin yes/' /etc/ssh/sshd_config
+RUN sed -ri 's/UsePAM yes/#UsePAM yes/g' /etc/ssh/sshd_config
+
+RUN mkdir /root/.ssh
+
 COPY slurm.conf /etc/slurm/slurm.conf
 COPY slurmdbd.conf /etc/slurm/slurmdbd.conf

 COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
 ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]

+EXPOSE 22
+ENV NOTVISIBLE "in users profile"
+RUN echo "export VISIBLE=now" >> /etc/profile
+
 CMD ["slurmdbd"]
diff --git a/docker-compose.yml b/docker-compose.yml
index f0862be..74fcb0e 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,8 +37,11 @@ services:
       - etc_slurm:/etc/slurm
       - slurm_jobdir:/data
       - var_log_slurm:/var/log/slurm
+    ports:
+      - "22"
     expose:
       - "6817"
+      - "22"
     depends_on:
       - "slurmdbd"

diff --git a/docker-entrypoint.sh b/docker-entrypoint.sh
index 9a1203a..1a7a16f 100755
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,6 +23,10 @@ fi

 if [ "$1" = "slurmctld" ]
 then
+    echo "---> Starting sshd ..."
+    ssh-keygen -A
+    /usr/sbin/sshd
+
     echo "---> Starting the MUNGE Authentication service (munged) ..."
     gosu munge /usr/sbin/munged

```

</details>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ matrix:
     - REPROMAN_TESTS_SSH=1
     - INSTALL_DATALAD=1
     - INSTALL_CONDOR=1
+    - SETUP_SLURM=1
   - python: 3.5
     env:
     - REPROMAN_TESTS_SSH=1
@@ -65,6 +66,9 @@ before_install:
   - if [ ! -z "${REPROMAN_TESTS_SSH:-}" ]; then
       sudo eatmydata tools/ci/prep-travis-forssh-sudo.sh;
       tools/ci/prep-travis-forssh.sh;
+    fi
+  - if [ ! -z "${SETUP_SLURM:-}" ]; then
+      tools/ci/setup-slurm-container.sh;
     fi
   - git config --global user.email "test@travis.land"
   - git config --global user.name "Travis Almighty"
--- a/reproman/support/jobs/job_templates/submission/slurm.template
+++ b/reproman/support/jobs/job_templates/submission/slurm.template
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+#SBATCH --output={{ shlex_quote(_meta_directory) }}/stdout.%a
+#SBATCH --error={{ shlex_quote(_meta_directory) }}/stderr.%a
+{#
+  TODO: We need to assess how we treat batch parameters across different
+  submitters---things like whether we should try to expose common names and, if
+  so, what are the discrepancies in the behavior, and how should we deal with
+  that. We should also revisit the goal of making it possible for the caller to
+  extend the submit file template to add stuff like parameters we do not expose
+  and environment modules.
+#}
+{% if memory is defined %}
+#SBATCH --mem={{ memory }}
+{% endif %}
+{% if num_processes is defined %}
+#SBATCH --cpus-per-task={{ num_processes }}
+{% endif %}
+{% if _num_subjobs == 1 %}
+#SBATCH --array=0
+{% else %}
+#SBATCH --array=0-{{ _num_subjobs - 1}}
+{% endif %}
+
+{{ shlex_quote(_meta_directory) }}/runscript $SLURM_ARRAY_TASK_ID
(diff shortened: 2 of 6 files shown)


satra:
@kyleam - you can take a look at this here:

https://github.com/nipy/nipype/blob/master/nipype/pipeline/plugins/slurm.py
(a much simpler worker in the new engine): https://github.com/nipype/pydra/blob/master/pydra/engine/workers.py#L165

satra:
also in pydra we are testing slurm in a container:  https://github.com/nipype/pydra/blob/master/ci/slurm.sh

kyleam:
@satra Thanks for the pointers.  

Just in terms of my question about getting structured output about jobs, what I could gather from a quick skim suggests that sadly we're going to have to stick with parsing the unstructured output with a regexp.

satra:
> Just in terms of my question about getting structured output about jobs, what I could gather from a quick skim suggests that sadly we're going to have to stick with parsing the unstructured output with a regexp.

indeed. the slurm output can be customized by a center. 

one of the things we will try to do in pydra, unless someone has already done it is to analyze the slurm configuration to find out of resources, qos, partitions. we could also consider a test job to determine how to parse output. but these are all fancy things relative to the user simply saying this is where to go and run.

PR #505, opened 2020-02-26
env -0 workaround where not available

### LINKED ISSUE #501: env -0 not supported on macOS
reproman/resource/session.py uses `env -0` to get the environment, but `-0` is not a supported option to `env` on macOS.

chaselgrove:
Fixes #501.

### DIFF
--- a/reproman/resource/session.py
+++ b/reproman/resource/session.py
@@ -17,6 +17,7 @@ import os
 import os.path as op
 import re
 from shlex import quote as shlex_quote
+import subprocess
 
 from reproman.support.exceptions import SessionRuntimeError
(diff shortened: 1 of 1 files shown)


reproman/resource/session.py
```diff
@@ -492,6 +493,12 @@ class POSIXSession(Session):
     # -0 is not provided by busybox's env command.  So if we decide to make it
     # even more portable - something to be done
     _GET_ENVIRON_CMD = ['env', '-0']
+    try:
+        subprocess.check_call(_GET_ENVIRON_CMD,
```
yarikoptic: This would run it locally.

Issue #446, opened 2019-08-07
Our tracer is incompatible with latest conda


kyleam:
As described in  gh-445:

> [...] our distributions.conda isn't compatible
> with the latest conda in at least two ways. We use an environment
> file that doesn't have an extension that conda now expects. That's an
> easy fix, and there's a patch in gh-443. The bigger issue is that
> environments seem to no longer have their own bin/conda unless conda
> is explicitly installed in the environment. (This mirrors the then
> seemingly hard-to-trigger problem initially described in gh-173.) As
> a result, our method for identifying an environment's root prefix is
> broken. Based on a quick scan of envs/NAME/, there may no longer be
> an easy way to map from an environment to its root prefix.

Re: #443

yarikoptic:
> easy fix, and there's a patch in [gh-443](https://github.com/ReproNim/reproman/issues/443). The bigger issue is that
> environments seem to no longer have their own bin/conda unless conda
> is explicitly installed in the environment. (This mirrors the then
> seemingly hard-to-trigger problem initially described in [gh-173](https://github.com/ReproNim/reproman/issues/173).) As

I guess we should check if there is anything else in the file tree which could be taken as a sign of it being a conda distribution.

kyleam:
If we are fine assuming that the environment is in `ROOT/envs/` (as we discussed on the call), we can just identify the root by walking up.  After doing that and some other fixes, I've managed to push the failures down the line in `test_conda_init_install_and_detect`, but I still can't get it passing.  I'm not sure if the issue is on our end or conda's, but I haven't had any luck yet triggering a similar failure outside of reproman.  At any rate, I've pushed the current state to [scratch/conda-update-for-upstream][0] for the time being.  The main message from the failure is shown in the last commit.

[0]: https://github.com/ReproNim/reproman/compare/scratch/conda-update-for-upstream

kyleam:
All right, some progress.  The confused state where conda can't resolve things (e.g., fc73bc3bd) seems to be coming from the non-default channels used in the root environment.  And I _think_ that's an issue because `install_packages` calls `conda install` without specifying channels while the commands that are given an environment yaml do consider the channels (because the channels are included in the yaml).  Anyway, I'll need to return to this later to try for a proper fix of specifying the channels, but the [two][0] [latest][1] pushes to scratch/conda-update-for-upstream are in a passing state at least.

[0]: https://travis-ci.org/github/ReproNim/reproman/builds/673584406
[...]

PR #507, opened 2020-04-09
Fix miniconda download

kyleam:
The download is failing due to a new redirect, causing `test_conda_init_install_and_detect` to fail.  Update the URL, and also tweak the `curl` call to avoid this issue in the future.

This isn't the full rework that our conda functionality needs (gh-446), but it's a prerequisite for that and should put master back into a passing state.


yarikoptic:
woohoo -- thank you @kyleam !

PR #510, opened 2020-04-21
TST: skip: Quiet Runner-based dependency checks

kyleam:
The point is to check whether the functionality is available, so don't
pollute the test output with alarming things like

    [ERROR  ] stderr| Error: communication error [..]
    [ERROR  ] Failed to run ['condor_status'] under None. [...]

---

This is just a minor cosmetic thing that was bothering me as I was working on another topic.

### DIFF
--- a/reproman/support/external_versions.py
+++ b/reproman/support/external_versions.py
@@ -46,24 +46,24 @@ _runner = Runner()
 
 def _get_annex_version():
     """Return version of available git-annex"""
-    return _runner.run('git annex version --raw'.split())[0]
+    return _runner.run(['git', 'annex', 'version', '--raw'])[0]
 
 
 def _get_git_version():
     """Return version of available git"""
-    return _runner.run('git version'.split())[0].split()[-1]
+    return _runner.run(['git', 'version'])[0].split()[-1]
 
 
 def _get_apt_cache_version():
     """Return version of available apt-cache."""
-    return _runner.run('apt-cache -v'.split())[0].split()[1]
+    return _runner.run(['apt-cache', '-v'])[0].split()[1]
 
 
 def _get_system_ssh_version():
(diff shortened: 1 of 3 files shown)


kyleam:
I've updated the series to avoid a few other spots where `.run()` failures were presented as errors despite being expected.  If there is worry about creating conflicts with external_versions.py or cmd.py updates that could come in via the branch 3rd, I'm okay dropping all of the commits after the first.

reproman/support/external_versions.py
```diff
@@ -46,24 +46,29 @@ def __cmp__(self, other):
 
 def _get_annex_version():
     """Return version of available git-annex"""
-    return _runner.run('git annex version --raw'.split())[0]
+    return _runner.run(['git', 'annex', 'version', '--raw'],
+                       expect_fail=True, expect_stderr=True)[0]
```
yarikoptic, line 50: it becomes such a common pattern here (unlike in original datalad version), I wonder if it wouldn't make sense to make a helper here (or may be even at the Runner interface, e.g. `try_run`) e.g.

```python
def _runner_try(cmd, **kwargs):
     return _runner_try(cmd, expect_fail=True, expect_stderr=True, **kwargs)
```
and use it instead of duplicating all the `expect_` kwargs?

kyleam: It (or rather a corrected version that doesn't call itself :) is an overkill in my book, but if you prefer, I'll update it tomorrow.

yarikoptic: Heh heh ;-) I am not of strong opinion on this, up to you

yarikoptic:
As for the 3rd... I think they so would be manageable, and I didn't check how much we already diverged, may be we would just pick up needed fixes as we go

PR #513, opened 2020-05-22
ENH: use PS4 to providing timing of operations in runscript

yarikoptic:
I think it could come very useful for various debugging attempts (not that it probably would "interfere" with the flushing approach @kyleam had done)

### DIFF
--- a/reproman/support/jobs/job_templates/runscript/base.template.sh
+++ b/reproman/support/jobs/job_templates/runscript/base.template.sh
@@ -4,6 +4,9 @@
 
 set -eu
 
+export PS4='> $(date +%T.%N) [$$] '
+set -x
+
 jobid={{ _jobid }}
 subjob=$1
 num_subjobs={{ _num_subjobs }}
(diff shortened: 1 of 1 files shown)


kyleam:
I prefer to not have this in the main line and to not switch to bash unless we really need to.  But if you disagree, feel free to merge.

yarikoptic:
As for bash - I would have preferred to avoid it too, emailed `dash` mailing list about the stalling (you CCed).  Unfortunately with `posh` there is no evaluation of PS4 as bash does so there is no time stamp.
I started to use PS4 more and more instead of ad-hoc INFO messages etc. POSIX aspect though: unfortunately  https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_05_03 says


`PS4`
When an execution trace ( set -x) is being performed in an interactive shell, before each line in the execution trace, the value of this variable shall be subjected to **parameter expansion** and written to standard error. The default value is "+ ". This volume of IEEE Std 1003.1-2001 specifies the effects of the variable only for systems supporting the User Portability Utilities option.

so it is a subject only to `parameter expansion` and not `command substitution`.  So I guess what `bash --posix` does is actually not following POSIX entirely in this respect, and I found no other way to embed at least time into PS4 :-(

yarikoptic:
since could have side effects with dash (upstream issued a fix but it will take to propagate), and due to no immediate need ATM, I will just let it RiP for now.

PR #514, opened 2020-05-25
runscript: Improve status file checks

kyleam:
This series has a commit from @chaselgrove and one from @yarikoptic.  The first commit isn't strictly needed because the second drops the `find` call, but I think it is good to have that change in case we end up going back to a `find` call.

Based on looking at <https://ss64.com/osx/grep.html>, I think this grep call should be okay on macOS.


yarikoptic:
woohoo -- thanks! looks good to me, let's proceed

PR #516, opened 2020-05-26
ENH: small tune ups to "jobs -s" output in case of missing external (datalad)

yarikoptic:
See individual commits.  Now would show
```
$> reproman --dbg jobs -s
[status: N/A - needs datalad] 20200107-213600-f814 on smaug via condor$ '{inputs}' '{outputs}' participant --participan...
...
```
instead of immediately crashing with 
```
$> reproman  jobs -s    
2020-05-25 20:18:57,108 [ERROR  ] DataLad is required for orchestrator 'datalad-pair-run' is missing. [orchestrators.py:__init__:488] (MissingExternalDependency) 
```

### DIFF
--- a/reproman/interface/jobs.py
+++ b/reproman/interface/jobs.py
@@ -16,6 +16,7 @@ import yaml
 
 from reproman.dochelpers import exc_str
 from reproman.interface.base import Interface
+from reproman.support.external_versions import MissingExternalDependency
 from reproman.support.jobs.local_registry import LocalRegistry
 from reproman.support.jobs.orchestrators import ORCHESTRATORS
 from reproman.resource import get_manager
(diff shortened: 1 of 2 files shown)


reproman/support/jobs/orchestrators.py
```diff
@@ -486,7 +486,8 @@ def __init__(self, resource, submission_type, job_spec=None,
                  resurrection=False):
         if not external_versions["datalad"]:
             raise MissingExternalDependency(
```
kyleam, line 488: Just a note: the datalad-bump branch already switched this over to `external_versions.check()`, which results results in `.name` being set in the `MissingExternalDependency` instance.

reproman/interface/jobs.py
```diff
@@ -16,6 +16,7 @@
 
 from reproman.dochelpers import exc_str
 from reproman.interface.base import Interface
+from reproman.support.external_versions import MissingExternalDependency
```
kyleam, line 19: Left-over import from earlier commit

reproman/interface/jobs.py
```diff
@@ -97,11 +98,16 @@ def show_oneline(job, status=False):
...
-        if orc_status == queried_status:
-            # Drop repeated status (e.g., our and condor's "running").
+        try:
+            orc = _resurrect_orc(job)
+            orc_status = orc.status
+            _, queried_status = orc.submitter.status
+            if orc_status == queried_status:
+                # Drop repeated status (e.g., our and condor's "running").
+                queried_status = None
+        except Exception as exc:
```
kyleam, line 108: As the failing tests indicate, this interferes with the expected error handling in the outer level.  So I think the options are (1) broaden the existing handling to catch either the `MissingExternalDependency` you were originally interested in (my preference) or a blanket exception or (2) more comprehensively rework the handling around orc resurrection/status query, adjust the tests, maybe rethink the outer handling.  I don't see a big advantage of the latter, so I'd lean to the narrow variant of 1.

If 2 is the choice, `show` should also be considered.

reproman/support/jobs/orchestrators.py
yarikoptic: Oh cool, One more reason to have that pr merged, which I guess I will do with adding datalad rc into some dependency within setup.py .

PR #518, opened 2020-05-26
ENH: Raise a dedicated JobError if any subjob fails (in --follow mode only)

### LINKED ISSUE #512: run and jobs --fetch: fail with non-0 exit codes if anything (subjob, fetch, etc) fails

yarikoptic:
Closes #512

### DIFF
--- a/reproman/support/exceptions.py
+++ b/reproman/support/exceptions.py
@@ -159,6 +159,23 @@ class OrchestratorError(RuntimeError):
     pass
 
 
+# Job errors
+class JobError(RuntimeError):
+    """`reproman run` execution related error.
+    """
+    def __init__(self, *args, failed=None):
+        super().__init__(*args)
+        self.failed = failed
+
+    def __str__(self):
+        s = super().__str__()
+        if self.failed:
+            if s:
+                s += ' '
+            s += 'Failed subjobs: %s' % ', '.join(map(str, self.failed))
+        return s
+
+
 #
 # SSH support errors, largely adopted from starcluster
 #
--- a/reproman/interface/run.py
+++ b/reproman/interface/run.py
@@ -24,6 +24,7 @@ from reproman.interface.base import Interface
 from reproman.interface.common_opts import resref_opt
 from reproman.interface.common_opts import resref_type_opt
 from reproman.support.constraints import EnsureChoice
+from reproman.support.exceptions import JobError
 from reproman.support.jobs.local_registry import LocalRegistry
 from reproman.support.jobs.orchestrators import Orchestrator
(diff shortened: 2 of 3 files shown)


reproman/support/exceptions.py
```diff
@@ -197,3 +197,22 @@ class SSHAccessDeniedViaAuthKeys(BaseException):
     """
     def __init__(self, user):
         self.msg = "access for user '%s' denied via authorized_keys" % user
+
+
+#
+# Run related exceptions
+#
+class JobError(RuntimeError):
+    """reproman run  execution related error.
```
kyleam, line 206: nit-pick: extra space

reproman/support/exceptions.py
```diff
@@ -197,3 +197,22 @@ class SSHAccessDeniedViaAuthKeys(BaseException):
...
+    """
+    def __init__(self, *args, failed=None):
+        super().__init__(*args)
```
kyleam, line 209: Why match/propagate `RuntimeError`s `args` but not its `kwargs`?

reproman/support/exceptions.py
```diff
@@ -197,3 +197,22 @@ class SSHAccessDeniedViaAuthKeys(BaseException):
     """
     def __init__(self, user):
         self.msg = "access for user '%s' denied via authorized_keys" % user
+
+
+#
+# Run related exceptions
+#
+class JobError(RuntimeError):
```
kyleam, line 205: I think it'd make sense to position this next the `OrchestratorError`.

reproman/support/exceptions.py
yarikoptic: argh - my habit for visual separation of commands from the text. Will do as """`reproman run`"""

reproman/support/exceptions.py
yarikoptic: true, will do

reproman/support/exceptions.py
yarikoptic: it seems that it takes no kwargs:
```
In [3]: RuntimeError(buga="duga")                                                                                
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-825a24e4a3fd> in <module>
----> 1 RuntimeError(buga="duga")

TypeError: RuntimeError() takes no keyword arguments
```
so I will keep as is

kyleam: Thank you for checking.  When inspecting objects with `pydoc`, I should remember to check for the second line below and not trust the signature if I see it, though in this case I don't think its suggestion is useful.

```
 |  __init__(self, /, *args, **kwargs)
 |      Initialize self.  See help(type(self)) for accurate signature.
```

PR #519, opened 2020-05-26
ENH: make --follow "faster"

yarikoptic:
See individual commits.

The last one -- I feel this might be of interest to other places as well were dynamic sleeping/reporting is desired, so might be generalized at some point.


yarikoptic:
Re first sleep removal - I guess time will show indeed.

PR #520, opened 2020-05-26
ENH: run: Specify constraints for --sub and --orc

kyleam:
If an invalid value for the submitter or orchestrator is given, run()
fails with an uninformative KeyError.  Add the accepted values as
constraints so that run() fails early with a helpful message.

This is an alternative to gh-517.

Suggested-by: @yarikoptic


yarikoptic:
Awesome, thanks

Issue #460, opened 2019-09-20
TODO: run - new orchestrator (or documented workflow) for local datalad run "inplace"


yarikoptic:
Might be just a doc issue.

**use case:** use `niceman run` directly on the host (laptop or smaug), with `condor` submitter, and parallelizing.

sample invocation:

```shell
reproman run --follow -r localshell --sub condor --orc datalad-local-run --bp "sl=10,20,30" sleep "{p[sl]}"
```

The doc for orchestrators (`reproman run --list orchestrators`) talks about "remote" directory/sibling execution for all of them, I guess because primary use case so far was remote execution.  But in my case resource is local.  I wonder if docstring could/should be adjusted to clarify that somehow (e.g. remove "remote" and use "directory/sibling on the resource")

With all `datalad-*` orchestrators, reproman works within a clone of the dataset under `~/.reproman/run-root/<uuid>`.  In my case I want computation to run without any cloning, directly in this dataset.  

I have checked `--list parameters` and decided to try specifying `working_directory`, i.e. my full invocation was

```shel
mkdir out; reproman run --follow -r localshell --sub condor --orc datalad-pair --jp working_directory=$PWD --bp "sl=1,2,3,4,5,6,7,8" bash -c 'sleep 10; touch out/"{p[sl]}"'
```

and I think that did the right thing!  I didn't spot a uuid directory being created under `~/.reproman/run-root/`

**Q to @kyleam : Was that the correct specification to fulfill my usecase?**

kyleam:
There is currently no orchestrator that just runs the command in the current dataset.  The current setup was designed around the usecase of executing on a remote dataset.

kyleam:
> I have checked `--list parameters` and decided to try specifying `working_directory`, i.e. my full invocation was
> 
> [...]
> and I think that did the right thing! I didn't spot a uuid directory being created under `~/.reproman/run-root/`
> 
> **Q to @kyleam : Was that the correct specification to fulfill my usecase?**

My original answer was to the title question, but to answer this question directly: That would be how to specify the local dataset as the "remote" working directory and should prevent any setup under the default `~/.reproman/run-root/` location.  As mentioned in the previous answer, the working directory being the same directory as the local dataset wasn't a usecase that was considered, but it might just work.

yarikoptic:
I have retitled this issue to correspond to the desired functionality.  Either it would require a new orchestrator or could be accomplished by tricking an original, let's pursue it.

PR #526, opened 2020-06-01
run: Add orchestrator for local, in-place runs

kyleam:
This series, in particular the second commit, adds a new orchestrator to support the use case described in gh-460.

---

- [x] Improve test coverage of patch.

### DIFF
--- a/reproman/support/jobs/orchestrators.py
+++ b/reproman/support/jobs/orchestrators.py
@@ -1101,6 +1101,46 @@ class DataladPairOrchestrator(
     name = "datalad-pair"
 
 
+class DataladNoRemoteOrchestrator(DataladOrchestrator):
+    """Execute a command in the current local dataset.
+
+    Conceptually this behaves like datalad-pair. However, the working directory
+    for execution is set to the local dataset. It is available for local shell
+    resources only.
+    """
+
+    name = "datalad-no-remote"
+    template_name = "datalad-pair"
+
+    @property
+    @cached_property
+    @borrowdoc(Orchestrator)
+    def working_directory(self):
+        return self.local_directory
+
+    def prepare_remote(self):
+        if not isinstance(self.session, ShellSession):
+            raise OrchestratorError(
+                "The {} orchestrator must be used with a local session, "
+                "but session for resource {} is {}"
+                .format(self.name, self.resource.name,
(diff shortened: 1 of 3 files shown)


reproman/support/jobs/orchestrators.py
```diff
@@ -479,6 +479,28 @@ def _datalad_format_command(ds, spec):
...
+    *args, **kwds
+        Arguments passed to `fn`.
+
+    Raises
+    ------
+    An OrchestratorError if a failure is encountered.
+    """
+    for res in fn(*args, **kwds):
+        lgr.debug("datalad publish result: %s", res)
+        if res["status"] in ["error", "impossible"]:
```
yarikoptic, line 500: Wouldn't it better to instead check if it is not in ok status?
```suggestion
        if res["status"] not in ["ok", "notneeded"]:

```
ATM it should be the same effect
Ref: http://docs.datalad.org/en/stable/customization.html?highlight=Error%20impossible#expected-behavior
but if a new status is introduced - we don't know now if it would be okish or errorish, so wouldn't it be safer to assume it to be errorish?

reproman/support/jobs/tests/test_orchestrators.py
```diff
@@ -433,18 +433,65 @@ def test_orc_datalad_run_container(tmpdir, dataset,
...
+    if should_pass:
+        (ds_a.pathobj / "foo").write_text("data")
+        ds_a.save()
+
+    ds_b = dl.clone(ds_a.path, op.join(topdir, "b"))
+    assert not ds_b.repo.file_has_content("foo")
+    with chpwd(ds_b.path):
+        orc = orcs.DataladNoRemoteOrchestrator(
+            shell, submission_type="local",
+            job_spec={"root_directory": op.join(topdir, "nm-run"),
```
yarikoptic, line 469: What is this nm-run subdirectory? Or nm is for "never mind" , is could be anything?

kyleam: It's a poorly chosen name copied over from the `job_spec` fixture, where "nm" was for "niceman".  Renamed in c67e828f9.

yarikoptic:
Thank you again! I will merge and test in action ;-)

Issue #440, opened 2019-07-24
run: handle presence of a sibling "smoother"?


yarikoptic:
1. initial run failed to another issue (yet to troubleshoot/report separately)
2. I have tried to rerun but even removing remotes etc in the dataset but it didn't help
3. only then I realized that in

```shell
…nim/reproman-master/docs/usecases/bids-fmriprep-workflow-NP/out > reproman run --follow -r "${RM_RESOURCE}" --sub "${RM_SUB}" --orc "${RM_ORC}" --jp container=containers/bids-mriqc data/bids data/mriqc participant,group
2019-07-24 15:05:27,464 [INFO   ] No root directory supplied for smaug; using '/home/yoh/.reproman/run-root' 
[INFO   ] Connecting ... 
ECDSA host key for IP address '129.170.233.9' not in list of known hosts.
[INFO   ] Considering to create a target dataset /home/yoh/proj/repronim/reproman-master/docs/usecases/bids-fmriprep-workflow-NP/out at /home/yoh/.reproman/run-root/1ab254ca-ad49-11e9-95c1-8019340ce7f2 of smaug 
[INFO   ] Fetching updates for <Dataset path=/home/yoh/proj/repronim/reproman-master/docs/usecases/bids-fmriprep-workflow-NP/out> 
.: smaug(+) [ssh://smaug/home/yoh/.reproman/run-root/1ab254ca-ad49-11e9-95c1-8019340ce7f2 (git)]                                                                                                                                  
[INFO   ] Adjusting remote git configuration 
[INFO   ] Running post-update hooks in all created siblings 
[...]

PR #527, opened 2020-06-04
run: Don't abort if sibling unexpectedly exists but has expected target

kyleam:
If the target directory on the remote doesn't exist, `DataladOrchestrator.prepare_remote()` checks whether a sibling of the same name exists and aborts if it does.  In gh-440, @yarikoptic points out that this is unnecessarily strict:

> I think that ideally it should not even complain/crash if remote is configured and has desired target path (I believe it is the fact here)

This series addresses that bit of gh-440.

### DIFF
--- a/reproman/support/jobs/orchestrators.py
+++ b/reproman/support/jobs/orchestrators.py
@@ -765,10 +765,19 @@ class PrepareRemoteDataladMixin(object):
             if not session.exists(self.working_directory):
                 remotes = repo.get_remotes()
                 if resource.name in remotes:
-                    raise OrchestratorError(
-                        "Remote '{}' unexpectedly exists. "
-                        "Either delete remote or rename resource."
-                        .format(resource.name))
+                    if repo.get_remote_url(resource.name) != target_path:
+                        raise OrchestratorError(
+                            "Remote '{}' already exists with another URL. "
+                            "Either delete remote or rename resource."
+                            .format(resource.name))
+                    else:
+                        lgr.debug(
+                            "Remote '%s' matches resource name "
+                            "and points to the expected target, "
+                            "which doesn't exist.  "
+                            "Removing remote and recreating",
+                            resource.name)
+                        repo.remove_remote(resource.name)
 
                 since = None  # Avoid since="" for non-existing repo.
             else:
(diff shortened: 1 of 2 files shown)


reproman/support/jobs/orchestrators.py
```diff
@@ -931,11 +942,12 @@ def fetch(self, on_remote_finish=None):
...
                  resource_name)
-        self.ds.repo.fetch(resource_name, "{0}:{0}".format(ref),
-                           recurse_submodules="no")
+        repo = self.ds.repo
+        repo.fetch(resource_name, "{0}:{0}".format(ref),
+                   recurse_submodules="no")
         failure = False
         try:
-            self.ds.repo.merge(ref)
+            repo.merge(ref)
```
yarikoptic, line 950: note to myself : In DataLad we seems to not overload `merge` in AnnexRepo, and GitRepo.merge seems to not call `.localsync`.  

question: so it is the `ds.update` call below in `else:` which will take care about merging data availability?

reproman/support/jobs/orchestrators.py
```diff
@@ -931,11 +942,12 @@ def fetch(self, on_remote_finish=None):
...
         except DCError as exc:
             lgr.warning(
```
yarikoptic, line 952: so it is just a warning... shouldn't `failure = True` be set?

reproman/support/jobs/orchestrators.py
kyleam: Yes, the `merge` is just a merge of the jobid ref.  cdce12249 (ENH: orchestrators(datalad-pair): Do a more targeted update, 2020-04-09) provides more rationale for its existence.

reproman/support/jobs/orchestrators.py
kyleam: Yeah, it'd make sense to set `failure = True`, because it'd prevent trying the `.get` call on outputs, which would presumably fail or at least not be useful if we failed to bring in the target tree.  However, I don't think it matters much one way or the other once we're in this unexpected error state, and I would prefer it not be done as part of this very unrelated pr.

PR #529, opened 2020-06-05
BF: orchestrators: Don't try to get outputs if ref merge fails

kyleam:
As of cdce12249 (ENH: orchestrators(datalad-pair): Do a more targeted
update, 2020-04-09), we first try to merge in the job ref, and if that
fails, update() isn't called.  Set failure=True so that we do not try
to get output files with a working tree that has not been updated.

Reported-by: @yarikoptic
https://github.com/ReproNim/reproman/pull/527#discussion_r435427464


yarikoptic:
Travis is really green not yellow. Thank you!

Issue #521, opened 2020-05-27
ls: exit with non-0 if listing of any resource has failed


yarikoptic:
```
(git)smaug:~/proj/repronim/reproman-master[doc-usecases]git
$> git describe
v0.2.1-98-gbaab991
(dev3) 2 15077.....................................:Wed 27 May 2020 03:50:18 PM EDT:.
(git)smaug:~/proj/repronim/reproman-master[doc-usecases]git
$> reproman ls local2
RESOURCE NAME        TYPE                 ID                  STATUS    
-------------        ----                 --                  ------    
2020-05-27 15:50:23,649 [WARNING] Manager did not return a resource for local2: Resource matching local2 not found [base.py:_get_resource_config:325] 
Use --refresh option to view updated status.

$> echo $?
0

```

AFAIK this is the only way we could check quickly if resource is actually known to create it if needed.  Otherwise we could  try to create it and ignore when that fails, e.g.

```
reproman create -t shell local || :
```  
but it could fail for various other reasons, not only that it already exists

PR #531, opened 2020-06-05
ls: Give non-zero exit code when an unknown resource is specified

kyleam:
This allows a command-line caller to use `reproman ls` to check if a resource exists, as described at gh-521.

### DIFF
--- a/reproman/interface/ls.py
+++ b/reproman/interface/ls.py
@@ -20,6 +20,7 @@ from ..support.param import Parameter
 from ..resource import get_manager
 from ..ui import ui
 from ..support.exceptions import ResourceError
+from ..support.exceptions import ResourceNotFoundError
 from ..dochelpers import exc_str
 
 from logging import getLogger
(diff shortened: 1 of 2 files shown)


reproman/interface/ls.py
```diff
@@ -100,4 +109,10 @@ def __call__(resrefs=None, verbose=False, refresh=False):
             manager.save_inventory()
         else:
             ui.message('Use --refresh option to view updated status.')
+
+        if unknown_resrefs:
+            raise ResourceNotFoundError(
+                "Could not find the following resources: {}"
+                .format(unknown_resrefs))
```
yarikoptic, line 116:
```suggestion
                .format(' '.join(unknown_resrefs)))
```

kyleam: I'd prefer to just format the list as is, but given you feel strongly enough to suggest a `.join`, I'll make that change.  I think space is a bad delimiter though.  I'll go with ', '.

kyleam:
> Also pr description just says "as described at [gh-521](https://github.com/ReproNim/reproman/issues/521)" - isn't it also closing it?

The second commit has a "Closes ..." statement.

yarikoptic:
Great, thank you. Let's merge when turns green

PR #532, opened 2020-06-16
DOC: ssh: Mention that SSH config nickname is accepted as host

kyleam:
As explained in e381f4b69 (ENH: ssh: Log connection based on Fabric's
values, 2018-11-27), the Fabric switch means that ssh_config files are
now considered, and the host attribute can point to a nickname, with
things like port configured via ssh_config.

Advertise this to `reproman backend-parameters` callers.


yarikoptic:
Thank you for the attention to detail!

PR #533, opened 2020-07-02
Prepend 'reproman-' to the AWS EC2 instance names

chaselgrove:
Using repronim environment names for EC2 instances means that names that
make sense for repronim (e.g. "aws") can be unhelpful in the EC2 console.
Prepending 'reproman-' to the EC2 name is an indication that the instance
belongs to reproman.


yarikoptic:
Sounds like a good idea, and we do something like that for e.g. credentials stored by datalad.  I looked through other backends and insofar I think this is a unique case. Let's proceed

Issue #536, opened 2020-07-09
ENH: recursive `get` for sessions (shell, ssh, etc) should support recursive mode of operation


yarikoptic:
It is necessary for any place where `get` is supposed to obtain a full hierarchy of directories like e.g. getting outputs of the job back to the local system.  

I will send a follow up initial PR for the plain orchestrator to fetch everything not just failed subjobs logs

PR #537, opened 2020-07-09
RF: plain orchestrator - fetch entire job directory

yarikoptic:
It would not work ATM since "get" does not support recursive
copying.
https://github.com/ReproNim/reproman/issues/536
is to address that and be used here after that also for
transfering all the --output(s)


yarikoptic:
superseded by #539 with the test

PR #539, opened 2020-07-29
Plain orchestrator basic fetch

### LINKED ISSUE #535: plain orchestrator: make sure that it at least fetches logs (stdout/err) even if not --output's were specified?

chaselgrove:
Modify the plain orchestrator to fetch status, stdout, and stderr back for all jobs.

Closes #535


yarikoptic:
THANK you @chaselgrove and @kyleam ! Let's proceed!

Issue #542, opened 2020-08-13
Travis: 'make -C docs' failure


kyleam:
The `make -C docs html doctest` phase is failing in one of the jobs (`REPROMAN_TESTS_SSH=1 REPROMAN_TESTS_DEPS=core REPROMAN_TESTS_ASSUME_SSP=1`).  AFAIK the failure was first observed in gh-490 but seems unlikely to be related.

```
Exception occurred:
  File "/usr/lib/python3.5/typing.py", line 1134, in _get_defaults
    code = func.__code__
AttributeError: type object 'GitRunner' has no attribute '__code__'
The full traceback has been saved in /tmp/sphinx-err-o66mmqnq.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile:61: recipe for target 'html' failed
make: *** [html] Error 2
make: Leaving directory '/home/travis/build/ReproNim/reproman/docs'
travis_time:end:26b4d320:start=1597348343230857820,finish=1597348345577634611,duration=2346776791,event=scriptThe command "PYTHONPATH=$PWD make -C docs html doctest" exited with 2.


Done. Your build exited with 1.
```

https://travis-ci.org/github/ReproNim/reproman/jobs/717707039#L1978

kyleam:
The python version used in that job is 3.5.2.  The attribute error above will be caught within `typing._get_defaults` starting with 3.5.3 (specifically cpython's 991d14fee1).

The other builds (even those that use 3.5) use a later version.  It's still not clear why this just started being an issue (perhaps a sphinx update).  One fix would be to bump `dist:` in that that job to get a python version with the fix.

PR #543, opened 2020-08-14
Travis: Fix 'make -C doc ...' failure in --system-site-packages job

kyleam:
The first two commits prepare the `virtualenv: system_site_packages: true` job for an upgrade from Xenial to Bionic.  The last commit does the upgrade to get the CPython fix described at gh-542.


yarikoptic:
Great, thank you!

PR #544, opened 2020-08-19
Add recursion to get and put for local and ssh sessions

### LINKED ISSUE #536: ENH: recursive `get` for sessions (shell, ssh, etc) should support recursive mode of operation

chaselgrove:
Closes #536

### DIFF
--- a/reproman/interface/tests/test_run.py
+++ b/reproman/interface/tests/test_run.py
@@ -493,3 +493,28 @@ def test_jobs_orc_error(context):
                 jobs(queries=[], status=True)
             assert "myshell" not in output.out
             assert "resurrection failed" in log.out
+
+
+def test_recursive_transfer(context):
+    run = context["run_fn"]
+    path = context["directory"]
+    jobs = context["jobs_fn"]
+
+    # Our script takes an inventory of the execute directory so we can be 
+    # sure that all of the files have transferred.  It then creates a tree 
+    # that we verify is returned.
+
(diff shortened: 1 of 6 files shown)


yarikoptic:
Please add a test

chaselgrove:
This tests a local resource.  How would you test the remote (ssh) transfer?

kyleam:
> How would you test the remote (ssh) transfer?

This PR is about a session method, so I think most of the extensive testing should happen in reproman/resource/tests/test_session.py.  That file already has some tests for `get()`, and I believe one of the cases is against a docker ssh container.

https://github.com/ReproNim/reproman/blob/b82a412ecccce33c885247dbd9464972b0e87c3f/reproman/resource/tests/test_session.py#L303

kyleam:
The [last build](https://travis-ci.org/github/ReproNim/reproman/builds/721440920) had three jobs stall in the test_conda phase.  I tried with and without the changes here to trigger these again on Travis but was unable to.  So perhaps they were just due to connectivity issues.

kyleam (approved):
Thanks @chaselgrove.  Looks good to me (assuming the tests still pass after my push, which adds a test to cover the recursive part of `transfer_recursive`).

yarikoptic:
Thank you @chaselgrove and @kyleam .  From a brief look at the code -- looks good to me too.

PR #553, opened 2020-11-05
LSF support

chaselgrove:
This PR adds an LSF submitter.

@kyleam, can you recommend the best way to test this?  I'd model off of condor and slurm, but I only see them in test_orchestrators.py (no references to CondorSubmitter or SlurmSubmitter in tests).

### DIFF
--- a/reproman/support/jobs/submitters.py
+++ b/reproman/support/jobs/submitters.py
@@ -359,11 +359,58 @@ class LocalSubmitter(Submitter):
         return status
 
 
+class LSFSubmitter(Submitter):
+    """Submit an LSF job.
+    """
+
+    name = "lsf"
+
+    def __init__(self, session):
+        super(LSFSubmitter, self).__init__(session)
+
+    @property
+    @borrowdoc(Submitter)
+    def submit_command(self):
+        # LSF can ignore the #BSUB directives if the script is given on 
+        # the command line rather than in stdin.  So we use /bin/bash here 
+        # and pipe the script to bsub in the submit script.
+        return ["/bin/bash"]
+
+    @borrowdoc(Submitter)
+    def submit(self, script, submit_command=None):
+        out = super(LSFSubmitter, self).submit(script, submit_command)
+        m = re.search('Job <(\d+)> is submitted to queue', out)
+        self.submission_id = m.group(1)
+        return self.submission_id
+
+    @property
+    @assert_submission_id
+    @borrowdoc(Submitter)
+    def status(self):
+        out, _ = self.session.execute_command(
+            "bjobs -noheader {}".format(self.submission_id))
+        parts = out.split()
+        if not parts:
+            # bjobs might not know about the job if it is still being queued 
+            # and it won't know about the job if some time has passed since 
+            # it terminated
+            return ("unknown", None)
+        assert parts[0] == self.submission_id
+        if parts[2] in ("PEND", "RUN"):
+            return ("waiting", parts[2])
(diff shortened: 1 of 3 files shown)


kyleam:
>  I'd model off of condor and slurm, but I only see them in test_orchestrators.py (no references to CondorSubmitter or SlurmSubmitter in tests).

Yeah, test_orchestrators is currently the only place that exercises the condor and slurm submitters.  The class names aren't there, but the submitter name gets passed to the orchestrator class ("slurm", "condor", ...).  For example, here's the slurm fixture and one of the slurm tests:

```python
@pytest.fixture(scope="module")
def ssh_slurm():
    skipif.no_ssh()
    skipif.no_slurm()
    from reproman.resource.ssh import SSH
    return SSH("slurm-res", host="slurm")

[...]

@pytest.mark.integration
def test_orc_datalad_slurm(check_orc_datalad, ssh_slurm):
    check_orc_datalad(ssh_slurm, orcs.DataladLocalRunOrchestrator, "slurm")
```

So, if you wanted to follow the same approach, you could start by `s/slurm/lsf/` and then implementing `no_lsf` in the `skip` module.

reproman/support/jobs/submitters.py
```diff
@@ -359,11 +359,58 @@ def status(self):
...
+            # bjobs might not know about the job if it is still being queued 
+            # and it won't know about the job if some time has passed since 
+            # it terminated
+            return ("unknown", None)
+        assert parts[0] == self.submission_id
+        if parts[2] in ("PEND", "RUN"):
+            return ("waiting", parts[2])
+        if parts[2] in ("DONE", "EXIT"):
+            import traceback
+            traceback.print_stack()
```
kyleam, line 403: Leftover debugging?

chaselgrove: Yes; gone.

chaselgrove:
> So, if you wanted to follow the same approach, you could start by `s/slurm/lsf/` and then implementing `no_lsf` in the `skip` module.

Thanks.  I've been trying both that and with a local environment, but the tests are failing for unclear reasons.  I'll return to this next week.

chaselgrove:
The test works as it should, but:

I don't use `skipif` because I use the presence of an environment variable to both trigger and configure the test.

Ideally I'd create `job_spec_lsf` for the LSF test, but I don't see how pytest would let me use that elegantly (without duplicating `check_orc_datalad` completely, the only change being `check_orc_datalad(job_spec, dataset) -> check_orc_datalad_lsf(job_spec_lsf, dataset)`).  So I patch `job_spec` in `check_orc_datalad` (though I suppose this could happen in `job_spec`).  Happy to consider a more elegant option.

kyleam:
> The test works as it should, but:
> 
> I don't use `skipif` because I use the presence of an environment variable to both trigger and configure the test.

Works for me.

> Ideally I'd create `job_spec_lsf` for the LSF test, but I don't see how pytest would let me use that elegantly (without duplicating `check_orc_datalad` completely, the only change being `check_orc_datalad(job_spec, dataset) -> check_orc_datalad_lsf(job_spec_lsf, dataset)`). So I patch `job_spec` in `check_orc_datalad` (though I suppose this could happen in `job_spec`). Happy to consider a more elegant option.

Yeah, I think what you landed on is probably the best option.

While I don't know anything about LSF, nothing looks off to me in terms of where the changes are being made or the general content.

Thanks.

kyleam (approved):

yarikoptic:
Let's proceed!

PR #562, opened 2020-11-24
BF: DataladOrchestrator: Set DATALAD_SSH_IDENTITYFILE in __init__

### LINKED ISSUE #561: Can't fetch datalad-pair job results
Fetching the job results for a datalad-pair orchestrated job on an aws-ec2 instance fails with:

```2020-11-24 16:40:14,942 [ERROR  ] ConnectionOpenFailedError: 'ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/ch/.cache/datalad/sockets/bc54bd86 ubuntu@3.238.172.53' failed with exitcode 255 [Failed to open SSH connection (could not start ControlMaster process)] [sshconnector.py:open:541] (ConnectionOpenFailedError) ```

To reproduce:

```
reproman create -t aws-ec2 -b image=ami-0fe4bc00534545c58 -b instance_type=t3.large -b key_filename=$HOME/.ssh/aws -b key_name=<key name> -b user=ubuntu nitrc-ce
datalad create data
cd data
reproman run -r nitrc-ce --orchestrator datalad-pair pwd
reproman jobs <job id>
```
This AMI is CE-LITE; the same happens on ami-07ae3592c03add705 (CE standard).

Using the plain orchestrator works as expected.

kyleam:
DataladOrchestrator.prepare_remote() sets DATALAD_SSH_IDENTITYFILE so
that DataLad can work with SSH targets, in particular those for our
AwsEc2 resource, that need a custom key file that's not specified in
ssh_config.  Doing this in prepare_remote(), however, is problematic
because prepare_remote() isn't necessarily called before other methods
like fetch() that also require DATALAD_SSH_IDENTITYFILE to be
configured.  From the command line, this prevents fetching a job with
`reproman jobs` for a resource that depends on a custom key.

Move the DATALAD_SSH_IDENTITYFILE handling to __init__() so that it is
also in effect for resurrected orchestrators.

Fixes #561.

---

Given the current SSH setup for our tests, I haven't come up with a straightforward way to test this.

@chaselgrove If you have time, could you verify that this fixes the issue you reported?

### DIFF
--- a/reproman/support/jobs/orchestrators.py
+++ b/reproman/support/jobs/orchestrators.py
@@ -527,6 +527,14 @@ class DataladOrchestrator(Orchestrator, metaclass=abc.ABCMeta):
             _datalad_check_container(self.ds, self.job_spec)
             _datalad_format_command(self.ds, self.job_spec)
 
+        if isinstance(self.session, SSHSession) and resource.key_filename:
+            # Make the identity file available to 'datalad sshrun' even
+            # if it is not configured in .ssh/config. This is
+            # particularly important for AWS keys.
+            os.environ["DATALAD_SSH_IDENTITYFILE"] = resource.key_filename
+            from datalad import cfg
+            cfg.reload(force=True)
+
     @property
     @cached_property
     @borrowdoc(Orchestrator)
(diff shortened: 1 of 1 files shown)


chaselgrove:
Fixes #561

yarikoptic:
Thank you!

PR #564, opened 2021-01-04
ENH: enable etelemetry

yarikoptic:
This is a simpler version of https://github.com/ReproNim/reproman/pull/493/files
now that similar code is provided by etelemetry natively.
I decided not to issue warning in case of inability to check, but rather
just log at .debug level

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ matrix:
 env:
   global:
     - REPROMAN_TESTS_DEPS=full-except-datalad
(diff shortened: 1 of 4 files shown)


yarikoptic:
`python -m pytest -s -rs --integration -vv reproman/distributions/tests/test_venv.py::test_venv_identify_distributions` (as in the last matrix run on travis with python3.8) fails for me locally too (with python3.9) on master.  So it is unrelated to these changes, proceeding

PR #565, opened 2021-01-20
TST: venv: Adjust test for Python 3.8 and later

kyleam:
test_venv_identify_distributions() checks how links inside the virtual
environment are handled by identify_distributions().  On Travis, this
test has started to fail for the Python 3.8 build.  The last passing
run for the 3.8 build had Python 3.8.0, while the failure has 3.8.6.

  failing: https://travis-ci.org/github/ReproNim/reproman/jobs/753369730
  passing: https://travis-ci.org/github/ReproNim/reproman/jobs/745766696

The failing assertion is due to the layout in the environment not
meeting two expectations:

  * We expect the paths of abc.py and machinery.py to be links to the
    outside.

    $ realpath $VENV/lib/python3.7/abc.py
    /usr/lib/python3.7/abc.py
    $ realpath $VENV/lib/python3.7/importlib/machinery.py
    /usr/lib/python3.7/importlib/machinery.py

    But when I test with Ubuntu 20.04 (Python 3.8.5), nothing exists
    in $VENV/lib/python3.8 except for the site-packages directory.

  * We expect $VENV/bin/python to link elsewhere _within_ the
    environment.

    $ realpath $VENV/bin/python
    $VENV/bin/python3.7

    But with Ubuntu 20.04, it points _outside_.

    $ realpath $VENV/bin/python
    /usr/bin/python3.8

Deal with the first one by only tracing abc.py and machinery.py when
those files exist in the environment.  Keep the test for the second
case (i.e. a link in the environment that points to another
non-package file in the environment) by tracing bin/pip rather
bin/python.

I have not identified the upstream change that led to this layout
change.

### DIFF
--- a/reproman/distributions/tests/test_venv.py
+++ b/reproman/distributions/tests/test_venv.py
@@ -85,12 +85,28 @@ def test_venv_identify_distributions(venv_test_dir):
             # ... and relative paths work.
             os.path.join("venv1", libpaths["filters.py"]),
             # A virtualenv file that isn't part of any particular package.
-            os.path.join("venv1", "bin", "python"),
-            # A link to the outside world ...
-            os.path.join("venv1", libpaths["abc.py"]),
-            # or in a directory that is a link to the outside world.
-            os.path.join("venv1", libpaths["machinery.py"])
+            os.path.join("venv1", "bin", "pip")
         ]
+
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you @kyleam !

Issue #548, opened 2020-10-08
DOC: provide a complete example walk-though for a typical bids-app case (e.g. on mriqc)


yarikoptic:
We have a nice doc for `run` already: https://reproman.readthedocs.io/en/latest/execute.html ([sources for docs are in the repo](https://github.com/ReproNim/reproman/tree/master/docs))

Based on the successful testing of "run on AWS" (following the #438 prototypical usecase), documentation should be extended with a section on a use case which would go though all stages to run a BIDS-app using `reproman` on a condor cluster in AWS.  Building upon the https://github.com/ReproNim/containers/#a-typical-workflow it should start from a new installation of reproman (from pypi) and go through

- configure the remote resource (AWS condor cluster) (credentials etc but not copy-pasted into doc like @yarikoptic did ;))
- start the remote resource
- `reproman run` to submit for computation (examples in the #438 could be of help, and also the [webinar on containers presenting "reproman run"](https://www.repronim.org/webinar-series.html) should be referenced)
- `reproman jobs` to show how to monitor for completion
- `reproman jobs ID` to fetch the results and show "what is inside"
- stop/delete the resource upon successful completion

yarikoptic:
as of b9bb1a26fb4da49671e9647f71077828dbe69ea3 (in 0.4.0)  README.md contains a prototypical example.  The only things missing from above are `reproman jobs` since uses `--follow`.  I think that example should be expanded within documentation to provide a more thorough presentation

PR #567, opened 2021-01-28
Various ENH + RF +BF to make aws_ec2 and ec2_condor (more) usable

yarikoptic:
Individual commits have more information.
Changes are quite sloppy but the goal was to make things work:

- the whole dance started with trying to establish a new `ec2_condor` and it asking me for which key_name I want as many times (in parallel) as the `size` ;)  now it would not ask at all!:
- asking for key_name and breeding keys is not nice/good -- now key matches the name by default and we "mint" key_filename to match
  - now if we are to ask for a key_name, we would offer to choose from already present locally keys
- ec2_condor was not really usable without explicit specification of base dir to reside somewhere under `~/nfs-mount`.  Now should function "by default" since `~/.reproman` would be symlinked from that location
- unresolved stumbling point is https://github.com/datalad/datalad/issues/5382 forbidding fetching outputs (and probably upload of inputs if not available "publicly").  For now it requires adding to `~/.ssh/config` entry like
```
Host 54.166.157.42
 IdentityFile /home/yoh/.local/share/reproman/ec2_keys/my-nitrc.pem 
```

Closes #567 

ATM running a "realistic" test case which entailed (work towards fulfilling #548)

- `reproman create yoh-aws-hpc2 -t aws-condor -b size=2 -b key_name=my-nitrc -b instance_type=t2.medium` to establish 2 node cluster
- `reproman run -r yoh-aws-hpc2 --sub condor --orc datalad-pair --jp "container=containers/bids-mriqc" --bp subj=02,13 --follow --input 'sourcedata/sub-{p[subj]}' --output . '{inputs}' . participant group -w workdir --participant_label '{p[subj]}'` to mimic the example of https://github.com/ReproNim/containers#a-typical-workflow .  Tomorrow will see if/where/how it fails

### DIFF
--- a/reproman/resource/aws_condor.py
+++ b/reproman/resource/aws_condor.py
@@ -158,9 +158,10 @@ class AwsCondor(Resource):
         for i, node in enumerate(self.nodes):
             if not node._ec2_instance.public_ip_address:
                 node._ec2_instance = node._ec2_resource.Instance(node.id)
+            is_central_manager = (i == 0)
             condor_config = Template(
                 central_manager_ip=central_manager_ip,
-                is_central_manager=(i == 0),
+                is_central_manager=is_central_manager,
                 worker_nodes=self.nodes[1:]
             ).render_cluster("condor_config.local.template")
             session = node.get_session()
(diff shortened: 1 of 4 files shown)


yarikoptic:
crap -- it took travis a night to run the tests, thought to migrate but somehow the repronim organization does not even give me an option to grant or request, like it was may be denied at some point?

 Separate issue discovered (unrelated): https://github.com/ReproNim/reproman/issues/568

 I have pushed e5c3b80c7f03b114548113ac0e02694cf546c8ff so we install released datalad, lets see also if above (and the log flood https://github.com/datalad/datalad/issues/5386 ) improves.

But overall I consider this in usable state without immediate side-effects

PR #571, opened 2021-02-08
ci: Migrate to GitHub Actions

kyleam:
With the pending travis.org shutdown, we need to migrate to travis.com or another CI service.  This series switches to using GitHub actions.

All the old .travis.yml jobs should have a corresponding job in .github/workflows/test.yml.  As noted in the commit that adds the condor job, there was initially a condor_q-related core dump, but the latest runs haven't been triggering it, so (assuming it doesn't pop up again in this PR) that's obviously something to keep an eye on and ideally figure out.

PR #572, opened 2021-02-11
Adjust tests for Git's init.defaultBranch

kyleam:
This series adjust for Git's 2.28's `init.defaultBranch` configuration option:

 * Update a few test that fail when `init.defaultBranch` is set to a value other than "master".

 * Set `init.defaultBranch` to a value other than "master" in the GitHub CI runs to (1) guard against regressions and (2) silence the many `init.defaultBranch` warnings emitted when running the tests.

~This PR is marked as a draft because it sits on top of gh-571.  Here's the effective diff: https://github.com/repronim/reproman/compare/migrate-to-github-ci...harcoded-master~


yarikoptic:
THANK YOU! Looks LGTM

PR #573, opened 2021-02-11
MNT: Avoid invalid escape sequences in strings

kyleam:
A good number of regexps and a few other strings don't use an r-prefix
when the string contains non-escape-sequence backslashes.  This
doesn't make a functional difference at the moment because, when
Python encounters a backslash that produces an invalid escape
sequence, it automatically escapes the backslash.  But Python 3.6 and
later give a DeprecationWarning (<https://bugs.python.org/issue32912>),
with the goal of eventually making invalid escape sequences a syntax
error.

---

~This PR is marked as a draft because it sits on top of gh-571.  Here's the effective diff: https://github.com/repronim/reproman/compare/migrate-to-github-ci...invalid-escapes~

### DIFF
--- a/reproman/cmdline/helpers.py
+++ b/reproman/cmdline/helpers.py
@@ -65,7 +65,7 @@ class HelpAction(argparse.Action):
             opt_args_str = '*Global options*'
             pos_args_str = '*Commands*'
             # tune up usage -- default one is way too heavy
-            helpstr = re.sub('^[uU]sage: .*?\n\s*\n',
+            helpstr = re.sub(r'^[uU]sage: .*?\n\s*\n',
                              'Usage: reproman [global-opts] command [command-opts]\n\n',
                              helpstr,
                              flags=re.MULTILINE | re.DOTALL)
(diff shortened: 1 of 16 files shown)


kyleam:
There are some test failures.  Marking as draft...

yarikoptic:
thank you @kyleam !  I had been fixing for those here and there (in other projects) and totally forgot about the trove of them here ;)

yarikoptic:
probably here (and datalad) we should tune up tests execution so we fail on deprecation warnings and only whitelist the ones out of our control etc.  We do have [such setup e.g. in dandi-cli](https://github.com/dandi/dandi-cli/blob/master/tox.ini#L23) so far for the short term pain and long term gain ;-)

kyleam:
> probably here (and datalad) we should tune up tests execution so we fail on deprecation warnings and only whitelist the ones out of our control etc.

Sounds good.  This PR takes care of the bulk of those that are in our control, but there are still a few remaining ones.

---

The latest push should resolve the test failures.  I think it's correct, but perhaps I'm missing some (untested) detail of the `apt-cache-policy` output.

PR #575, opened 2021-03-03
MNT: Use 'datalad push' instead of deprecated 'datalad publish'

kyleam:
'datalad publish' was marked as deprecated in 0.14 in favor of
'datalad push', which was added in 0.13 (ReproMan's current minimum
version).  For the purpose of prepare_remote(), the main differences
are 1) the '' value for --since must be spelled as '^', and 2) push()
must be called with data=auto to retain the behavior of the publish()
call (though going forward it's worth considering whether data=nothing
should be used instead).


yarikoptic:
Looks good, thank you!

> though going forward it's worth considering whether data=nothing should be used instead

right -- probably might as well be `nothing` for that invocation since we just created a sibling and it would have no wanted set anyways... and if it is an existing one, having it as `auto` leaves possibility to trigger transmission of data which is not pertinent at that stage (of just establishing the remote, not when we transferring the inputs), so indeed "nothing" might be better.

PR #576, opened 2021-03-04
BF: orchestrators: Call datalad-push with --data=nothing

kyleam:
When 5f982f216 (2021-03-03) switched datalad-publish to datalad-push,
data="auto" was used to retain the behavior of the publish() call.
However, as mentioned in that commit and gh-575 [*], data="nothing" is
preferable here.  The original publish() call didn't use
transfer_data="auto" because there was an intent to transfer data
according to any configured preferred content; I just thoughtlessly
left the option at the default value.

[*] https://github.com/ReproNim/reproman/pull/575#issuecomment-790090050


yarikoptic:
cool, thanks @kyleam !

PR #574, opened 2021-02-25
ENH: slurm launcher

jbwexler:
Sorry for the delay, I've finally been able to work on this. This fix turns out to be pretty simple and seems to work well so far but I haven't done any kind of extensive testing yet. Just thought I would let you guys take a look. Let me know if you have questions. I think there's probably a more elegant way of creating the SlurmLauncherSubmitter class than how did it, which was simply to copy and paste the SlurmSubmitter class and make a few minor edits.

As I write this it occurs to me that we should also add some of the if statements in slurm_launcher.template to slurm.template in order to fix https://github.com/ReproNim/reproman/issues/549 

Some example output:

> (main) jbwexler@login2>reproman run --follow -r local --sub slurm_launcher --orc datalad-no-remote --batch-parameter sub=1,2,3,4 --input sub-0{p[sub]} --jp num_processes=4 --jp num_nodes=4 --jp walltime=00:02:00 --jp queue=normal sh ../test.sh
> 2021-02-25 13:57:49,928 [INFO   ] Making inputs available
> 2021-02-25 13:57:50,153 [INFO   ] Submitting 20210225-135748-648d
> 2021-02-25 13:57:50,158 [INFO   ] No root directory supplied for local; using '/home1/03201/jbwexler/.reproman/run-root'
> 2021-02-25 13:57:50,314 [INFO   ] Submitting /scratch/03201/jbwexler/reproman_test/ds000001/.reproman/jobs/local/20210225-135748-648d/submit
> 2021-02-25 13:57:50,561 [INFO   ] Job 20210225-135748-648d submitted as slurm_launcher job 3286023
> 2021-02-25 13:57:50,571 [INFO   ] Registered job 20210225-135748-648d
> 2021-02-25 13:57:50,596 [INFO   ] Waiting on job 3286023: PENDING. Next heartbeat in 5 seconds
> 2021-02-25 13:57:55,856 [INFO   ] Waiting on job 3286023: PENDING. Next heartbeat in 5 seconds
> 2021-02-25 13:58:01,730 [INFO   ] Waiting on job 3286023: PENDING. Next heartbeat in 12 seconds
> 2021-02-25 13:58:14,224 [INFO   ] Waiting on job 3286023: PENDING. Next heartbeat in 26 seconds
> 2021-02-25 13:58:45,751 [INFO   ] Waiting on job 3286023: PENDING. Next heartbeat in 60 seconds
[...]

PR #577, opened 2021-03-04
Enh/launcher

jbwexler:
I ended up creating a new branch to keep it a bit cleaner. Continuation of https://github.com/ReproNim/reproman/pull/574

### DIFF
--- a/reproman/interface/run.py
+++ b/reproman/interface/run.py
@@ -199,6 +199,12 @@ JOB_PARAMETERS = collections.OrderedDict(
          """Supported by Condor and PBS submitters."""),
         ("num_nodes, walltime",
          """Supported by PBS submitter."""),
+        ("queue",
+         """Supported by Slurm submitter."""),
+        ("launcher",
+         """If set to "true", the job will be run using Launcher, rather than 
+        as a job-array. See https://github.com/TACC/launcher for more info. 
+        Supported by Slurm and PBS submitters."""),
     ]
 )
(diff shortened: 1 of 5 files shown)


jbwexler:
It looked like pbs was the only other type of submitter that is supported by Launcher at the moment, so I only added launcher compatibility to slurm and pbs. https://github.com/TACC/launcher/tree/master/plugins

kyleam:
Thanks for the update, @jbwexler.  Looks good.  I've pushed a fixup for the test failures and will look into adding a test that uses launcher.

kyleam:
I said:
> and will look into adding a test that uses launcher.

The new tests pass, but codecov flagged some of the new bits in test_orchestrators.py as unexecuted.  Either the tests aren't working as I intended or coverage is confused.  I'll first make sure that the new assertion is being executed locally and then I will push a temporary commit to this PR's branch to check that it's being triggered by the github workflow.

kyleam:
> then I will push a temporary commit to this PR's branch to check that it's being triggered by the github workflow.

The expected lines were triggered.  I'll drop the tip commit and conclude that codecov was/is confused.

https://github.com/ReproNim/reproman/pull/577/checks?check_run_id=2040423024

kyleam (approved):
Thanks again, @jbwexler.

jbwexler:
No problem, thanks for taking the time to incorporate the PR!

PR #584, opened 2022-12-15
Get tests passing

asmacdo:
Its been a while, many versions have shifted, Python 3.6 is EOL and Ubuntu 16.04 is End of Standard Support (EOL is 2026 if we really want to keep it)

This is the first set of problems that I was able to fix, but this doesn't get all the tests passing. Im having a bit of a struggle getting the remaining failing tests to replicate locally.

~- [ ] KeyError: 'p'~ https://github.com/ReproNim/reproman/issues/588
- [x] Conda conflicts
- [x] lib module

## KeyError: 'p'    
`reproman/support/jobs/tests/test_orchestrators.py::test_orc_datalad_concurrent[sub:local-orc:local-run] - KeyError: 'p'`

Suspect: https://github.com/ReproNim/reproman/blob/master/reproman/support/jobs/orchestrators.py#L470

Note: I haven't been able to reproduce this one locally yet (environment setup challenge)

## Conda conflicts
`reproman/distributions/tests/test_conda.py::test_conda_init_install_and_detect`

It looks like this type of problem has occurred a few times before. Its not clear to me if these package versions are related to installed package versions or if this is all in a mock environment. I suppose the approach is to play with the versions until it works.

        UnsatisfiableError: The following specifications were found to be incompatible with each other:
        
        Output in format: Requested package -> Available versions
        
        Package pypy3.6 conflicts for:
        pip=20.0.2 -> python -> pypy3.6[version='7.3.0.*|7.3.1.*|7.3.2.*|7.3.3.*|>=7.3.3|>=7.3.2|>=7.3.1']
        conda=4.8.2 -> conda-package-handling[version='>=1.3.0'] -> pypy3.6[version='7.3.0.*|7.3.1.*|7.3.2.*|7.3.3.*|>=7.3.1|>=7.3.2|>=7.3.3']
        
        Package wheel conflicts for:
        pip=20.0.2 -> wheel
        python=3.9 -> pip -> wheel
        
        Package pip conflicts for:
        python=3.9 -> pip
        conda=4.8.2 -> python[version='>=3.8,<3.9.0a0'] -> pipThe following specifications were found to be incompatible with your system:
        
          - feature:/linux-64::__glibc==2.27=0
          - feature:|@/linux-64::__glibc==2.27=0
          - python=3.9 -> libgcc-ng[version='>=10.3.0'] -> __glibc[version='>=2.17']

## Lib module
Reported: `reproman/interface/tests/test_delete.py::test_delete_interface - SystemExit: 1`
Actual issue?: `reproman.support.exceptions.ResourceError: Failed to import resource: module 'lib' has no attribute 'ERR_load_RAND_strings' [rand.py:<module>:211]. Similar backends: aws-ec2`

I'm guessing the failure to import means the resource isnt created, so delete fails


asmacdo:
If preferred, theres no reason to hammer CI here, I'm running against my own fork. https://github.com/asmacdo/reproman/pull/2

reproman/tests/test_cmdline_main.py
```diff
@@ -77,6 +77,7 @@ def test_help_np():
     # but order is still not guaranteed (dict somewhere)! TODO
     # see https://travis-ci.org/reproman/reproman/jobs/80519004
     # thus testing sets
+    # TODO(asmacdo) On fedora "*Global Options*" renders as "Options" and is not included in sections
```
yarikoptic, line 80: sounds like a python-version related thing we also fixed for some other case , rests in ad-hoc tuning of argparser setup -- are you sure this is still the case?

reproman/tests/skip.py
```diff
@@ -56,6 +56,7 @@ def windows():
 # function to CONDITION_FNS.
 
 
+# TODO(asmacdo) maybe should make new `no_debian`
```
yarikoptic, line 59: may be.... it is a bit ambigious on the word though, might be `not_on_debian` since there could be `debian` available via WSL on windows etc ;)

yarikoptic: what was your idea/use-case here more specifically?

reproman/resource/tests/test_session.py
```diff
@@ -394,19 +394,20 @@ def import_resource(mod, cls):
...
-        ("singularity", "PTYSingularitySession")
-    ],
-    ids=lambda x: x[1])
-def test_session_singularity(location, singularity_resource, check_methods):
-    """Test sessions that depend on `singularity_resource` fixture.
-    """
-    cls = import_resource(*location)
-    session = cls(singularity_resource.name)
-    check_methods(location[1], session)
+# TODO(asmacdo) we need to figure out how to work with apptainer (and singularity?)
```
yarikoptic, line 397: should they just be skipped if there is no singularity and otherwise run?

reproman/tests/test_cmdline_main.py
asmacdo: nope- you fixed this one :)

reproman/resource/tests/test_session.py
asmacdo: May be, but I assume there is an "oldest singularity" we can support. Unfortunately when apptainer is installed, singularity is aliased to apptainer, but apptainer versions started over at 1.0.

reproman/tests/skip.py
asmacdo: 2 tests were not skipped on fedora:
- https://github.com/ReproNim/reproman/pull/584/files#diff-2ed961bebca8aef436f7a673805596bae4c80aeed8836eb3593d87a0123db3e7R129
 - https://github.com/ReproNim/reproman/pull/584/files#diff-2ed961bebca8aef436f7a673805596bae4c80aeed8836eb3593d87a0123db3e7R140

I put this todo to make sure that skipif no apt is appropriate for those 2, or if it should just be not deb.

reproman/resource/tests/test_session.py
yarikoptic: then skip if apptainer is installed, but generally should be ran if singularity (not apptainer'ed) is installed.

CONTRIBUTING.md
```diff
@@ -112,8 +112,12 @@ we outline the workflow used by the developers:
 
 Development environment
 -----------------------
+TODO(asmacdo) include python-devel for header files
+TODO(asmacdo) include sqlite3 (on fedora35 sqlite2, sqlite-devel did the trick)
+TODO(asmacdo) include sqlite3 (on fedora35 sqlite2, sqlite-devel did the trick)
+TODO(asmacdo) install datalad
 
-We support Python 3 (>= 3.6).
+We support Python 3 (>= |MinPython|).
```
yarikoptic, line 120: I think we had this confusion somewhere else as well -- nothing would replace this here, so hardcode 3.7 (we drop 3.6)
```suggestion
We support Python 3 (>= 3.7).
```

reproman/distributions/conda.py
```diff
@@ -60,7 +61,7 @@ def get_conda_platform_from_python(py_platform):
     return None
 
 
-def get_miniconda_url(conda_platform, python_version):
+def get_miniconda_url(conda_platform, python_version, conda_version):
```
yarikoptic, line 64: adjust docstring to reflect adding `conda_version`. But overall I feel that ideally we should allow for it to be just `latest` if not specified. I don't think we would be able to find installations for all versions of `conda` -- people might install for some conda version and then upgrade it.  Since not directly needed ATM

reproman/distributions/tests/test_singularity.py
```diff
@@ -21,7 +21,8 @@
 @mark.skipif_no_singularity
 @pytest.mark.xfail(reason="Singularity Hub is down", run=False)
 @pytest.mark.xfail(
-    external_versions["cmd:singularity"] >= '3',
+	# TODO cmd:apptainer version is 1.1.4
+    bool(external_versions["cmd:singularity"]),
```
yarikoptic, line 25: I feel that it is yet another case for `skipif_apptainer` or alike instead of just xfailing it.

reproman/interface/tests/test_execute.py
```diff
@@ -124,6 +124,8 @@ def test_trace_docker(docker_container, trace_info):
 @pytest.mark.integration
 @mark.skipif_no_network
 @mark.skipif_no_apt_cache
+# Error: trace.sqlite3 doesnt exist in the directory https://github.com/VIDA-NYU/reprozip/issues/387
+@pytest.mark.xfail(reason="Reprozip binary is currently broken.")
```
yarikoptic, line 128: I would just make it a bit more specific and not blame reprozip itself
```suggestion
@pytest.mark.xfail(reason="Our extracted standalone reprozip tracer is currently broken.")
```

reproman/support/external_versions.py
```diff
@@ -88,6 +88,15 @@ def _get_singularity_version():
     out = _try_run(["singularity", "--version"])[0]
     return out.split(' ')[-1].split("-")[0].split("+")[0]
 
+def _get_apptainer_version():
+    """Return version of available singularity."""
```
yarikoptic, line 92:
```suggestion
    """Return version of available apptainer."""
```

reproman/ui/progressbars.py
```diff
@@ -89,7 +89,8 @@ def finish(self):
                 self._pbar.finish()
             super(progressbarProgressBar, self).finish()
 
-    progressbars['progressbar'] = progressbarProgressBar
+	# TODO(asmacdo) progress bar fails for some reason, skipping for now
+    # progressbars['progressbar'] = progressbarProgressBar
```
yarikoptic, line 93: I think we removed support for the `progressbar` -- only `tqdm` . But I am not sure why we would fail here at all since we just should skip if no `progressbar` module... anyways -- I would recommend to just strip off `progressbar` support and rely only on `tqdm` and make it the one `setup.py` installs

yarikoptic (changes requested):
various suggestions etc

README.md
```diff
@@ -27,7 +27,8 @@ internals and/or contributing to the project.
 
 # Installation
 
-ReproMan requires Python 3 (>= 3.6).
+TODO(asmacdo) Set up sphinx substitution
+ReproMan requires Python 3 (>= |MinPython|).
```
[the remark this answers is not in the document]
asmacdo: Already changed in latest

CONTRIBUTING.md
```diff
@@ -112,8 +112,12 @@ we outline the workflow used by the developers:
 
 Development environment
 -----------------------
+TODO(asmacdo) include python-devel for header files
+TODO(asmacdo) include sqlite3 (on fedora35 sqlite2, sqlite-devel did the trick)
+TODO(asmacdo) include sqlite3 (on fedora35 sqlite2, sqlite-devel did the trick)
```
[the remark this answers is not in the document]
asmacdo: removed in latest commit

reproman/distributions/conda.py
asmacdo: fixed

reproman/resource/tests/test_session.py
asmacdo: skipped in latest commit

reproman/tests/skip.py
asmacdo: apt-cache is sufficieent for now

reproman/distributions/tests/test_singularity.py
asmacdo: I went with any singularity version to account for apptainer. We can come back to this and use skipif_apptainer

reproman/support/external_versions.py
asmacdo: I removed this function for now

reproman/ui/progressbars.py
asmacdo: removed progressbar class!

README.md
```diff
@@ -54,7 +54,10 @@ all necessary packages from that repository.
 
 ## Dependencies
 
-Our `setup.py` and corresponding packaging describes all necessary dependencies.
+Python 3.7+ **with .h extensions** This is provided by `python-dev` on
```
yarikoptic, line 57:
```suggestion
Python 3.8+ **with header files** possibly needed to build some extensions without wheels. They are provided by `python3-dev` on
```

yarikoptic (approved):
Left one wording fix suggestion, but otherwise good to proceed

PR #592, opened 2023-01-18
Use fixed rztracer and fix up the unittest for it to use another binary than ls

yarikoptic:
See individual commits with more information and reasoning.

### DIFF
--- a/reproman/interface/tests/test_execute.py
+++ b/reproman/interface/tests/test_execute.py
@@ -124,8 +124,6 @@ def test_trace_docker(docker_container, trace_info):
 @pytest.mark.integration
 @mark.skipif_no_network
 @mark.skipif_no_apt_cache
-# Error: trace.sqlite3 doesnt exist in the directory https://github.com/VIDA-NYU/reprozip/issues/387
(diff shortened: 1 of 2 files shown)


reproman/interface/tests/test_execute.py
```diff
@@ -136,7 +134,7 @@ def test_trace_local(trace_info):
                    return_value=ResourceManager()):
             with patch("reproman.interface.execute.CMD_CLASSES",
                        {"trace": trace_info["class"]}):
-                execute("ls", ["-l"], trace=True, resref="testing-local")
```
asmacdo, line 139: interesting why this change?

asmacdo (approved):
Looks sane to me. For future us, the commit stream used here is present on the [extract-c-tracer branch](https://github.com/ReproNim/reprozip/tree/extract-c-tracer)

The changes there can be summarized as: 
 - merge in a handful of updates from upstream reprozip
 - rewrite the our custom logging to work with the new changes
 
[This PR and disscussion](https://github.com/VIDA-NYU/reprozip/pull/178) is useful background on how this fork came about. IIUC, it comes down Reproman needing to use a python logger instead of the non-python logger that reprozip uses out of the box. This is done with a custom log.c file that emulates the python logs.

PR #614, opened 2023-06-01
--version should return version, not license

asmacdo:


yarikoptic:
CI is not happy since needs some TLC but unlikely from this PR changes

PR #615, opened 2023-06-01
tolerate codespell

asmacdo:


yarikoptic:
;-) it is green now -- the way to go in!

PR #627, opened 2025-09-02
Add tutorial

asmacdo:
Built on https://github.com/ReproNim/reproman/pull/626 which should merge first.

I tested this on typhon. (I had to use podman instead of docker since I am not in the docker group, but it still worked well).

TODO:
 - [x] merge #626 
 - [X] workdir add with datalad run
 - [X] Explain datalad a little bit, ie the dataset needs to be installed but the data itself does not need to be fetched locally.
 - [X] Go into a little more detail about whats happening during run, ie data specified by --input is fetched
 - [X] Be more newbie friendly and explain the docker/podman invocation, especially volume mounts.
 - [X] do not use abbreviated parameters in the tutorial ie --orc
 - [X] fix list spacing
 - [X] clarify reproman installation on localhost not remote resource

### DIFF
--- a/docs/source/execution.rst
+++ b/docs/source/execution.rst
@@ -54,7 +54,7 @@ Choosing an orchestrator
 ------------------------
 
 Orchestrators are responsible for preparing the remote and collecting the results.
- The complete set of orchestrators, accompanied by descriptions, can be seen by
+The complete set of orchestrators, accompanied by descriptions, can be seen by
 calling ``reproman run --list=orchestrators``.
 
 .. note::


docs/source/tutorial-ssh.rst
```diff
@@ -0,0 +1,178 @@
...
+
+This tutorial demonstrates ReproMan's power in creating reproducible, traceable computational workflows across SSH-accessible computing environments.
+
+Overview
+========
+
+We'll cover two workflows:
+
+**Part 1: Hello World Example**
+1. Create a ReproMan SSH resource
```
asmacdo, line 17: todo: fix spacing, does not render as numbered list

docs/source/tutorial-ssh.rst
```diff
@@ -0,0 +1,178 @@
...
+1. Set up a DataLad dataset with input data
+2. Execute MRIQC quality control analysis remotely  
+3. Collect and examine results with full provenance
+
+Prerequisites
+=============
+
+- ReproMan installed (``pip install reproman``) 
+- Access to a remote server via SSH
+- For Part 2: DataLad support (``pip install 'reproman[full]'``)
```
asmacdo, line 31: TODO make it clear that install is on local machine, reproman does not need to be installed on resource, but datalad does)

PR #629, opened 2025-09-03
Add docs build test to CI

asmacdo:
- [ ] see if minimal install steps will work
- [ ] break docs build and make sure test fails


docs/Makefile
```diff
@@ -2,9 +2,7 @@
 #
 
 # You can set these variables from the command line.
-# Treat warnings as errors.... hierarchy.rst is just a dump of notes ATM
-# SPHINXOPTS    = -W
-SPHINXOPTS    =
+SPHINXOPTS    = -W
```
yarikoptic, line 5: are we really ready for this? ;)

asmacdo: Yes it passed without errors.

The "discussion" docs have been moved to markdown in design/ which were probably causing most of the problems. I only had to make minimal changes to avoid.

The weird thing was that without this the doc build passed if it completed with errors (seems like it this should only affect warnings?) I guess non-fatal errors were treated as warnings?

asmacdo:
Since rtd has deploy previews now, it seems this isnt necessary (though there could still be value in testing "-W" but not using that for actual deploys). I've marked as a draft. 

The changes to pass with -W, and the removal of the docs build from the test jobs should still be kept.

yarikoptic:
please rebase @asmacdo -- CI must be green now

asmacdo:
I've changed my mind about not needing the docs test build. This test is going to be more strict than the rtd build because of `-W`, and it also exercises the make target (which we want to work for local development). Its kind of nice to have that separation since we will still have a deploy preview even in cases that the docs test fails due to warnings/non-critical-errors.

yarikoptic:
green is good, let's proceed

PR #630, opened 2025-09-03
remove pycrpyto

asmacdo:
- I did not see any direct usage of this, ie grep -rn "Crypto"
- The install of pycrpyto fails in CI with: longintrepr.h: No such file or directory (installing dev python with headers did not help)
- project is archived, unmaintained, and has known vulnerabilities

Locally I didnt see any extra tests failing. This could just be a legacy requirement?

To test I created a new pyenv, created a new ssh-type resource (typhon), and I was able to `login` and `execute`. I am not confident that this is sufficient, but maybe we should just remove anyway, and if something breaks we fix it with some other library.


asmacdo:
This dep was added in https://github.com/ReproNim/reproman/pull/101/ (back in the niceman days)

Commit message indicates that it was necessary to login to AWS EC2. I haven't worked with reproman on EC2 yet, but expanding the tutorial to cover HTCondor and AWS is something I want to get to soon anyway. 

IMO
 - Even if this breaks EC2, we should not be depending on an unmaintained crypto library
 - I'd still like to remove this now to fix CI installation (and rtd docs build)
 - Ill keep this in mind when working with EC2 and address if it is broken. Filed https://github.com/ReproNim/reproman/issues/633 so we dont lose track.

PR #632, opened 2025-09-04
Fix readthedocs build

asmacdo:
rtd now uses  v2  configuration, this PR updates to the new requirements.

To test I had rtd build from this branch. I've now reverted that setting change after build completed.

- [x] merge https://github.com/ReproNim/reproman/pull/630 first


asmacdo:
Checked again with the gemini suggestion and docs build as expected

